PR #1, opened 2024-01-17
build: add nix dev + build environment + ci builds

mrcjkb:
The obligatory flakeification :smile:

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,81 @@
+{
+  description = "Luarocks <3 Rust";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
+    flake-parts.url = "github:hercules-ci/flake-parts";
+    pre-commit-hooks = {
+      url = "github:cachix/pre-commit-hooks.nix";
+      inputs.nixpkgs.follows = "nixpkgs";
+    };
+  };
+
+  outputs = inputs @ {
+    self,
+    nixpkgs,
+    flake-parts,
+    pre-commit-hooks,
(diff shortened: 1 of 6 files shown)


flake.nix
```diff
@@ -0,0 +1,83 @@
+{
+  description = "devShell for Rust projects";
```
teto, line 2: thats a bit generic

mrcjkb: Oh yeah, that's from my template...

mrcjkb:
@teto for some reason, the extra `buildInputs` I add to the devShell (`rust-analyzer`, `rustfmt`, ...) aren't available on the `PATH`.
Any idea what could be the issue?

teto:
I am never too sure about the rules but because it's not a mkShell but the program buildInputs, in which case you need propagatedBuildInputs to propagate to PATH .
Maybe if you append `.inputDerivation` at the end ?

mrcjkb:
> I am never too sure about the rules but because it's not a mkShell but the program buildInputs, in which case you need propagatedBuildInputs to propagate to PATH . Maybe if you append `.inputDerivation` at the end ?

:thinking: none of it worked. Maybe it's something that `rustPlatform.buildRustPackage` does...
I ended up using `mkShell` and appending `rocks.buildInputs ++ rocks.nativeBuildInputs` to the shell's `buildInputs`.
A bit hacky, but it works :shrug:

teto (approved):
I've had similar issues in the past so I get it. Approved

mrcjkb:
Oops... forgot to set it to squash...

PR #2, opened 2024-01-21
[FEATURE]: Unpacking logic

vhyrro:
Starting work on rock unpacking. Seems to be a pretty fiddly topic just looking at the luarocks source code :sweat_smile:


vhyrro:
There are two breaking changes in this PR:
- `download` now returns a `DownloadedRock` instead of a tuple of strings
- `unpack` has been split into `unpack` and `unpack-remote` to make things manageable

cc @teto @mrcjkb :)

mrcjkb (commented):
Not sure why CI is failing. Maybe it needs another `cargo update`?

vhyrro:
Also no idea why the CI is failing here :/

PR #8, opened 2024-01-30
feat(lib/rockspec): parse build information

mrcjkb:
Followed by #9 

In a follow-up PR, I will replace the `BuildType` enum with a `BuildBackendSpec` type that we will parse using an intermediate struct, similar to what I did with the source spec.


vhyrro:
LGTM

PR #11, opened 2024-02-06
feat(rockspec): per platform overrides

mrcjkb:
Stacked on #10


rocks-lib/src/rockspec/dependency.rs
```diff
@@ -58,6 +61,33 @@ impl<'de> Deserialize<'de> for LuaDependency {
     }
 }
 
+impl<'lua> FromLua<'lua> for PerPlatform<Vec<LuaDependency>> {
+    fn from_lua(value: Value<'lua>, lua: &'lua Lua) -> mlua::Result<Self> {
+        match &value {
+            list @ Value::Table(tbl) => {
+                let default = lua.from_value(list.clone())?;
+                let per_platform = match tbl.get("platforms")? {
```
mrcjkb, line 69: TODO: the per platform override should have the default merged into it.

vhyrro:
I think we can merge the PR now as-is and then make improvements/changes to the actual codebase later on. What do we think?

PR #14, opened 2024-02-09
fix(platform): remove `Ord` instance of `PlatformIdentifier`

mrcjkb:
It really shouldn't have an `Ord` instance.

I'm not sure about how we sort platforms by specificity (see the comment I added).

### DIFF
--- a/rocks-lib/src/rockspec/platform.rs
+++ b/rocks-lib/src/rockspec/platform.rs
@@ -1,11 +1,7 @@
 use eyre::{eyre, Result};
 use itertools::Itertools;
 use mlua::{FromLua, Lua, LuaSerdeExt as _, Value};
-use std::{
-    cmp::{Ordering, Reverse},
-    collections::HashMap,
-    marker::PhantomData,
-};
+use std::{cmp::Ordering, collections::HashMap, marker::PhantomData};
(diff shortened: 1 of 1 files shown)


mrcjkb:
🤔 We should probably group and filter the platforms that can be sorted...

vhyrro:
Is there a very concrete problem that could occur here? As long as there isn't I believe we'll be fine (famous last words). Or, in other words, as long as the more specific matches appear before the less specific matches, we don't *really* care about what order the rest will be in, right?

mrcjkb:
> Is there a very concrete problem that could occur here? As long as there isn't I believe we'll be fine (famous last words). Or, in other words, as long as the more specific matches appear before the less specific matches, we don't _really_ care about what order the rest will be in, right?

The problem is that we don't have control over the order in which the rockspec specifies the platform overrides.
If it has non-orderable platforms between the orderable ones, we can't sort them properly.

mrcjkb:
> Perhaps we really should then partition the vector into specialized platforms and the regular platforms and just chain them together after the fact (such that the specialized platforms are guaranteed to be at the front). I'm pretty sure itertools has some crazy function to do all of that in one go 😂

It ended up being a lot simpler :smile:

PR #18, opened 2024-06-01
feat: add builtin build step

vhyrro:
This PR will add a functional `builtin` build step for rocks.

Additionally, this PR refactors a few bits and bobs related to the build process to make everything cleaner and more readable with less duplication :)


vhyrro:
I decided to do something slightly controversial but I believe it's a good decision in the long run (cc @mrcjkb) - I decided to break the rocktree schema and implemented my own.

Instead of the all-over-the-place directory structures I've now reduced a rock tree to the following:
```
/home/vhyrro/.local/share/rocks/
└── 5.1
    └── neorg@scm-1
        ├── etc // contains extra files (documentation)
        ├── lib // contains shared library objects
        └── src // contains source code for the rock
```

All rocks have an associated version number and the directories that distinguish various parts of the rock are very clear-cut and easy to understand.

Below are some extra bits of information about this change:
- A rocks server and a rocks tree is no longer "the same thing" - trying to merge them together like luarocks does it makes the whole implementation much more complicated. A rocks server should host rockspecs or packed rocks, whereas a rocks tree should only contain unpacked rocks. I do not plan on breaking the directory layout of *servers*, as those are important for interoperability with luarocks. Rather, I'm only interested in changing the directory structure of rocks trees.
- Rocks stores its data in a different place to luarocks, so keeping rocks tree compatibility is not really incentivized.
- If a user wants to mirror their old rocks tree in `rocks` then they can simply reinstall all their favourite rocks using our CLI. The end result is the same, the backend is different.

vhyrro:
All the boilerplate is in place now... all that's left is fixing a weird deserialization issue that I can't figure out quite just yet.

vhyrro:
Leaving it as-is for now, I'm finding it hard to debug the issue :p

vhyrro:
This is ready to review now. Not all features are fully set in stone yet, for instance I need to figure out how to read source.defines and source.incdirs. Additionally, more `url = "schema://"`s need to be supported (only git is supported for now). I'll be working on those in the meantime, exciting to see this feature in a working state :)

rocks-lib/src/build/mod.rs
```diff
@@ -0,0 +1,70 @@
...
+    // TODO(vhyrro): Instead of copying bit-by-bit we should instead perform all of these
+    // operations in the temporary directory itself and then copy all results over once they've
+    // succeeded.
+
+    let tree = Tree::new(
+        &config.tree,
+        config
+            .lua_version
+            .as_ref()
+            .unwrap_or(&crate::config::LuaVersion::Lua51),
```
vhyrro, line 44: `lua_version` should be read from a configuration file instead of defaulting to Lua 5.1 here :)

lua-cjson.rockspec
```diff
@@ -0,0 +1,40 @@
+package = "lua-cjson"
```
mrcjkb, line 1: question: is this for manual/automatic testing?
I would put something like this in `rocks-lib/resources/test/`

rocks-bin/src/build.rs
```diff
@@ -0,0 +1,29 @@
...
+use rocks_lib::{config::Config, rockspec::Rockspec};
+
+#[derive(Args)]
+pub struct Build {
+    directory: PathBuf,
+}
+
+pub fn build(data: Build, config: &Config) -> Result<()> {
+    if data
+        .directory
```
mrcjkb, line 15: question: why the name `directory` here?

rocks-bin/src/main.rs
```diff
@@ -1,13 +1,26 @@
...
 
+mod build;
 mod download;
 mod rockspec;
 mod search;
 mod unpack;
 
+fn parse_lua_version(s: &str) -> Result<LuaVersion, String> {
+    match s {
+        "5.1" | "51" | "jit" | "luajit" => Ok(LuaVersion::Lua51),
```
mrcjkb, line 14: question: would `luajit` be considered valid? I recall seeing that somewhere.

rocks-bin/src/main.rs
```diff
@@ -41,8 +54,8 @@ struct Cli {
 
     /// Which Lua installation to use.
     // TODO(vhyrro): Add option validator for the version here.
```
mrcjkb, line 56:
```suggestion
```

rocks-lib/src/build/mod.rs
```diff
@@ -0,0 +1,70 @@
+use crate::{
+    config::Config,
+    rockspec::{Build, BuildBackendSpec, RockSourceSpec, Rockspec},
+    tree::Tree,
+};
+use eyre::Result;
+use git2::Repository;
+
+pub fn build(rockspec: Rockspec, config: &Config, no_install: bool) -> Result<()> {
```
mrcjkb, line 9: suggestion(readability, non-blocking): I'm not really a fan of "negative" boolean parameters.

```suggestion
pub fn build(rockspec: Rockspec, config: &Config, install: bool) -> Result<()> {
```

Perhaps it would be even nicer to omit the parameter and extract the logic into a separate function.

rocks-lib/src/build/mod.rs
mrcjkb: Might also be nice to be able to set it from an environmen variable

rocks-lib/src/rockspec/build/builtin.rs
```diff
@@ -1,23 +1,30 @@
...
 #[derive(Debug, PartialEq, Deserialize, Default, Clone)]
 pub struct BuiltinBuildSpec {
     /// Keys are module names in the format normally used by the `require()` function
-    pub modules: HashMap<String, ModulesSpec>,
+    pub modules: HashMap<String, ModuleType>,
 }
 
 #[derive(Debug, PartialEq, Clone)]
-pub enum ModulesSpec {
+pub enum ModuleType {
```
mrcjkb, line 19: thought: I find the word "type" here a bit confusing, given that rust has a nominal type system. 
The type here is `ModuleSpec` (or `ModuleType`), and each of the variants is a constructor.

That's why I chose `Spec`.

mrcjkb (commented):
[review incomplete] To be continued at a later time 😃

lua-cjson.rockspec
vhyrro: Manual testing, I'll be removing it. It got added automatically on accident because I use `jj` instead of `git` hahah

rocks-lib/src/build/mod.rs
vhyrro: Neither am I, I was following the function signature that `luarocks` uses here. I thought it could come in handy once we start working on the compatibility scripts for various "installer" rocks (like `luarocks-build-rust-mlua`). The function invoked there does supply a `no_install`.

I can only assume it was created that way because under 90% of cases you want to build and install, it's very unlikely you'll want to just build. I'm happy to remove this in general though in order make the code cleaner.

rocks-lib/src/rockspec/build/builtin.rs
vhyrro: To me, a spec is for structures, not for variations of a single bit of data (like the enum in this case).

It's not a convention to call types `xyzType` in Rust either, I just feel like it's much easier to grasp the intent at a glance. It's the "type" of "module" to be used :p

This is all bikeshedding in the end, so I don't mind whether it's called this or that, I just find it a more deliberate naming scheme.

rocks-lib/src/tree/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+    version: &'a LuaVersion,
+    /// The root of the tree.
+    root: &'a PathBuf,
+}
+
+/// Change-agnostic way of referencing various paths for a rock.
+pub struct TreeLayout {
+    pub etc: PathBuf,
+    pub lib: PathBuf,
+    pub src: PathBuf,
```
mrcjkb, line 28: suggestion: Perhaps these should be `Option`s, since not every rock has all of these paths?

rocks-lib/src/rockspec/build/builtin.rs
```diff
@@ -54,6 +61,98 @@ pub struct ModulePaths {
...
+                }
+                ModuleType::ModulePaths(data) => {
+                    let path = PathBuf::from(destination_path);
+                    std::fs::create_dir_all(path.parent().unwrap())?;
+
+                    // TODO: Defines, libraries
+                    cc::Build::new()
+                        .shared_flag(true)
+                        .files(&data.sources)
+                        .includes(&data.incdirs)
```
mrcjkb, line 100: praise: woah, this looks so much cleaner than the lua implementation :star_struck:

rocks-lib/src/rockspec/build/builtin.rs
vhyrro: Fair enough, I'll move it back to `ModulesSpec` for consistency's sake :)

rocks-lib/src/rockspec/build/builtin.rs
vhyrro: Had to add a bunch more flags for everything to work now but yeah it's very nice to work with :D

rocks-lib/src/rockspec/build/builtin.rs
```diff
@@ -54,6 +87,139 @@ pub struct ModulePaths {
...
+                    let destination_path =
+                        lua_module_to_pathbuf(destination_path, std::env::consts::DLL_SUFFIX);
+                    let target = output_paths.lib.join(destination_path);
+
+                    let parent = target.parent().expect("TODO");
+                    let file = target.file_name().expect("TODO");
+
+                    std::fs::create_dir_all(parent)?;
+
+                    cc::Build::new()
```
mrcjkb, line 123: question: do we need to add lua's `include/` here?

suggestion: we should add a test that builds a simple lua C library and make sure we can use it with lua.

vhyrro: Great question... when testing with lua-cjson, there were no compilation issues. We should definitely test on a range of different rockspecs, since I can indeed imagine that this may cause problems if not implemented.

rocks-lib/src/tree/mod.rs
vhyrro: I'm slightly inclined to create all of these anyway just for consistency's sake. Then external apps don't need to check for the existence of the directory before they try to access any content inside (makes globbing across all rocks simpler too). It's not like a single directory takes up gigabytes of storage.

I'm impartial to this change, so we can go one way or the other here :p

rocks-lib/src/build/mod.rs
vhyrro: What I did for now is I removed the `no_install` flag from the library API (it's completely useless there), but kept it in the internal build code for the time being. Then, depending on how the compatibility layer pans out, we can remove or alter that flag.

rocks-bin/src/main.rs
vhyrro: LuaJIT can be a valid target (since mlua can be compiled to use it), so sure, this should be allowed. Luarocks itself doesn't recognize this though.

What I'm curious about is whether it could be possible to supply custom parameters here based on a list defined in e.g. a config file. That could allow us to use Neovim as an interpreter super easily!

vhyrro:
I recognize that some parts of the code are slightly smelly and imperfect (including the large amount of for loops hah), but I'm not afraid of aggressively refactoring this later down the line once I set up the test suite for this and any potential regression tests. I'm prioritizing development speed over beauty for the time being.

rocks-lib/src/tree/mod.rs
max397574: how would you create them then?
just empty? like this external apps just would have to check if the contents exists instead of checking if the folder exists

rocks-lib/src/rockspec/build/builtin.rs
vhyrro: You know what, I'm going to add a TODO to add behaviour for this, but for now I'd say we don't need it. It's a very complex problem to solve (even with all the helper crates for compiling lua), so much so that it'd suck up days (maybe weeks) of development time. Let's postpone this and work on other aspects of the codebase :)

vhyrro:
Waiting for the CI to complete and I believe we can merge :)

PR #26, opened 2024-06-30
feat: add `rocks list` command

vhyrro:
This PR sets up the necessary boilerplate to list out all available rocks. Listing outdated rocks requires more work and will be worked on separately.


rocks-bin/src/list.rs
```diff
@@ -0,0 +1,45 @@
...
+            .as_ref()
+            .ok_or_eyre("lua version not supplied!")?,
+    )?;
+    let available_rocks = tree.list();
+
+    // TODO(vhyrro): Add `outdated` support.
+
+    if list_data.porcelain {
+        for (name, versions) in available_rocks {
+            println!("{}: {}", name, versions.join(" "));
```
mrcjkb, line 29: suggestion (non-blocking): What if we ditch the luarocks "porcelain" output in favour of json?

mrcjkb (approved):

rocks-bin/src/list.rs
vhyrro: Love the sound of that idea. I'll keep the `--porcelain` name but change the internals to produce JSON instead.

rocks-bin/src/list.rs
```diff
@@ -0,0 +1,43 @@
+use clap::Args;
+use eyre::{OptionExt as _, Result};
+use itertools::Itertools as _;
+use rocks_lib::{config::Config, tree::Tree};
+use text_trees::{FormatCharacters, StringTreeNode, TreeFormatting};
+
+#[derive(Args)]
+pub struct List {
```
mrcjkb, line 8: suggestion: name this `ListCmd` to clarify that this isn't a list. 
Clap has macros that let you specify the name on the CLI

rocks-bin/src/list.rs
```diff
@@ -0,0 +1,43 @@
...
+    #[arg(long)]
+    outdated: bool,
+    #[arg(long)]
+    porcelain: bool,
+}
+
+pub fn list(list_data: List, config: &Config) -> Result<()> {
```
mrcjkb, line 15: suggestion:

```suggestion
pub fn list_installed(list_data: List, config: &Config) -> Result<()> {
```

rocks-bin/src/list.rs
vhyrro: Yeah good call, I'll change that.

PR #27, opened 2024-06-30
refactor!: produce JSON output when --porcelain is provided in `rocks search`

vhyrro:
This PR extends (and is based) on the changes made in #26. Running `rocks search --porcelain` will now also produce JSON output.

### DIFF
--- a/rocks-bin/src/search.rs
+++ b/rocks-bin/src/search.rs
@@ -1,3 +1,5 @@
+use std::collections::HashMap;
+
 use clap::Args;
 use eyre::Result;
 use itertools::Itertools;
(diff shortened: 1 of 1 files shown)


rocks-bin/src/search.rs
```diff
@@ -23,35 +25,36 @@ pub struct Search {
...
+        for (key, versions) in rock_to_version_map.into_iter().sorted() {
+            let mut tree = StringTreeNode::new(key.to_owned());
 
-                println!("{}", tree.to_string_with_format(&formatting)?);
+            for version in versions {
+                tree.push(version.to_owned());
             }
+
+            println!("{}", tree.to_string_with_format(&formatting).unwrap());
         }
```
vhyrro, line 58: note: this could be rewritten as a match statement for utmost clarity I guess, but I'm impartial to cosmetics like this. Thoughts are welcome.

mrcjkb (approved):

PR #29, opened 2024-07-02
refactor: move Luarock class into separate file

vhyrro:
This PR attempts to refactor the components of `rockspec/dependency.rs`.

Question: does `LuaDependency` also theoretically counts as a "luarock", but I've decided to keep that in the old `dependency.rs` file, which I feel makes sense. Feedback is welcome on this.

### DIFF
--- a/rocks-lib/src/lib.rs
+++ b/rocks-lib/src/lib.rs
@@ -1,5 +1,6 @@
 pub mod build;
 pub mod config;
+pub mod luarock;
 pub mod manifest;
 pub mod rocks;
 pub mod rockspec;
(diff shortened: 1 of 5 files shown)


mrcjkb:
> Question: does LuaDependency also theoretically counts as a "luarock", but I've decided to keep that in the old dependency.rs file, which I feel makes sense. Feedback is welcome on this.

The difference between a `LuaRock` and a `LuaDependency` is that the former has a `Version` and the latter has a `VersionReq` (constraints).

mrcjkb (approved):

PR #30, opened 2024-09-05
feat!: make CLI more ergonomic

vhyrro:
stacked on #31

This PR aims to refactor the CLI to be generally easier to use. Note: this is a breaking change and breaks compatibility with luarocks itself.

### DIFF
--- a/rocks-bin/src/list.rs
+++ b/rocks-bin/src/list.rs
@@ -7,8 +7,6 @@ use text_trees::{FormatCharacters, StringTreeNode, TreeFormatting};
 #[derive(Args)]
 pub struct ListCmd {
     #[arg(long)]
-    outdated: bool,
-    #[arg(long)]
     porcelain: bool,
 }
 
@@ -21,8 +19,6 @@ pub fn list_installed(list_data: ListCmd, config: &Config) -> Result<()> {
             .ok_or_eyre("lua version not supplied!")?,
     )?;
     let available_rocks = tree.list();
-
-    // TODO(vhyrro): Add `outdated` support.
 
     if list_data.porcelain {
         println!("{}", serde_json::to_string(&available_rocks)?);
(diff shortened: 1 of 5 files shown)


vhyrro:
woops, have to fix the commit history

vhyrro:
These are all the breakages worth making at this point in time I believe :p

rocks-bin/src/list.rs
```diff
@@ -6,8 +6,6 @@ use text_trees::{FormatCharacters, StringTreeNode, TreeFormatting};
 
 #[derive(Args)]
 pub struct ListCmd {
-    #[arg(long)]
-    outdated: bool,
```
mrcjkb, line 10: question: this is something we use in rocks.nvim.
Is there a plan for an equivalent?

mrcjkb (approved):

rocks-bin/src/list.rs
vhyrro: Yeah absolutely. I've been considering putting this under `rocks update --dry-run` or `rocks update --list-only`, but other ideas are welcome. Having this under "list" didn't make much sense to me.

vhyrro: Implemented in the form of a `rocks update --list` subcommand. I think that fits the bill very nicely. Happy to hear thoughts on that.

mrcjkb: I'm not sure how I feel about it being an `update --list` flag. 
npm, yarn and ruby gems all have an `outdated` subcommand. pip has a `list --outdated` flag.

An `update --dry-run` flag makes more sense to me, as the name of the flag implies that the update won't be applied.

vhyrro: The thing with `--dry-run` is that it doesn't hint at the fact that it'll print machine-readable output about updates. I think that having a dedicated `outdated` command is nice and descriptive. Maybe even `rocks debug outdated`? `list --outdated` is not very discoverable, I only discovered it was a thing when you told me a while back :p

mrcjkb: `outdated` sounds good to me.

PR #31, opened 2024-09-05
fix: make clippy happy

vhyrro:
This PR fixes clippy's complaints about invalid target OS checks. As it turns out, checking for cygwin is not a simple task, but is there any reason that we should check for cygwin explicitly?

### DIFF
--- a/rocks-lib/src/rockspec/platform.rs
+++ b/rocks-lib/src/rockspec/platform.rs
@@ -50,17 +50,17 @@ impl PartialOrd for PlatformIdentifier {
 
 /// Retrieves the target compilation platform and returns it as an identifier.
 pub fn get_platform() -> PlatformIdentifier {
-    if cfg!(linux) {
+    if cfg!(target_os = "linux") {
         PlatformIdentifier::Linux
-    } else if cfg!(macos) {
+    } else if cfg!(target_os = "macos") {
         PlatformIdentifier::MacOSX
-    } else if cfg!(cygwin) {
-        PlatformIdentifier::Cygwin
-    } else if cfg!(freebsd) {
+    } else if cfg!(target_os = "freebsd") {
         PlatformIdentifier::FreeBSD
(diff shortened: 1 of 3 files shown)


mrcjkb:
> is there any reason that we should check for cygwin explicitly?

That's one of the platforms you can specify in the rockspec's platform overrides. 
I'm not sure if luarocks has an implementation to check for it, but it allows you to override the platform in the config.
Our checks are at compile time, so if we can't check for cygwin as a target OS, we would have to do do that at runtime, or just add a config option too.

vhyrro:
Thanks for clarifying. I'll try to add a runtime check for this, then :)

vhyrro:
This is the most foolproof method I could conjure up from reading plenty of documentation. A thing to test: does the first if check (just checking for "linux") take priority over the next checks, or does cygwin advertise itself as "windows"?

mrcjkb:
> or does cygwin advertise itself as "windows"?

cygwin is just a set of tools, so the OS should still advertise itself as windows.

rocks-lib/src/rockspec/platform.rs
```diff
@@ -50,17 +50,17 @@ impl PartialOrd for PlatformIdentifier {
...
-    } else if cfg!(macos) {
+    } else if cfg!(target_os = "macos") {
         PlatformIdentifier::MacOSX
-    } else if cfg!(cygwin) {
-        PlatformIdentifier::Cygwin
-    } else if cfg!(freebsd) {
+    } else if cfg!(target_os = "freebsd") {
         PlatformIdentifier::FreeBSD
+    } else if which::which("cygpath").is_ok() {
+       PlatformIdentifier::Cygwin
```
mrcjkb, line 60:
```suggestion
        PlatformIdentifier::Cygwin 
```

mrcjkb (approved):

PR #33, opened 2024-09-08
feat!: add `rocks new` command

vhyrro:
This pull request removes the `write-rockspec` command and replaces it with the `rocks new` command, designed to create a new Lua project.

- Refactored the CLI experience
- Fall back to the current system username if a contributor name couldn't be inferred from a remote repo
- `project.rockspec` generation

Stacked on #30.


rocks-bin/src/project/write_new.rs
```diff
@@ -0,0 +1,301 @@
...
+    /// The project's name.
+    #[arg(long)]
+    name: Option<String>,
+
+    /// The description of the project.
+    #[arg(long)]
+    description: Option<String>,
+
+    /// The license of the project. Generic license names will be inferred.
+    #[arg(long, value_parser = clap_parse_license)]
```
mrcjkb, line 37: question: Does this fail if the license can't be parsed?

Which can be the case if

- the project has an esoteric/custom license 
- there's no license 
- there are multiple licenses 

If so, should we just fall back to `UNKNOWN` and print a warning?

rocks-bin/src/project/write_new.rs
```diff
@@ -0,0 +1,301 @@
...
+        .find_position(|&c| c != '-' && c != '_' && c != ',' && c.is_ascii_punctuation())
+    {
+        Err(format!("Unexpected punctuation '{}' found at column {}. Lists are comma separated but names should not contain punctuation!", char, pos))
+    } else {
+        Ok(input.split(',').map(|str| str.trim().to_string()).collect())
+    }
+}
+
+fn parse_license(input: &str) -> LicenseId {
+    spdx::imprecise_license_id(input).unwrap().0
```
mrcjkb, line 79: issue: This `unwrap` looks suspicious.
it's a private function, so I assume checks are performed before calling this? 
If so, I would suggest appending something like `_partial` to the name or marking this function as unsafe.

rocks-bin/src/project/write_new.rs
vhyrro: I took the KISS approach and allowed for the following:
- SPDX-recognized license
- `none` for no license

When no license is provided, the `project.rockspec` has an `*** enter license ***` field instead. I don't believe the rockspec format even allows for multiple licenses, but do correct me if I'm wrong. Implementing that shouldn't take long :)

mrcjkb: rockspecs don't allow for multiple licenses, but I've encountered projects where automatic license detection libraries failed because they detected more than one license. 
I presume this command would fail in such a case?

mrcjkb (approved):
rustfmt check is failing. 
otherwise, lgtm.

rocks-bin/src/project/write_new.rs
vhyrro: The way it's structured it won't fail when there's more than one license. I believe it'll just pick the first one it finds. We're retrieving this data from github and we're receiving the data in a struct from a crate so it's impossible for things to go haywire with variable length data :)

PR #35, opened 2024-09-10
feat: add `rocks install` command

vhyrro:
This PR adds a functioning `rocks install` command, which is just a combination of `unpack-remote` and `build`.

I've currently only managed to successfully install the `say` rock, because it only relies on lua as a dependency and uses `git+https://` as its source (the only two things we currently support :p). I guess adding more build possibilities should be next on the list.


rocks-bin/src/install.rs
```diff
@@ -0,0 +1,15 @@
+use eyre::Result;
+use rocks_lib::config::Config;
+
+#[derive(clap::Args)]
+pub struct Install {
+    /// Name of the rock to install.
+    name: String,
+    /// Rocks version to install.
+    version: Option<String>,
+}
```
mrcjkb, line 10: suggestion: I'm a fan of [parse, don't validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/).

Can we apply this here, by making the version a `semver::Version` and perhaps by creating a wrapper type for rock names that allow only valid characters?

It would require more involved arg parsing, but it results in nicer errors if users pass in invalid arguments.

Maybe this is for another PR, as it would mean refactoring the `Download` command, too.

rocks-lib/src/operations/install.rs
```diff
@@ -0,0 +1,38 @@
...
+        &rock_name,
+        rock_version.as_ref(),
+        Some(temp.path().to_path_buf()),
+        config,
+    )
+    .await?;
+
+    super::unpack_src_rock(temp.path().join(rock.path), Some(temp.path().to_path_buf()))?;
+
+    let rockspec = walkdir::WalkDir::new(&temp)
```
mrcjkb, line 23: suggestion: `let rockspec_path = ...`

mrcjkb (approved):

PR #42, opened 2024-09-11
feat: `rocks outdated` command

vhyrro:
This command will display outdated rocks in a nice tree-view.

### DIFF
--- a/rocks-bin/src/outdated.rs
+++ b/rocks-bin/src/outdated.rs
@@ -1,3 +1,5 @@
+use std::collections::HashMap;
+
 use clap::Args;
 use eyre::{OptionExt as _, Result};
 use itertools::Itertools;
@@ -14,7 +16,7 @@ pub struct Outdated {
     porcelain: bool,
 }
 
-pub async fn outdated(oudated_data: Outdated, config: &Config) -> Result<()> {
+pub async fn outdated(outdated_data: Outdated, config: &Config) -> Result<()> {
     let tree = Tree::new(
         &config.tree,
         config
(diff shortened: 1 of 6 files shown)


rocks-bin/src/outdated.rs
```diff
@@ -0,0 +1,76 @@
...
+    porcelain: bool,
+}
+
+pub async fn outdated(outdated_data: Outdated, config: &Config) -> Result<()> {
+    let tree = Tree::new(
+        &config.tree,
+        config
+            .lua_version
+            .as_ref()
+            .ok_or_eyre("lua version not supplied!")?,
```
mrcjkb, line 25: suggestion: extract to a method for `Config`.

mrcjkb, line 25: suggestion: add some help to the error message. Perhaps something like: `"lua version not set. You can supply it via '--lua-version' or set it in the config."`

mrcjkb (approved):

rocks-bin/src/outdated.rs
vhyrro: Do you mean a getter that returns an Option<&T>? If so, I agree.

vhyrro: Note: on second glance, there is some benefit to allowing the configuration to be modified at runtime, so I don't think we should provide read-only fields for the config. Thoughts?

mrcjkb: > Do you mean a getter that returns an Option<&T>? If so, I agree.

I meant a getter that returns a `Result`.

> Note: on second glance, there is some benefit to allowing the configuration to be modified at runtime, so I don't think we should provide read-only fields for the config. Thoughts?

Do you have anything concrete in mind? If not, I'd say yagni for now.

vhyrro:
Note: this no longer builds when rebased on top of the latest changes. However, all of the breakages here are amended in #45, so we should ideally merge both at the same time.

rocks-bin/src/outdated.rs
vhyrro: I was thinking that project detection could dynamically alter the `--tree` flag, but on second glance it can actually be done ahead of time in the default configuration, so you're right, YAGNI. However, I think that's fit for a separate PR, so I'll add a TODO comment and deal with that there :)

Issue #38, opened 2024-09-10
Stronger typed CLI (parse, don't validate)


mrcjkb:
suggestion: I'm a fan of [parse, don't validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/).

Can we apply this here, by making the version a `semver::Version` and perhaps by creating a wrapper type for rock names that allow only valid characters?

It would require more involved arg parsing, but it results in nicer errors if users pass in invalid arguments.

Maybe this is for another PR, as it would mean refactoring the `Download` command, too.

_Originally posted by @mrcjkb in https://github.com/nvim-neorocks/rocks/pull/35#discussion_r1751785967_

PR #43, opened 2024-09-11
refactor: introduce `PackageName` for luarocks package names

mrcjkb:
An initial refactor in preparation for #38


vhyrro (approved):

vhyrro:
Maybe we should call `LuaRock` LuaPackage instead 🤔

PR #45, opened 2024-09-11
feat: `rocks install-lua` / Lua compilation

vhyrro:
stacked on #42

TODOs:
- [x] Compilation of LuaJIT (requires one more crate)
- [x] Usage in build code (but only when it's required)
- [x] Proper detection of already installed Lua headers.
- [x] Customizable header location?

### DIFF
--- a/rocks-lib/Cargo.toml
+++ b/rocks-lib/Cargo.toml
@@ -36,8 +36,8 @@ lets_find_up = "0.0.3"
 remove_dir_all = "0.8.3"
 openssl = "0.10.66"
 lua-src = "547.0.0"
-target-lexicon = "0.12.16"
 luajit-src = "210.5.10"
+target-lexicon = "0.12.16"
 
 [dev-dependencies]
 httptest = { version = "0.16.1" }
--- a/rocks-lib/Cargo.toml
+++ b/rocks-lib/Cargo.toml
@@ -35,6 +35,9 @@ which = "6.0.3"
 lets_find_up = "0.0.3"
 remove_dir_all = "0.8.3"
 openssl = "0.10.66"
+lua-src = "547.0.0"
+target-lexicon = "0.12.16"
+luajit-src = "210.5.10"
 
 [dev-dependencies]
 httptest = { version = "0.16.1" }
--- a/rocks-lib/src/config/mod.rs
+++ b/rocks-lib/src/config/mod.rs
@@ -9,6 +9,7 @@ pub enum LuaVersion {
     Lua53,
     Lua54,
     LuaJIT,
+    LuaJIT52,
     // TODO(vhyrro): Support luau?
     // LuaU,
 }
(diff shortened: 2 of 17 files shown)


rocks-lib/Cargo.toml
```diff
@@ -32,6 +32,8 @@ zip = "0.6.6"
 insta = { version = "1.39.0", features = ["redactions", "yaml"] }
 which = "6.0.3"
 lets_find_up = "0.0.3"
+lua-src = "547.0.0"
```
max397574, line 35: 😳

rocks-lib/src/config/mod.rs
```diff
@@ -145,7 +166,7 @@ impl Default for Config {
             only_server: None,
             only_sources: None,
             namespace: "".into(),
-            lua_dir: None,
+            lua_dir: Config::get_default_cache_path().unwrap().join("lua"),
```
mrcjkb, line 169: note(non-blocking): I'm not sure if a cache is the right location to install lua.
Perhaps a data directory (e.g. in `get_default_data_path()`) would be more suitable?

rocks-lib/src/lua/mod.rs
```diff
@@ -0,0 +1,70 @@
+use eyre::Result;
+use std::path::PathBuf;
+use target_lexicon::Triple;
+
+use crate::config::{Config, LuaVersion};
+
+pub struct Lua {
+    pub include_dir: PathBuf,
+    pub lib_dir: PathBuf,
+}
```
mrcjkb, line 10: suggestion: This might become confusing, as we also have `mlua::Lua`.

What if we call this `LuaLib` and the module `lua_lib`?

rocks-lib/src/lua/mod.rs
```diff
@@ -0,0 +1,70 @@
...
+            let target = &host.to_string();
+            let host_operating_system = &host.operating_system.to_string();
+
+            let (include_dir, lib_dir) = match version {
+                LuaVersion::LuaJIT => {
+                    let build = luajit_src::Build::new()
+                        .target(target)
+                        .host(host_operating_system)
+                        .out_dir(output)
+                        .lua52compat(true)
```
mrcjkb, line 32: question: Should this be configurable?

rocks-lib/src/lua/mod.rs
vhyrro: Maybe `LuaInstallation` for verbosity's sake? `LuaLib` is still a tad ambiguous I'd say.

rocks-lib/src/lua/mod.rs
vhyrro: Yeah it should, but that'll require a second `LuaJIT52` version and a separate `--lua-version jit52|luajit52` flag. I'll get going with that then :)

vhyrro:
Further question: should `install-lua` use the `--lua-version` flag instead of requiring a separate positional argument? Therefore `rocks install-lua 5.1` would become `rocks --lua-version 5.1 install-lua`.

This has the added benefit of "smart inference" when in e.g. a lua project. When in a one, running just `rocks install-lua` will infer the lua version to pull for that specific project. Happy to hear feedback on that too.

max397574:
sounds really reasonable
imo if something is optional a flag makes much more sense than a positional argument

rocks-lib/src/lua/mod.rs
vhyrro: Changed it to `lua_installation` but happy to shorten it down to e.g. `lua_install` if you prefer.
I don't mind longer names when they're nicely unambiguous like in this case.

vhyrro:
Also addressed that, now `--lua-version` will be used instead.

mrcjkb (approved):

mrcjkb:
suggestion: If we want to make `rocks-lib` embeddable, we might want to consider putting each lua version behind a feature flag (`rocks-bin` would enable all features by default).
Could be done in a follow-up PR.

Thoughts?

PR #47, opened 2024-09-11
refactor: `LuaDependency` -> `LuaPackageReq`

mrcjkb:
This PR

- Moves the `LuaDependency` type to `lua_package::LuaPackageReq`, which we can later use for arguments to commands like `install`, `search` and `download`.
- Moves the `platform` related trait implementations of the type to the `platform` module (where the trait is defined, so we don't have an orphan instance).


max397574:
Perhaps this is just me
but the first thing that came to my mind when reading LuaPackageReq was Request
and only after I saw that it was renamed from Dependency I realized that it's Requirement

I'm wondering
what is the reasoning behind this rename?
To me requirement sounds like something more general then a dependency. E.g. a certain system requirement
which afaict isn't the case here

mrcjkb:
> Perhaps this is just me but the first thing that came to my mind when reading LuaPackageReq was Request and only after I saw that it was renamed from Dependency I realized that it's Requirement
> 
> I'm wondering what is the reasoning behind this rename? To me requirement sounds like something more general then a dependency. E.g. a certain system requirement which afaict isn't the case here

I cargo culted the `Req` suffix from `semver::VersiomReq`.
The more general name is because I intend to use it in commands like `rocks [search|download|install|...]`.
For example, `rocks install foo <2.0.0` would install the latest version of `foo` that is less than 2.0.0`.
In that case, `LuaPackageReq` is not a dependency.

max397574 (approved):

PR #55, opened 2024-09-12
ci: don't run release build on PR + run on mac

mrcjkb:
Closes #53 

Nix by default builds rust packages with the `release` profile.
We don't need this for running tests on PR.

This PR introduces the following changes:

- Build with the `release` profile on push to `master`.
- Run checks (including checks) with the `debug` profile on PR.
- Add builds on darwin.
- Add `libgit2` to the `buildInputs` (required on darwin).

PR #58, opened 2024-09-12
feat: support luarocks dev versions

### LINKED ISSUE #48: [Feature] support luarocks dev versions
with the upcoming changes in #38, due to the use of `semver::VersionReq`, we won't support `dev`, `scm` or `git` as a version.

We should provide our own wrapper types for `Version` and `VersionReq` that add support (perhaps they should be enums).

mrcjkb:
Stacked on #55.
Closes #48.

This adds wrapper types around `semver::Version` and `semver::VersionReq` that add support for luarocks `dev` versions.

### DIFF
--- a/rocks-lib/src/lua_package/version.rs
+++ b/rocks-lib/src/lua_package/version.rs
@@ -1,15 +1,134 @@
+use std::{cmp::Ordering, fmt::Display, str::FromStr};
+
 use eyre::Result;
 use html_escape::decode_html_entities;
 use semver::{Error, Version, VersionReq};
 
+/// **SemVer version** as defined by <https://semver.org>.
+/// or a **Dev** version, which can be one of "dev", "scm", or "git"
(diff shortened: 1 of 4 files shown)


rocks-lib/src/lua_package/version.rs
```diff
@@ -1,15 +1,133 @@
...
+        }
+    }
+}
+
+impl FromStr for PackageVersionReq {
+    type Err = eyre::Error;
+
+    fn from_str(text: &str) -> Result<Self, Self::Err> {
+        let regex: Regex = Regex::new(r"^==\s*(dev|git|scm)|\b(dev|git|scm)\b$").unwrap();
+        if regex.is_match(text) {
```
vhyrro, line 112: suggestion: same here, a collection of `trim_start_matches`, `trim` and `matches!()` could probably supersede invoking an entire regex engine here, unless I'm misinterpreting certain edge cases of this regex.

vhyrro (approved):

PR #61, opened 2024-09-13
feat(cli): version requirements as arguments

### LINKED ISSUE #38: Stronger typed CLI (parse, don't validate)

### LINKED ISSUE #49: [Feature] version constraints for install, download and search commands
Related to #38

mrcjkb:
Closes #38 
Closes #49 

- Uses `LuaPackageReq` as the CLI args.
- Changes `ManifestMetadata` to have `PackageVersion`s as keys, instead of `String`s.


rocks-lib/src/luarock/mod.rs
mrcjkb, line 1: The `luarock` module seems to have been re-added accidentally during a rebase.

rocks-lib/src/manifest/metadata.rs
```diff
@@ -23,11 +28,12 @@ impl ManifestMetadata {
     pub fn new(manifest: &String) -> Result<Self> {
         let lua = Lua::new();
 
-        lua.load(manifest).exec()?;
+        lua.load(dbg!(manifest)).exec()?;
```
vhyrro, line 31: issue: leftover debug call 😅

rocks-lib/src/manifest/metadata.rs
```diff
@@ -42,19 +48,71 @@ impl ManifestMetadata {
...
+            lua_package_req.name().to_owned(),
+            version.to_owned(),
+        ))
+    }
+}
+
+#[derive(serde::Deserialize)]
+pub struct ManifestRockEntry {
+    /// e.g. "linux-x86_64", "rockspec", "src", ...
+    pub arch: String,
```
vhyrro, line 88: question: we'd probably benefit from making this an enum too, wouldn't we? Should this be given a TODO for the future? Note that the `Triple` type can be used to represent a "linux-arch" string quite nicely (with serialization and such).

rocks-lib/src/lua_package/mod.rs
```diff
@@ -14,11 +14,14 @@ pub struct LuaPackage {
...
-        })
+    pub fn new(name: PackageName, version: PackageVersion) -> Self {
+        Self { name, version }
+    }
+    pub fn parse(name: String, version: String) -> Result<Self> {
+        Ok(Self::new(
+            PackageName::new(name),
+            PackageVersion::parse(&version)?,
+        ))
     }
```
vhyrro, line 25: question: any benefit from doing `impl From<(String, String)>` as a bonus, or is that too much "magic"?

vhyrro (commented):
Big fan, makes the API so much more robust now :)

vhyrro:
Oh, there's also the `Rockspec` which still has a `String` as its version object. Should that be addressed in this PR as well?

rocks-lib/src/manifest/metadata.rs
mrcjkb: Created #62

rocks-lib/src/lua_package/mod.rs
mrcjkb: The implementation would have to panic if the version can't be parsed, as it doesn't return a `Result`.

mrcjkb: We have a `FromStr`, so I'd say yagni to `From` for now.

mrcjkb:
> Oh, there's also the `Rockspec` which still has a `String` as its version object. Should that be addressed in this PR as well?

Done. See the second commit.

PR #64, opened 2024-09-14
feat: project-local rocktrees

vhyrro:
This is a draft implementation of project-local rock trees. If `rocks` detects it's in a project, it will always interact with the project-local rock tree first. Pass `--no-project` to prevent these interactions.

### DIFF
--- a/rocks-bin/src/list.rs
+++ b/rocks-bin/src/list.rs
@@ -10,13 +10,10 @@ pub struct ListCmd {
     porcelain: bool,
 }
 
-pub fn list_installed(list_data: ListCmd, config: &Config) -> Result<()> {
+pub fn list_installed(list_data: ListCmd, config: Config) -> Result<()> {
     let tree = Tree::new(
-        &config.tree,
-        config
-            .lua_version
-            .as_ref()
-            .ok_or_eyre("lua version not supplied!")?,
+        config.tree,
+        config.lua_version.ok_or_eyre("lua version not supplied!")?,
     )?;
     let available_rocks = tree.list();
 
--- a/rocks-bin/src/list.rs
+++ b/rocks-bin/src/list.rs
@@ -12,8 +12,11 @@ pub struct ListCmd {
 
 pub fn list_installed(list_data: ListCmd, config: Config) -> Result<()> {
     let tree = Tree::new(
-        config.tree,
-        config.lua_version.ok_or_eyre("lua version not supplied!")?,
+        config.tree().clone(),
+        config
+            .lua_version()
+            .cloned()
+            .ok_or_eyre("lua version not supplied!")?,
     )?;
     let available_rocks = tree.list();
 
--- a/rocks-lib/src/build/mod.rs
+++ b/rocks-lib/src/build/mod.rs
@@ -4,7 +4,7 @@ use crate::{
     rockspec::{utils, Build as _, BuildBackendSpec, RockSourceSpec, Rockspec},
     tree::{RockLayout, Tree},
 };
-use eyre::{OptionExt as _, Result};
+use eyre::Result;
 use git2::Repository;
 
 mod builtin;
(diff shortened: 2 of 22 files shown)


rocks-bin/src/list.rs
```diff
@@ -10,13 +10,10 @@ pub struct ListCmd {
     porcelain: bool,
 }
 
-pub fn list_installed(list_data: ListCmd, config: &Config) -> Result<()> {
+pub fn list_installed(list_data: ListCmd, config: Config) -> Result<()> {
```
vhyrro, line 13: note: I'll try to remove references to the config in all other invocations in `main.rs` for consistency.

rocks-lib/src/config/mod.rs
```diff
@@ -33,6 +35,32 @@ impl FromStr for LuaVersion {
...
+        ] {
+            if lua.version_req().matches(&possibility.parse().unwrap()) {
+                return Ok(version);
+            }
+        }
+
+        Err(eyre!("no valid matches for `lua` found!"))
+    }
+}
+
```
vhyrro, line 63: question: I'm completely torn on this. Should the `Rockspec` have a `.lua_version()` getter, or should we keep it like this? This approach allows the use of `rockspec.try_into()` as well as `let version = LuaVersion::from(&rockspec);`.

I feel like this is a bit more fitting as it performs a direct conversion into a `LuaVersion`, but I know of no best practices for these sorts of decisions.

rocks-lib/src/config/mod.rs
```diff
@@ -173,8 +185,16 @@ impl Default for Config {
...
-            lua_version: None,
-            tree: data_path,
+            lua_version: current_project
+                .as_ref()
+                .map(|project| LuaVersion::try_from(project.rockspec()))
+                .transpose()
+                .unwrap(),
+            tree: current_project
+                .as_ref()
+                .map(|project| project.root().join("tree"))
```
vhyrro, line 195: issue: this needs to be moved out into a different place. You can't test for `--no-project` in time here, as it'll always be `false` in the default configuration.

rocks-lib/src/config/mod.rs
mrcjkb: From the docs, [When to implement `From`](https://doc.rust-lang.org/std/convert/trait.From.html#when-to-implement-from):

> - The conversion is lossless: semantically, it should not lose or discard information.
> - The conversion is value-preserving: the conceptual kind and meaning of the resulting value is the same, even though the Rust type and technical representation might be different.
> - The conversion is obvious: it’s the only reasonable conversion between the two types. Otherwise it’s better to have it be a named method or constructor

So I'd argue that a getter is the better choice here. 

Also, I'd suggest a getter that returns an `Option` instead of a `Result`, as some rockspecs don't have a `dependencies` table.

vhyrro:
Some of this code is not really clean, I'll be honest. The next step will be to finish off the project functionality (with even more hacky code), write tests for everything and then start aggressively refactoring :)

rocks-lib/src/build/mod.rs
```diff
@@ -62,11 +62,13 @@ pub fn build(rockspec: Rockspec, config: &Config) -> Result<()> {
     // operations in the temporary directory itself and then copy all results over once they've
     // succeeded.
 
-    let lua_version = LuaVersion::try_from(&rockspec)?;
+    // TODO(vhyrro): Create a unified way of accessing the Lua version with centralized error
+    // handling.
+    let lua_version = config.lua_version.clone().or(rockspec.lua_version()).ok_or_eyre("lua version not set! Please provide a version through `--lua-version <ver>`")?;
```
mrcjkb, line 67: question: what if the rockspec's lua version is incompatible with the config's lua version? 
Would it be better if the rockspec took precedence here?

vhyrro: I think conflicts should result in an error. If a user explicitly requires Lua 5.1, but their project is in conflict, then allowing it could only cause problems down the line (think about executing in scripts, for example, where a user downloads a 5.1 package only to find out it has ended up in 5.3 instead). Supplying a Lua version is an explicit action, so we're doing the user a favour by letting them know about conflicts.

mrcjkb: right, I was thinking of how luarocks lets you specify the default lua version in a global config file.

vhyrro: In the case of a global config then these shouldn't clash, you're right. I've been considering calling the future config value `preferred_lua_version` for this reason :)

vhyrro:
Okay I think everything is done for this PR.

mrcjkb (approved):

PR #74, opened 2024-09-16
feat: automatic dependency installation

vhyrro:
This PR automatically installs the dependencies of a given project.

TODOs:
- [x] Don't install dependency if it's already present

### DIFF
--- a/rocks-lib/src/build/mod.rs
+++ b/rocks-lib/src/build/mod.rs
@@ -4,6 +4,7 @@ use crate::{
     rockspec::{utils, Build as _, BuildBackendSpec, RockSourceSpec, Rockspec},
     tree::{RockLayout, Tree},
 };
+use async_recursion::async_recursion;
 use eyre::{OptionExt as _, Result};
 use git2::Repository;
 
@@ -32,7 +33,32 @@ fn install(
     Ok(())
 }
 
-pub fn build(rockspec: Rockspec, config: &Config) -> Result<()> {
+#[async_recursion]
+pub async fn build(rockspec: Rockspec, config: &Config) -> Result<()> {
+    // TODO(vhyrro): Create a unified way of accessing the Lua version with centralized error
+    // handling.
+    let lua_version = rockspec.lua_version();
+    let lua_version = config.lua_version().or(lua_version.as_ref()).ok_or_eyre(
+        "lua version not set! Please provide a version through `--lua-version <ver>`",
+    )?;
+
+    let tree = Tree::new(config.tree().clone(), lua_version.clone())?;
+
+    // Recursively build all dependencies.
+    // TODO: Handle build dependencies as well.
+    for dependency_req in rockspec
+        .build_dependencies
+        .current_platform()
+        .iter()
+        .filter(|req| tree.has_rock(req).is_none())
+    {
+        // NOTE: This recursive operation will create another `tree` object
+        // which will in turn acquire another lock to the filesystem.
+        // Once we implement fs locks, this could become a problem, so a more sophisticated
+        // filesystem acquiring mechanism will have to be devised.
+        crate::operations::install(dependency_req.clone(), config).await?;
+    }
+
     // TODO(vhyrro): Use a more serious isolation strategy here.
     let temp_dir = tempdir::TempDir::new(&rockspec.package.to_string())?;
 
@@ -62,18 +88,6 @@ pub fn build(rockspec: Rockspec, config: &Config) -> Result<()> {
     // operations in the temporary directory itself and then copy all results over once they've
     // succeeded.
 
-    // TODO(vhyrro): Create a unified way of accessing the Lua version with centralized error
-    // handling.
-    let rockspec_version = rockspec.lua_version();
-    let lua_version = config
-        .lua_version()
-        .or(rockspec_version.as_ref())
-        .ok_or_eyre(
-            "lua version not set! Please provide a version through `--lua-version <ver>`",
-        )?;
-
-    let tree = Tree::new(config.tree().clone(), lua_version.clone())?;
-
     let output_paths = tree.rock(&rockspec.package, &rockspec.version)?;
 
     let lua = LuaInstallation::new(lua_version, config)?;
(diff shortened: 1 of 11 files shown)


mrcjkb:
> Don't install dependency if it's already present

We should also think about what to do with dev dependencies, since they might be outdated and incompatible. I'd suggest to always reinstall dev dependencies.

For semver dependencies, we might want to check if there's a newer version that satisfies the version bounds and update to it, to benefit from potential bug fixes.

vhyrro:
> > Don't install dependency if it's already present
> 
> We should also think about what to do with dev dependencies, since they might be outdated and incompatible. I'd suggest to always reinstall dev dependencies.
> 
> For semver dependencies, we might want to check if there's a newer version that satisfies the version bounds and update to it, to benefit from potential bug fixes.

I can't think of a context in which a freshly installed rock will break as a result of outdated dev dependencies. I believe `rocks update` is the thing that should do all the updating magic :)

vhyrro:
Oh i misunderstood. I thought you meant actual dev_dependencies, but you're referring to `dev` versions. The lockfile will store their pinned state, won't it?

rocks-lib/src/build/mod.rs
```diff
@@ -32,7 +33,35 @@ fn install(
...
 
-pub fn build(rockspec: Rockspec, config: &Config) -> Result<()> {
+#[async_recursion]
+pub async fn build(rockspec: Rockspec, config: &Config) -> Result<()> {
+    // TODO(vhyrro): Create a unified way of accessing the Lua version with centralized error
+    // handling.
+    let rockspec_version = rockspec.lua_version();
+    let lua_version = config
+        .lua_version()
+        .or(rockspec_version.as_ref())
```
max397574, line 43: kinda confusing variable name
at first thought that this could be like different versions of the rockspec spec

perhaps just inline?

max397574 (approved):
haven't tested but code looks rly good to me

rocks-lib/src/build/mod.rs
vhyrro: You're right. That variable name is a brainfart, I'll change that. It can't be inlined else the compiler will complain that a temporary value is freed immediately upon creation.

mrcjkb:
> Oh i misunderstood. I thought you meant actual dev_dependencies (a `rocks` feature for dependencies while developing a project), but you're referring to `dev` versions. The lockfile will store their pinned state, won't it? Cargo, for instance, doesn't touch git dependencies, just stores the commit hash in the lockfile and calls it a day.

We can/should store the pinned state in the lockfile, but what if someone has force-uploaded a rockspec?
We could use the lockfile (rockspec's sha256sum) to determine if we need to reinstall a package.

rocks-lib/src/build/mod.rs
vhyrro: Done

rocks-lib/src/build/mod.rs
```diff
@@ -32,7 +33,35 @@ fn install(
...
+    for dependency_req in rockspec
+        .build_dependencies
+        .current_platform()
+        .iter()
+        .filter(|req| tree.has_rock(req).is_none())
+    {
+        // NOTE: This recursive operation will create another `tree` object
+        // which will in turn acquire another lock to the filesystem.
+        // Once we implement fs locks, this could become a problem, so a more sophisticated
+        // filesystem acquiring mechanism will have to be devised.
```
mrcjkb, line 61: question: a more sophisticated fs acquiring mechanism than what?
Afaik, we don't acquire anything yet?

mrcjkb (approved):

mrcjkb:
rustfmt.

> [!TIP]
>
> Sadly, jj doesn't support pre-commit or pre-push hooks.
> Before pushing, you can run `pre-commit run --all-files`.

rocks-lib/src/build/mod.rs
vhyrro: Forgive my 11pm brain for that one. While we don't acquire any "locks", there will be many tree objects and that could very quickly become a problem with concurrent writes. In this case it's fine, because we always await each installation operation to finish before starting the next one.

rocks-lib/src/build/mod.rs
```diff
@@ -32,7 +33,32 @@ fn install(
...
+    // TODO(vhyrro): Create a unified way of accessing the Lua version with centralized error
+    // handling.
+    let lua_version = rockspec.lua_version();
+    let lua_version = config.lua_version().or(lua_version.as_ref()).ok_or_eyre(
+        "lua version not set! Please provide a version through `--lua-version <ver>`",
+    )?;
+
+    let tree = Tree::new(config.tree().clone(), lua_version.clone())?;
+
+    // Recursively build all dependencies.
```
mrcjkb, line 47: thought: we're going to have to implement an equivalent to the luarocks loader to support multiple versions of the same dependency.
And for people who don't have the loader enabled, we might want to consider adding a flag that disallows conflicting dependencies. Which would mean we'll need a more sophisticated dependency resolution.

mrcjkb: creating an issue for this so that we can get this merged

rocks-lib/src/tree/list.rs
```diff
@@ -22,30 +22,32 @@ impl Tree {
                     .unwrap()
                     .split_once('@')
                     .unwrap();
-                (name.to_string(), version.to_string())
+                (
+                    PackageName::new(name.to_string()),
+                    PackageVersion::parse(version)
+                        .expect("invalid version found in rocktree. This is a bug!"),
```
mrcjkb, line 28: suggestion: include the path to the error message

rocks-lib/src/build/mod.rs
vhyrro: This could be a bit of metadata that is read from the tree's lockfile in the future (e.g. a dependency or rock could have an `only` or `singleton` flag attached). Good point though.

rocks-lib/src/tree/list.rs
vhyrro: All of these errors are currently placeholders, either serving as a "please implement a serious error handling strategy here in the future" or "issue a nice `ariadne` error here so the user isn't confused".

I'm banking on the fact that since we'll have a lot of tests we can aggressively refactor for better error messages and less `clone()` invocations without fearing any breakages once the project hits its MVP stage. I'll add the path here for clarity though :)

mrcjkb (approved):

PR #78, opened 2024-09-19
feat: `make` build type

vhyrro:
This PR adds the `make` build type in rockspecs.

This is a draft as I haven't tested it whatsoever on real world rocks :)

PR #80, opened 2024-09-20
fix: version and rock name parsing errors (#79)

### LINKED ISSUE #79: [Bug]: Latest version fetching is borked
Some recent changes have caused `rocks install <package>` (without a version constraint) to fail:
> thread 'main' panicked at rocks-lib/src/operations/download.rs:33:54:
called `Option::unwrap()` on a `None` value

vhyrro:
This PR was supposed to be much longer, but I decided that we were layering on too much complexity with no benefit whatsoever.

This PR fixes #79 in two ways: firstly, it fixes the broken version checking code (and now defaults to the latest version correctly). Secondly, it permits rock names to now also have `.`s in their names.
Instead of trying to create a generic catch-all parser, I think restricting rocks to this format is both sane and expected. During the big error rewrite, we can issue nice error messages to the user so they're never confused.

### DIFF
--- a/rocks-lib/src/lua_package/mod.rs
+++ b/rocks-lib/src/lua_package/mod.rs
@@ -8,7 +8,6 @@ pub use version::{PackageVersion, PackageVersionReq};
 mod outdated;
 mod version;
 
-// TODO: We probably need a better name for this
 pub struct LuaPackage {
     name: PackageName,
     version: PackageVersion,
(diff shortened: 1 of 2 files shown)


vhyrro:
I'll rebase this onto master instead as #78 is still a WIP

mrcjkb:
looks like a rebase gone wrong? 
I don't see any changes.

PR #83, opened 2024-09-21
feat: implement `rocks purge` and `rocks remove`

vhyrro:
This PR adds basic support for `rocks purge` and `rocks remove`.

There are no tests for this yet, I'll get those set up tomorrow.

### DIFF
--- a/rocks-bin/src/purge.rs
+++ b/rocks-bin/src/purge.rs
@@ -1,4 +1,5 @@
 use eyre::Result;
+use inquire::Confirm;
 use rocks_lib::{config::Config, tree::Tree};
 
 pub fn purge(config: Config) -> Result<()> {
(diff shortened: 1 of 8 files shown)


rocks-bin/src/purge.rs
```diff
@@ -0,0 +1,13 @@
+use eyre::Result;
+use rocks_lib::{config::Config, tree::Tree};
+
+pub fn purge(config: Config) -> Result<()> {
```
mrcjkb, line 4: question: would it make sense to prompt for confirmation by default, and add a `--skip-prompt`  flag?

rocks-lib/src/tree/mod.rs
```diff
@@ -50,6 +50,10 @@ impl Tree {
         self.root.join(self.version.to_string())
     }
 
+    pub fn root_for(&self, rock_name: &PackageName, rock_version: &PackageVersion) -> PathBuf {
```
mrcjkb, line 53: suggestion: why not take a `&LuaPackage` as an argument?

vhyrro: That's to keep consistency with the `rock()` function, which also takes a name and version. Those should be refactored to just use `LuaPackage` eventually though.

mrcjkb (approved):

vhyrro:
Will merge this after #94

PR #87, opened 2024-09-22
ci: add cargo-upgrade action

mrcjkb:


vhyrro:
question: this won't upgrade to new major releases, right?

mrcjkb:
> question: this won't upgrade to new major releases, right?

No idea. It opens a PR, so I don't think it's a problem if it does.

vhyrro (approved):

PR #96, opened 2024-09-27
feat(install): implement rockspec URLs

mrcjkb:
This adds support for rockspecs with `.zip` or `.tar`/`.tar.gz` archives in their source URLs.

I've tested it manually with haskell-tools.nvim (which has a zip archive in the `source.url`).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -322,6 +322,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "cfb"
+version = "0.7.3"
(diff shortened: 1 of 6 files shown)


vhyrro (approved):

vhyrro:
Damn our dependency count is quite crazy 😄

mrcjkb:
Our binary is already almost 3 times as big as nvim :sweat_smile:

PR #105, opened 2024-09-28
feat: progress spinners

### LINKED ISSUE #66: [refactor]: Move to `indicatif`
[indicatif](https://github.com/console-rs/indicatif) is a library for displaying nice looking progress bars for various tasks. We should incorporate this for things like manifest pulling, file downloading and building.

mrcjkb:
Closes #66.

Here's an example of the output:

scrot

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1397,6 +1397,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "indicatif"
+version = "0.17.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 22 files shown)


mrcjkb:
:thinking: this implementation has some strange behaviour.
The currently active spinner will alternate between the "installing <rock> " text and the text it's supposed to be displaying.
Not sure if this is a bug in indicatif, or if it's not supposed to be used in this way.

mrcjkb:
It works nicely if we pass a `&MultiProgress` around.

vhyrro (approved):

PR #109, opened 2024-10-01
fix(version): trim specrev when parsing

mrcjkb:
specrev is likely to lead to all sorts of issues.
Let's not support it (at least for now).


mrcjkb:
Self-merging, as this is trivial.

mrcjkb:
Build failure looks like a flake. #78, which is stacked on this, is passing.

vhyrro (approved):

PR #111, opened 2024-10-03
feat: lockfile support

vhyrro:
This pull request adds long-awaited lockfile support. While not fully rigorous yet, I've tried to have it handle as many cases as I could. I've improved error handling in a few places, made `tree.list()` use the lockfile as its source of truth and added the whole hashing logic.

Things to do after this PR (I'll make these into issues):
- `rocks lock` commands
- Pinning support (`rocks install --pin` and `rocks pin`)


vhyrro:
Tests are failing because of serialization order... I wonder how to fix that.

vhyrro:
Fully ready for review now :)

rocks-lib/src/lockfile/mod.rs
```diff
@@ -0,0 +1,199 @@
...
+
+use eyre::{bail, Result};
+use itertools::Itertools;
+use serde::{Deserialize, Serialize};
+use sha2::{Digest, Sha256};
+
+use crate::lua_package::{LuaPackage, PackageName, PackageVersion};
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct LockedPackage {
```
mrcjkb, line 12: question: if we want this to work with nix, we will need to hash the source content (archive/repo/..
) and the rockspec. 

Is this still planned? It would likely require us to change where we add a rock to the lockfile (to after fetching the sources).

rocks-lib/src/lockfile/mod.rs
```diff
@@ -0,0 +1,199 @@
...
+
+        hex::encode(hasher.finalize())
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Lockfile {
+    #[serde(skip_serializing, skip_deserializing)]
+    filepath: PathBuf,
+    // TODO: Serialize this directly into a `Version`
```
mrcjkb, line 47: suggestion: todo comments are easy to forget. Better to create an issue if it's too much for this PR.

mrcjkb (approved):

mrcjkb:
That `tree_list` test again 😢

rocks-lib/src/lockfile/mod.rs
vhyrro: I always go through my codebase with todo-comments.nvim, so I rarely miss them. But sure, adding an issue sounds like an idea.

rocks-lib/src/lockfile/mod.rs
vhyrro: Right, that's interesting. I believe we'll want to create a separate hash for this, though, but I'm not sure how the complexity maps over to Nix.
The basic ID is useful because other parts of the application can construct the unique ID just from the name, version and pinning information.

If you'd like to check if a rock exists, you don't want to have to read the entire source content, nor do you want to pull down the rockspec and hash that as well, just to look up the rock in question.

I do think that each rock should also have a `sources_hash` or something similar, which is a hash of the rockspec and the source content. The rockspec already contains version and name information, and pinning information isn't important in a source hash. Wdyt?

vhyrro:
Fixed conflicts but am unable to figure out how on earth to fix the tree_list failure. It's supposed to work since I've enabled deep sorting through the redactions, but insta still doesn't seem to care :(

rocks-lib/src/lockfile/mod.rs
mrcjkb: I don't think we need it for lookups. But it's useful for integrity checks, in case someone force-uploaded a modified rockspec or source archive (which defeats the purpose of locking and is a supply chain vulnerability).
The benefit of this for Nix is that it can use a cached version if the source or rockspec hash has changed. We could one day implement something similar.
And if there's nothing in the cache and a mismatched rockspec/source hash, it would be reported as an error (with an option to update the lock entry, warning that it could be a supply chain attack).

rocks-lib/src/lockfile/mod.rs
mrcjkb: A benefit of issues is that sometimes an external contributor will open a PR for it.

mrcjkb:
> unable to figure out how on earth to fix the tree_list failure. It's supposed to work since I've enabled de

`lockfile.rocks()` returns a `HashMap`, so there's no ordering.

mrcjkb:
Fixed it. `sorted_redaction` doesn't work because we have a nested `Vec`, where the `Vec<PackageVersion>` isn't sorted.

rocks-lib/src/lockfile/mod.rs
mrcjkb: Created #116, so we can merge this first.

vhyrro:
Thank you for taking care of these things, I was busy with other stuff. Appreciate it :)

PR #119, opened 2024-10-06
feat!: major refactors to internal data structures

vhyrro:


rocks-bin/src/outdated.rs
```diff
@@ -34,11 +34,12 @@ pub async fn outdated(outdated_data: Outdated, config: Config) -> Result<()> {
     // However, this should also take into account dependency constraints made by other rocks.
     // This will naturally occur with lockfiles and should be accounted for directly in the
     // `has_update` function.
-    let rock_list = tree
-        .as_rock_list()?
-        .into_iter()
+    let rock_list = tree.as_rock_list()?;
+    let rock_list = rock_list
+        .iter()
```
vhyrro, line 39: heads up: now required as a temporary value is created in the process

rocks-bin/src/remove.rs
```diff
@@ -20,13 +21,28 @@ pub async fn remove(remove_args: Remove, config: Config) -> Result<()> {
...
+        &RemotePackage::new(remove_args.name.clone(), target_version.clone()).into_package_req(),
+    ) {
+        Some(package) => {
+            rocks_lib::operations::remove(&MultiProgress::new(), package, &config).await
+        }
+        None => {
+            eprintln!("Could not find {}@{}", remove_args.name, target_version);
+            Ok(())
+        }
+    }
```
vhyrro, line 47: The API now distinguishes physical rocks on the filesystem from remote rocks. This distinction now forces code from `rocks remove` to verify that the rock exists before removing it.

rocks-lib/src/build/mod.rs
```diff
@@ -94,10 +95,14 @@ pub async fn build(progress: &MultiProgress, rockspec: Rockspec, config: &Config
...
-    let output_paths = tree.rock(&LuaPackage::new(
-        rockspec.package.clone(),
-        rockspec.version.clone(),
-    ))?;
+    // TODO(vhyrro): The `build` operation needs more metadata, specifically the dependency
+    // and pinning information.
+    let package = LuaPackage::from(
+        &RemotePackage::new(rockspec.package.clone(), rockspec.version.clone()),
+        LockConstraint::Unconstrained,
+    );
```
vhyrro, line 103: Since this is quite serious I'm in the process of making a follow-up pull request to make this work.

rocks-lib/src/lockfile/mod.rs
```diff
@@ -6,10 +6,12 @@ use itertools::Itertools;
...
 use sha2::{Digest, Sha256};
 
-use crate::lua_package::{LuaPackage, PackageName, PackageVersion, PackageVersionReq};
+use crate::lua_package::{
+    PackageName, PackageReq, PackageVersion, PackageVersionReq, RemotePackage,
+};
 
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, PartialOrd, Ord)]
-pub struct LockedPackage {
+pub struct LuaPackage {
```
vhyrro, line 14: This is the first major rename in the API: a `LuaPackage` now represents a physical package present on the filesystem, full of important metadata like dependencies and version constraints.

rocks-lib/src/lua_package/mod.rs
```diff
@@ -8,12 +8,12 @@ pub use version::{PackageVersion, PackageVersionReq};
 mod outdated;
 mod version;
 
-pub struct LuaPackage {
+pub struct RemotePackage {
```
vhyrro, line 11: This is the second major rename: a struct with only name and version information is merely a representation of a remote package somewhere on `luarocks.org`. This explains the rename, I think it conveys intent much more clearly.

rocks-lib/src/lua_package/mod.rs
```diff
@@ -40,15 +40,15 @@ impl LuaPackage {
 /// A lua package requirement with a name and an optional version requirement.
 #[derive(Debug, Clone, PartialEq)]
 #[cfg_attr(feature = "clap", derive(clap::Args))]
-pub struct LuaPackageReq {
+pub struct PackageReq {
```
vhyrro, line 43: Might seem like an odd rename, but I found `LuaPackageReq` to miss the point a bit. It's not a requirement on a `LuaPackage`, since those are always physical rocks in a rocktree.

In our code, we use this struct as general requirement matching on any sort of package, so I just called it `PackageReq`.

rocks-lib/src/lua_package/mod.rs
```diff
@@ -130,16 +130,11 @@ impl<'de> Deserialize<'de> for LuaPackageReq {
 
 /// A luarocks package name, which is always lowercase
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Hash)]
-pub struct PackageName {
-    name: String,
-}
+pub struct PackageName(String);
```
vhyrro, line 133: Wrapping these up as [newtypes](https://doc.rust-lang.org/rust-by-example/generics/new_types.html), since this is what these are at the end of the day.

vhyrro (commented):
I believe explanations in the form of a set of comments is easier to digest :)

rocks-lib/src/build/mod.rs
vhyrro: #121

rocks-lib/src/lockfile/mod.rs
mrcjkb: Let's set add this info as a docstring.
I'm wondering, would `LocalPackage` be a more suitable name?

mrcjkb (approved):

vhyrro:
I like the duality of `LocalPackage` and `RemotePackage`, sure.

Issue #123, opened 2024-10-07
[Bug] unable to unpack some `src.rock` archives


mrcjkb:
Unpacking `say` fails with

```
called `Result::unwrap()` on an `Err` value: invalid Zip archive: No valid central directory found
```

It works with luarocks.

mrcjkb:
This is because `say` has a `-3` specrev. Earlier versions likely have broken `src.rock` archives.
It looks like we might need to add basic specrev support.

PR #124, opened 2024-10-07
feat!(unpack): don't download src.rock to disk

mrcjkb:
While investigating #123, I noticed that we save `src.rock` files to disk before unpacking them, which is unnecessary.
This PR introduces a change where we unpack the `src.rock` from memory.
It's breaking because it removes the `--keep-rockspec` option to `rocks debug unpack` (we can re-add that later if needed).


mrcjkb:
Unfortunately, this doesn't fix #123.

I don't know why github won't let me remove the "merging this may close this issue" link.

vhyrro:
I have a fundamental question for this change: `.src.rock` files contain, well, the source, which could be potentially quite large. Are we really comfortable with storing all of that in memory?

mrcjkb:
> Are we really comfortable with storing all of that in memory?

The way it's implemented now, we're downloading it to memory and then writing it to disk. 
We'd have to use a stream if we wanted to reduce the memory footprint.
In practise, I think `.src.rock` archives (as well as packed rocks) are pretty small, so I'd say YAGNI for now.

One thing I'm confused about (for another issue): Why does `rocks install` unpack the `.src.rock` and then fetch the source from the rockspec's `source.url` if the `.src.rock` contains the source? Shouldn't we fall back to that if there's no `.src.rock`?

vhyrro:
Honestly I did this as a result of a personal grudge almost. Why on earth is the source processed with the rock in luarocks? What's the point of defining a source at that point anyway?

I really wanted to only download the rockspec without the source, but guess what, not happening. You need to know the name of the author of the rock to get the rockspec. sigh.

mrcjkb:
> In the implementation I really wanted to only download the rockspec without the source, but guess what, not happening. You need to know the name of the author of the rock to get the rockspec. sigh.

It's not documented, but instead of 

- https://luarocks.org/manifests/<author>/<package>-<version>.rockspec
  e.g. https://luarocks.org/manifests/neorocks/rocks.nvim-2.40.2-1.rockspec

you can also do

- https://luarocks.org/<package>-<version>.rockspec
  e.g. https://luarocks.org/rocks.nvim-2.40.2-1.rockspec

vhyrro:
Oh wow, I honestly believe we should be downloading the rockspecs only, purely for speed reasons.

The luarocks.org server is horridly slow for certain operations, so we should try to perform as much data grabbing on our end as possible.

This *will* mean that some very old rocks may cease to work, since their original sources were deleted. But I honestly don't mind that at all, and find the rockspec-only approach to be clean and elegant. Frankly, it'll also help us ensure that a source is reproducible and will make debugging less of a pain :p

mrcjkb:
> The luarocks.org server is horridly slow for certain operations, so we should try to perform as much data grabbing on our end as possible.

From what I've observed, searching (or parsing?) the manifest seems to be what currently takes the longest in `rocks install` (why on earth is it a Lua table? 😭).
But I agree, we should only fall back to the `.src.rock` if all else fails (opened #126).

vhyrro:
I'd have to take a look at the luarocks.org API a little closer. Surely someone has thought of having an endpoint to check for the existence of a rock or to return some rocks given some search criteria?

That way we wouldn't have to pull down the whole new manifest each time :smiling_face_with_tear:

PR #131, opened 2024-10-08
feat(ui): add some progress bars

mrcjkb:
We might end up removing them again when we parallelise our builds, but for now they might improve the UI for larger builds.

Stacked on #129

### DIFF
--- a/rocks-lib/src/build/builtin.rs
+++ b/rocks-lib/src/build/builtin.rs
@@ -10,12 +10,14 @@ use crate::{
     tree::RockLayout,
 };
 use eyre::{OptionExt as _, Result};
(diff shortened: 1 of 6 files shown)


rocks-lib/src/build/builtin.rs
```diff
@@ -39,6 +47,7 @@ impl Build for BuiltinBuildSpec {
                     )?
                 }
                 ModuleSpec::SourcePaths(files) => {
+                    bar.set_message("Comppiling C files...");
```
NTBBloodbath, line 50: Small typo here and also in Line #61 (I didn't want to duplicate the same review comment :p)

vhyrro (approved):

PR #135, opened 2024-10-10
feat!: allow `package@version` constraints

vhyrro:
What it says on the tin. Instead of `rocks install package >=1.0.0`, one must now invoke `rocks install package>=1.0.0`. Also added support for `rocks install package@1.0.0`.


mrcjkb (approved):

mrcjkb:
Definitely much simpler if we want to support bulk installs 👍

PR #136, opened 2024-10-11
feat: add `rocks info` command

vhyrro:
This feature is so uninspiring that I stopped working on it halfway, thus it's a draft :smile: 

I'd like to first implement `rocks doc` and have a general interface for grabbing documentation, so that it can be showed to the user when invoking `rocks info`.

I renamed `rocks show` to `rocks info` because `show` has always felt incredibly unintuitive to me. We may even want to make `rocks info` work for any rock, both remote and local, and display whether it's installed or not in the output.

### DIFF
--- a/rocks-bin/src/info.rs
+++ b/rocks-bin/src/info.rs
@@ -2,7 +2,10 @@ use clap::Args;
 use eyre::Result;
 use indicatif::MultiProgress;
 use rocks_lib::{
-    config::{Config, LuaVersion}, operations::download_rockspec, package::PackageReq, tree::Tree
+    config::{Config, LuaVersion},
+    operations::download_rockspec,
+    package::PackageReq,
+    tree::Tree,
 };
 
 #[derive(Args)]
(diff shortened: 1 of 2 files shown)


vhyrro:
I realized that I don't actually need `rocks doc` as a prerequisite for this so I made this as-is :)

rocks-bin/src/info.rs
```diff
@@ -0,0 +1,33 @@
...
+        println!("Currently installed in {}", tree.root().display());
+    }
+
+    println!("Package name: {}", rockspec.package);
+    println!("Package version: {}", rockspec.version);
+    println!();
+
+    println!("Summary: {}", rockspec.description.summary.unwrap_or("None".into()));
+    println!("Description: {}", rockspec.description.detailed.unwrap_or("None".into()));
+    println!("License: {}", rockspec.description.license.unwrap_or("None (all rights reserved by the author)".into()));
```
mrcjkb, line 29:
```suggestion
    println!("License: {}", rockspec.description.license.unwrap_or("Unknown (all rights reserved by the author)".into()));
```

mrcjkb (changes requested):
rustfmt is failing

PR #138, opened 2024-10-14
feat: make `rocks fmt` also format `project.rockspec`

vhyrro:


rocks-bin/src/format.rs
```diff
@@ -39,5 +39,18 @@ pub fn format() -> Result<()> {
             Ok::<_, eyre::Report>(())
         })?;
 
+    // Format the rockspec
+
+    let rockspec = project.root().join("project.rockspec");
```
mrcjkb, line 44: question: Does this only format a `project.rockspec` in the root directory, but no others?
Shouldn't we do this for any rockspec in the project root?

vhyrro: Having several rockspecs lying around is a `luarocks` thing, not a `rocks` thing. Rocks will derive new, in-memory rockspecs from `project.rockspec` for any operations like uploading. Therefore, in my opinion, we shouldn't touch the other rockspecs (which aren't ours and aren't tracked by `rocks`).

mrcjkb: sounds good to me

mrcjkb (approved):

PR #141, opened 2024-10-15
fix(manifest): don't return fallback result in `latest_match()`

vhyrro:
Fixed a logic bug and added regression case.


vhyrro:
Merging since this is trivial

PR #142, opened 2024-10-15
feat(build): install regular dependencies as opposed to just build dependencies

vhyrro:
This takes us one step closer to being fully functional. I've since tested a few rocks and some are failing with deserialization errors, something to look into next :)


rocks-bin/src/build.rs
```diff
@@ -64,13 +65,23 @@ pub async fn build(data: Build, config: Config) -> Result<()> {
     let tree = Tree::new(config.tree().clone(), lua_version)?;
 
     // Ensure all dependencies are installed first
-    // TODO: Handle regular dependencies as well.
-    let build_dependencies = rockspec.build_dependencies.current_platform();
+    let build_dependencies = rockspec
```
mrcjkb, line 68: suggestion: let's not worry about installing build dependencies at all until we implement a feature that needs them. 

I could imagine it being better to install them into a separate tree (lockfile? 🤔).

vhyrro: I think a temporary directory would be quite interesting for that. I'll remove the build dependencies then :)

mrcjkb (approved):

PR #148, opened 2024-10-18
feat: rust build backend

mrcjkb:
Technically, it would be possible for us to depend on cargo and use it as a library.
I'm not sure if we want to do that though. @vhyrro what do you think?
I don't think there would be a benefit, as people would still need to install rustc.

This is modeled according to the luarocks-build-rust-mlua rockspec.
There's only one difference. The luarocks build backend allows `features` to either be a `string` or a `string[]` list.
I couldn't be bothered to implement that. It doesn't make sense name-wise and it's not documented, so YAGNI.


mrcjkb:
Ready for review. Tested on `toml-edit`.

mrcjkb:
Still need to add an integration test for it.

max397574 (commented):
just looked at the code
lgtm
didn't have time to test out though

vhyrro:
We shouldn't rely on cargo as a library imo - anyone who has rustc will have cargo, so it amounts to the same thing in the end :)

I'll review the code now

rocks-lib/src/rockspec/build/utils/mod.rs
```diff
@@ -71,6 +71,26 @@ pub fn compile_c_files(
...
+/// the extension for external libraries.
+pub fn external_lib_extension() -> &'static str {
+    if cfg!(target_os = "windows") {
+        "dll"
+    } else if cfg!(target_os = "macos") {
+        "dylib"
+    } else {
+        "so"
+    }
+}
```
vhyrro, line 92: nit: This can be entirely replaced with `std::env::consts::DLL_EXTENSION`

vhyrro (approved):

PR #150, opened 2024-10-21
fix: properly unpack tar archives

vhyrro:
This PR fixes the installations of `dkjson`, which doesn't specify a `dir` name, even though it theoretically should.

This is fixed by implementing luarocks's "smart" search in tar files, which just involves checking if the unpacked contents of the tar is a single directory and recursively unpacking the directory instead.

### DIFF
--- a/rocks-lib/src/operations/fetch.rs
+++ b/rocks-lib/src/operations/fetch.rs
@@ -4,7 +4,9 @@ use flate2::read::GzDecoder;
 use git2::build::RepoBuilder;
 use git2::FetchOptions;
 use indicatif::MultiProgress;
+use itertools::Itertools;
 use std::fs::File;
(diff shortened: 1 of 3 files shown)


rocks-lib/src/operations/fetch.rs
```diff
@@ -134,19 +161,64 @@ async fn unpack<R: Read + Seek + Send>(
...
+
+                    entries.into_iter().all(|entry| {
+                        entry
+                            .path()
+                            .unwrap()
+                            .to_str()
+                            .unwrap()
+                            .starts_with(directory.to_str().unwrap())
+                    })
+                };
```
mrcjkb, line 198: suggestion(readability): let's extract this into a named function.

mrcjkb (approved):

PR #151, opened 2024-10-21
fix: replace `git://` URIs with `https://`

vhyrro:
stacked on #150 

Because the `git://` protocol is oftentimes blocked by traditional firewalls, we instead redirect the git URL to `https://` for better compatibility.


vhyrro:
tests failing, will fix

mrcjkb (approved):

PR #154, opened 2024-10-22
feat: `rocks pin`

vhyrro:


vhyrro:
Would be wise to implement a `rocks unpin` too, I guess?

rocks-bin/src/pin.rs
```diff
@@ -0,0 +1,24 @@
...
+
+#[derive(Args)]
+pub struct Pin {
+    package: PackageReq,
+}
+
+pub fn pin(data: Pin, config: Config) -> Result<()> {
+    let tree = Tree::new(config.tree().clone(), LuaVersion::from(&config)?)?;
+
+    if let Some(mut rock) = tree.has_rock(&data.package) {
```
mrcjkb, line 19: question: wouldn't you want to pin an exact `Package` instead of a `PackageReq`?

Otherwise, what if there are multiple rocks installed that match the `PackageReq`?

PR #157, opened 2024-10-23
feat: `rocks unpin` command

vhyrro:


rocks-bin/src/pin.rs
```diff
@@ -9,17 +9,17 @@ use rocks_lib::{
 };
 
 #[derive(Args)]
-pub struct Pin {
+pub struct ChangePin {
     package: RemotePackage,
 }
 
-pub fn pin(data: Pin, config: Config) -> Result<()> {
+pub fn pin(data: ChangePin, config: Config, pin: bool) -> Result<()> {
```
mrcjkb, line 16: issue: This feels confusing to read: a `pin` function that takes a `pin: bool` argument.

Can we split it into separate functions for pinning and unpinning, or pass in an enum to avoid boolean blindness?

rocks-lib/src/operations/pin.rs
```diff
@@ -6,8 +6,8 @@ use crate::{lockfile::LocalPackage, tree::Tree};
 
 // TODO(vhyrro): Differentiate pinned LocalPackages at the type level?
 
-pub fn pin(package: &mut LocalPackage, tree: &Tree) -> Result<()> {
-    if package.pinned() {
+pub fn pin(package: &mut LocalPackage, tree: &Tree, pin: bool) -> Result<()> {
+    if pin == package.pinned() {
         bail!("Rock {} is already pinned!", package.to_package());
```
mrcjkb, line 11: issue: The error message no longer makes sense when unpinning.

mrcjkb (changes requested):

rocks-bin/src/pin.rs
vhyrro: Funnily enough that was my first instinct, but it duplicated code unnecessarily and it felt worse than adding a boolean switch here. I was struggling with an alternative name: `repin`, `set_pin_state`? They're all pretty messy. I'll wrap the bool in an enum and also make other parts of the codebase use this enum for pinning.

mrcjkb: I think `set_pinned_state` is a good name (and I still prefer enums over booleans for readability).

vhyrro: Let me know if there's anything you'd change with what I just pushed :)

mrcjkb (approved):

rocks-bin/src/pin.rs
mrcjkb: lgtm

PR #160, opened 2024-10-23
fix: don't allow installation of conflicting rocks

vhyrro:


rocks-lib/src/build/mod.rs
```diff
@@ -143,6 +143,14 @@ pub async fn build(
     );
     package.pinned = pinned;
 
+    if tree.lockfile()?.get(&package.id()).is_some() {
+        bail!(
+            "Unable to build {}, identical rock already found in tree {}",
+            package.into_package_req(),
+            tree.root().display()
+        );
+    }
```
mrcjkb, line 152: question: Shouldn't we perform this check before the build/install and skip the step with a `progress.println("rock is already installed. Skipping.")` message (and then exit with `0`)?

We should definitely avoid encountering this error when installing dependencies.

In case of the `rocks build` command, we probably want to prompt whether to overwrite the existing rock.

vhyrro: We can add a check before the install/build steps on top, but we should definitely have a check internally here too - I'm thinking ahead to consumers of our API who will unknowingly and happily overwrite rocks in their rocktree upon invoking `install()`.

Currently, this code is somewhat of a problem, since `eyre` doesn't differentiate error types. Once we migrate to `thiserror` or any stricter error library we can ignore this error whenever we know it's unimportant. Furthermore, we could supply `build()` with a `force` flag which will prevent this error from being issued.

Either way, we should have some sort of flag builtin here which requires a conscious decision from the API consumer, what do you think?

mrcjkb: I think a `force` flag would be more useful than stricter error handling here.

vhyrro: Done. I currently supply these flags statically on a best-guess basis (e.g. `rocks install` has `force = true`). I'll see if I have some time later today to implement the prompts :)

PR #161, opened 2024-10-24
feat(lib)!: move to `thiserror`

vhyrro:
~~Still very much a draft (probably not even halfway).~~

This PR moves the entire library codebase to `thiserror` for proper error handling. Some error types are purposefully verbose for better error messages later down the line.

Feel free to review this PR while it's still in its infancy, so we can wrap up issues and inconsistencies ASAP. I'll be honest, this sort of refactoring really messes with your head after a while so a chunk of the refactors were  done on autopilot, so feel free to be harsh with the reviews, lol.

Also, do take your time with the reviews, because looking at the github changelist makes my head spin :smile:


rocks-lib/src/manifest/metadata.rs
```diff
@@ -25,7 +24,7 @@ impl<'de> serde::Deserialize<'de> for ManifestMetadata {
 impl ManifestMetadata {
     // TODO(vhyrro): Perhaps make these functions return a cached, in-memory version of the
     // manifest if it has already been parsed?
-    pub fn new(manifest: &String) -> Result<Self> {
+    pub fn new(manifest: &String) -> mlua::Result<Self> {
```
vhyrro, line 27: note: we should not be returning internal mlua types, this should be wrapped in a newtype

rocks-lib/src/build/builtin.rs
```diff
@@ -99,7 +99,7 @@ fn autodetect_modules(build_dir: &Path) -> Result<HashMap<String, ModuleSpec>> {
         })
         .map(|file| {
             let diff: PathBuf = pathdiff::diff_paths(build_dir.join(file.into_path()), build_dir)
-                .ok_or_eyre("unable to autodetect modules")?;
+                .expect("failed to autodetect modules");
```
mrcjkb, line 102: question: This panics. Do we want this?

I'd prefer to avoid panicking, in case we need to perform cleanup (e.g. remove a partial install) somewhere higher in the stack.

mrcjkb (commented):
Overall, this looks like a good change to me.

rocks-lib/src/build/builtin.rs
vhyrro: In this specific case I'm not afraid - the panic happens if `pathdiff` ever fails to diff paths, which is a very rare occurence in of itself. It's so niche (and the only error in that function), so propagating it is a waste of time in my eyes.

vhyrro:
Fully ready to review now - did all the cleanup that I wanted to do :)

mrcjkb (approved):

PR #178, opened 2024-10-31
fix: incorrect exporting of variables in fish, don't error when `--append` is passed on empty environment variables

vhyrro:


mrcjkb (approved):

mrcjkb:
rustfmt is failing

PR #185, opened 2024-11-06
feat: `rocks upload`

vhyrro:
TODOs:
- [x] DRY for `format!()` calls
- [x] Proper signature creation. In my opinion, signatures should be enabled and enforced by default.

### DIFF
--- a/rocks-bin/src/upload.rs
+++ b/rocks-bin/src/upload.rs
@@ -1,0 +1,21 @@
+use clap::Args;
+use eyre::Result;
+use rocks_lib::{
+    config::Config,
+    project::Project,
+    upload::{upload_from_project, ApiKey, SignatureProtocol},
+};
+
+#[derive(Args)]
+pub struct Upload {
+    #[arg(long, default_value_t)]
+    sign_protocol: SignatureProtocol,
+}
+
+pub async fn upload(data: Upload, config: Config) -> Result<()> {
+    let project = Project::current()?.unwrap();
+
+    upload_from_project(&project, &ApiKey::new()?, data.sign_protocol, &config).await?;
+
+    Ok(())
+}
--- a/rocks-lib/Cargo.toml
+++ b/rocks-lib/Cargo.toml
@@ -17,7 +17,7 @@ httpdate = "1.0.3"
 itertools = "0.13.0"
 mlua = { version = "0.10.0", features = ["luajit52", "serialize", "send"] }
 pathdiff = "0.2.1"
-reqwest = "0.12.0"
+reqwest = { version = "0.12.0", features = ["json", "multipart"] }
 semver = "1.0.22"
 serde = { version = "1.0.197", features = ["derive"] }
 serde_json = { version = "1.0.114", features = ["preserve_order"] }
@@ -47,6 +47,7 @@ sha2 = "0.10.8"
 hex = { version = "0.4.3" }
 fs_extra = "1.3.0"
 thiserror = "2.0.0"
+gpgme = "0.11.0"
 
 [dev-dependencies]
 httptest = { version = "0.16.1" }
--- a/rocks-lib/src/upload/mod.rs
+++ b/rocks-lib/src/upload/mod.rs
@@ -1,0 +1,267 @@
+use std::env;
+use std::io::Read;
+
+use crate::TOOL_VERSION;
+use crate::{config::Config, project::Project};
+use clap::ValueEnum;
+use gpgme::{Context, Data};
+use reqwest::{
+    multipart::{Form, Part},
+    Client,
+};
+use serde::Deserialize;
+use serde_enum_str::Serialize_enum_str;
+use thiserror::Error;
+
+#[derive(Deserialize, Debug)]
+pub struct VersionCheckResponse {
+    version: String,
+}
+
+#[derive(Error, Debug)]
+pub enum ToolCheckError {
+    #[error(transparent)]
+    Request(#[from] reqwest::Error),
+    #[error("`rocks` is out of date with {0}'s expected tool version! `rocks` is at version {TOOL_VERSION}, server is at {server_version}", server_version = _1.version)]
+    ToolOutdated(String, VersionCheckResponse),
+}
+
+#[derive(Error, Debug)]
+pub enum UserCheckError {
+    #[error(transparent)]
+    Request(#[from] reqwest::Error),
+    #[error("invalid API key provided")]
+    UserNotFound,
+}
+
+#[derive(Error, Debug)]
+#[error("could not check rock status on server: {0}")]
+pub struct RockCheckError(#[from] reqwest::Error);
+
+#[derive(Error, Debug)]
+#[error(transparent)]
+pub enum UploadError {
+    Lua(#[from] mlua::Error),
+    Request(#[from] reqwest::Error),
+    RockCheck(#[from] RockCheckError),
+    #[error("rock already exists on server: {0}")]
+    RockExists(String),
+    #[error("unable to read rockspec: {0}")]
+    RockspecRead(#[from] std::io::Error),
+    #[error("{0}.\nHINT: If you'd like to skip the signing step supply `--sign-protocol none` to the CLI")]
+    Signature(#[from] gpgme::Error),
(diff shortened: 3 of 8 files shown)


rocks-bin/src/upload.rs
```diff
@@ -0,0 +1,19 @@
...
+use eyre::Result;
+use rocks_lib::{config::Config, project::Project, upload::upload_from_project};
+
+pub async fn upload(config: Config) -> Result<()> {
+    let project = Project::current()?.unwrap();
+
+    upload_from_project(
+        &project,
+        env::var("ROCKS_API_KEY")
+            .map_err(|_| eyre!("No API key provided! Please set the $ROCKS_API_KEY variable."))?,
```
mrcjkb, line 13: suggestion(for-another-issue): let's add a config option that allows us to use `libsecret` to fetch API keys,

mrcjkb, line 13: suggestion: We could also prompt for the API key if it's not set.

rocks-lib/src/upload/mod.rs
```diff
@@ -0,0 +1,177 @@
...
+    ToolCheck(#[from] ToolCheckError),
+    UserCheck(#[from] UserCheckError),
+    RockCheck(#[from] RockCheckError),
+    #[error("rock already exists on server: {0}")]
+    RockExists(String),
+}
+
+pub async fn upload_from_project(
+    project: &Project,
+    api_key: String,
```
mrcjkb, line 53: issue: This should be a newtype with a masked `Display` implementation.

rocks-lib/src/upload/mod.rs
```diff
@@ -0,0 +1,177 @@
...
+        .file_name(format!(
+            "{}-{}.rockspec",
+            rockspec.package, rockspec.version
+        ))
+        .mime_str("application/octet-stream")?;
+    //let signature = reqwest::multipart::Part::text("something").file_name("project.rockspec.sig");
+    let multipart = Form::new().part("rockspec_file", rockspec);
+
+    client
+        .post(format!("{}{}/{}/upload", config.server(), "api/1", api_key))
```
mrcjkb, line 106: question(security): Can we somehow tell the client that the url contains sensitive information?

mrcjkb (changes requested):

rocks-lib/src/upload/mod.rs
vhyrro: Unsure, I'll have to look into that. That's not on us though, I don't know who thought putting the API key in the URL is a good idea :sweat_smile:

rocks-bin/src/upload.rs
vhyrro: I'm personally against this - I'd rather the user drop their API key in a `.env` file or something as opposed to prompting it :)

mrcjkb: fine by me. I don't feel strongly about it

rocks-lib/src/upload/mod.rs
mrcjkb: If the client doesn't log anything or include the url in error messages, we should be fine.

> I don't know who thought putting the API key in the URL is a good idea

You raise a good point. Firewalls and proxies will likely log the URL unhashed. This is potentially a major security issue that we should raise with luarocks-site.

vhyrro: Not sure if it's *major*, but it's definitely larger-than-minor. Any sort of logging immediately leaks the API key :/

mrcjkb: which is a risk for supply chain attacks.

rocks-lib/src/upload/mod.rs
vhyrro: done :)

vhyrro:
question: should I gate `gpgme` behind a feature flag that's enabled by default, or should we bundle GPG in our CI?

mrcjkb:
> question: should I gate `gpgme` behind a feature flag that's enabled by default, or should we bundle GPG in our CI?

We can add `gpg` to the `buildInputs` (or `propagatedBuildInputs`? cc @teto) in the `overlay.nix`.

rocks-lib/src/upload/mod.rs
teto: it is major omg :o

teto:
sure. Whatever works :)

rocks-lib/src/upload/mod.rs
mrcjkb: luarocks uses https, so the path will be encrypted. But there is still a risk on the server side.

mrcjkb (approved):

rocks-lib/Cargo.toml
```diff
@@ -15,9 +15,9 @@ git2 = "0.19.0"
 html-escape = "0.2.13"
 httpdate = "1.0.3"
 itertools = "0.13.0"
-mlua = { version = "0.10.0", features = ["luajit52", "serialize", "send"] }
+mlua = { git = "https://github.com/vhyrro/mlua", features = ["luajit52", "serialize", "send"] }
```
mrcjkb, line 18: question: is this change needed for this feature? Or is it left over from the `rocks add` feature?
If it's needed, we'll have to add

```nix
outputHashes = {
  "mlua-0.10.0" = "<hash>";
};
```
to the cargo lockfile import in the nix build expression.

vhyrro: It's needed in this PR - I'm wondering whether I could refactor the code to stop this from being a requirement though. I'll look into it in 2 hours.

vhyrro:
Everything's done - all that's left is making the CI happy now.

rocks-lib/src/upload/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+            let mut signature = Data::new()?;
+
+            match ctx.sign_detached(rockspec_content.clone(), &mut signature) {
+                Ok(_) => {
+                    let mut signature_str = String::new();
+                    signature.read_to_string(&mut signature_str)?;
+
+                    Some(signature_str)
+                }
+                // TODO: Issue a warning here that signing wasn't successful.
```
mrcjkb, line 132: suggestion: Le'ts not issue a warning. Instead, let's default to returning an `Err`, suggesting to pass in a `--skip-sign` flag.
I think it's better if people have to explicitly skip signing.

rocks-lib/src/upload/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+
+pub async fn upload_from_project(
+    project: &Project,
+    api_key: &ApiKey,
+    config: &Config,
+) -> Result<(), UploadError> {
+    let client = Client::new();
+
+    helpers::ensure_tool_version(&client, config.server()).await?;
+    helpers::ensure_user_exists(&client, api_key, config.server()).await?;
```
mrcjkb, line 101: suggestion: since the API key could be leaked if the URL is not HTTPS, how about we check that it is and return an error if it isn't?

```suggestion

    helpers::ensure_https(config.server())?;
    helpers::ensure_tool_version(&client, config.server()).await?;
    helpers::ensure_user_exists(&client, api_key, config.server()).await?;
```

We could also have `config.server()` perform the check :thinking:

mrcjkb (approved):

mrcjkb:
> Everything's done - all that's left is making the CI happy now.

I've fixed CI. You'll need to pull the changes in your checkout @vhyrro

rocks-lib/src/upload/mod.rs
vhyrro: Done. For consistency's sake I made it a `--sign-protocol` flag, allowing the user to select their protocol of choice. Whenever GPG issues an error, the user also receives a hint saying they can pass `--sign-protocol none` to skip the signing step :)

rocks-lib/src/upload/mod.rs
vhyrro: Sounds like a plan. Instead of checking the URL string, I'll see if there's a way to force `reqwest` to make HTTPS calls only.

mrcjkb (approved):

Issue #182, opened 2024-11-03
[bug] `rocks install luafilesystem` installs C source to `src/lfs.lua`.


mrcjkb:
Excerpt from the rockspec:

```lua
build = {
   type = "builtin",
   modules = {
      lfs = "src/lfs.c"
   },
   copy_directories = {
      "doc", "tests"
   }
}
```

PR #187, opened 2024-11-07
fix: install `src/<lib>.c` modules correctly

mrcjkb:
Addresses #182.

However, this doesn't fully solve the issue, because:

- [x] `cc-rs` appears to be building static libraries, prepending `lib` and appending `.a` even if `shared_flag` is set to `true`.
  - https://github.com/rust-lang/cc-rs/issues/594
- #188.


mrcjkb:
This one is ready. I will address #188 in a separate PR.

PR #189, opened 2024-11-07
fix(lockfile): add transitive dependencies

mrcjkb:
This initialises and flushes the lockfile outside of the recursive `install_impl` call,
ensuring the lockfile doesn't get recreated before it's flushed.


vhyrro:
LGTM, I don't think the lifetime variables do anything though, might be best to just remove them. You'd probably need a PhantomData for the desired effect.

vhyrro (approved):

PR #194, opened 2024-11-13
tests(rockspec): more detailed luasystem assertions

mrcjkb:


mrcjkb:
Self-merging, as this is trivial.

PR #199, opened 2024-11-16
feat: implement UserData for all major structs

vhyrro:
What it says on the tin: this PR adds a `lua` feature flag and implements UserData for all major data types in the codebase. This will be complemented by a second PR that adds a `rocks-lua` module, allowing `rocks` to be packaged and used directly in any Lua application.

Notes: we could implement __tostring methods for all these types later down the road, making debugging them easier. What do we think?

### DIFF
--- a/rocks-lib/src/package/mod.rs
+++ b/rocks-lib/src/package/mod.rs
@@ -13,6 +13,7 @@ pub use version::{PackageVersion, PackageVersionReq};
 
 #[derive(Clone, Debug)]
 #[cfg_attr(feature = "clap", derive(clap::Args))]
+#[cfg_attr(feature = "lua", derive(mlua::FromLua))]
 pub struct RemotePackage {
     name: PackageName,
     version: PackageVersion,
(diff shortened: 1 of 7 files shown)


rocks-lib/src/config/mod.rs
```diff
@@ -9,6 +9,7 @@ use crate::{
 };
 
 #[derive(Debug, Clone, PartialEq, Eq)]
+//#[cfg_attr(feature = "lua", derive(serde::Serialize, serde::Deserialize))]
```
vhyrro, line 12: oops

vhyrro:
I'll go through clippy's complaints and fix them up :+1:

rocks-lib/src/package/mod.rs
```diff
@@ -124,6 +140,15 @@ impl From<RemotePackage> for PackageReq {
...
 }
 
+#[cfg(feature = "lua")]
+impl mlua::UserData for PackageReq {
+    fn add_methods<M: mlua::UserDataMethods<Self>>(methods: &mut M) {
+        methods.add_meta_method("__eq", |_, this, package: RemotePackage| {
+            Ok(this.matches(&package))
+        });
+    }
+}
```
mrcjkb, line 150: question: Does `__eq` make sense here?
I'd lean towards a `matches` method instead.

vhyrro: Good call - brain was on autopilot mode, I thought this was `RemotePackage:matches(&RemotePackage)`, but this is a PackageReq

mrcjkb (approved):

PR #202, opened 2024-11-16
feat(install): multiple packages in one command

mrcjkb:
This adds an interface for installing multiple packages at once (no parallel builds yet).


rocks-bin/src/install.rs
```diff
@@ -26,32 +30,30 @@ pub async fn install(data: Install, config: Config) -> Result<()> {
...
+            let build_behaviour: Result<BuildBehaviour> =
+                match tree.has_rock_and(&req, |rock| pin == rock.pinned()) {
+                    Some(_) if !data.force => {
+                        if Confirm::new(&format!("Package {} already exists. Overwrite?", req))
+                            .with_default(false)
+                            .prompt()?
+                        {
+                            Ok(BuildBehaviour::Force)
+                        } else {
+                            Ok(BuildBehaviour::NoForce)
```
mrcjkb, line 46: TODO:

- [ ] This doesn't prevent the package from being installed.

vhyrro (approved):
Apart from the issue you raised code looks good to me :)

PR #209, opened 2024-11-19
refactor: move away from `indicatif::MultiProgress`

vhyrro:
~~Still *loads* of work left to do with this PR, so don't review yet.~~

This PR refactors the CLI experience. We stop exposing a library struct and instead create our own newtype wrappers. These wrappers are then propagated throughout the codebase. In addition, instead of sending MultiProgresses everywhere, we now primarily send ProgressBars, since most operations operate on a single thing and only need a single progress bar to display progress.

### DIFF
--- a/rocks-bin/src/build.rs
+++ b/rocks-bin/src/build.rs
@@ -1,5 +1,4 @@
 use eyre::eyre;
-use indicatif::MultiProgress;
 use inquire::Confirm;
 use itertools::Itertools;
 use std::path::PathBuf;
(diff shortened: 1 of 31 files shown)


vhyrro:
PR is starting to take shape. I'd like to do one last refactor of the `MultiProgress` API since I'm not exactly happy with it yet, a little too flaky :)

vhyrro:
There are a few extraneous variables here and there still, I'll go and clean that up :)

mrcjkb (approved):
praise: Just tested this. What an improvement! :rocket:

PR #210, opened 2024-11-19
feat: `rocks run`

mrcjkb:
Could be useful for debugging commands installed with rocks (like `busted`, `luarocks`, `nlua`, ...).


vhyrro (approved):
Looks good to me. Will conflict with #209, but the changes don't look too drastic (just a few MultiProgress related stuff :p).

.github/workflows/integration-tests.yml
```diff
@@ -10,6 +10,11 @@ jobs:
       - uses: DeterminateSystems/nix-installer-action@main
       - uses: DeterminateSystems/magic-nix-cache-action@main
       - uses: Swatinem/rust-cache@v2
+      - name: Install neovim
+        uses: rhysd/action-setup-vim@v1 # Used by the 'run' integration test
+        with:
+          neovim: true
+          version: v0.10.2
```
max397574, line 17: I got a function laying around somewhere to automatically get the latest tag
let me find this rq

max397574 (approved):

.github/workflows/integration-tests.yml
max397574: up to you if you want to add it
https://github.com/max397574/care.nvim/blob/61f564c71ac3642f4297428e3bdb03af6b1ebd4a/.github/workflows/typecheck.yml#L14

mrcjkb: Actually, for this workflow if you omit `version`, it'll install the latest stable version.

PR #211, opened 2024-11-19
refactor(build): install dependencies in one pass

mrcjkb:


vhyrro (approved):
This also happens to fix an annoyance in #209, awesome!

PR #214, opened 2024-11-21
refactor(install): flatten packages before installing

mrcjkb:
Sorry for all the clones :sweat_smile: 

This will allow us to parallelise installs.

Process:

- Before installing, recursively gather information about `LocalPackage`s to install into a flattened vector.
  This introduces a `LocalPackageSpec` type, which does not contain the source hash info, as we don't know it yet.
  The important thing is that this type has enough info to construct the `LocalPackageId`.
- Filter out duplicates, by the `LocalPackageSpecId`s.
- Build the packages (because we don't have any duplicates, we can now parallelise this in a follow-up PR)
  and gather the installed `LocalPackage`s, in a `HashMap`, with the `LocalPackageId` as the key.
- Use the previously gathered `LocalPackageSpec`s and the `HashMap` to reconstruct the dependency tree and populate the lockfile.

I worked on this in one go and am pretty tired. It will need lots of cleanup (maybe in a follow-up PR, to avoid conflicts?).
Please feel free to roast it!


rocks-lib/src/lockfile/mod.rs
```diff
@@ -57,39 +62,46 @@ impl<'de> Deserialize<'de> for PinnedState {
     }
 }
 
-// TODO(vhyrro): Move to `package/local.rs`
-#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, PartialOrd, Ord)]
-#[cfg_attr(feature = "lua", derive(mlua::FromLua))]
-pub struct LocalPackage {
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub struct LocalPackageSpec {
```
mrcjkb, line 66: note: I think this should be `pub(crate)`, as it's only needed internally.

rocks-lib/src/operations/install.rs
```diff
@@ -43,77 +51,118 @@ where
...
+    let all_packages = join_all(packages.iter().map(|(behaviour, pkg)| {
+        flattened_install_specs(&bar, (*behaviour, pkg.clone()), &pin, config)
+    }))
+    .await
+    .into_iter()
+    .try_collect::<_, Vec<Vec<PackageInstallSpec>>, InstallError>()?
+    .into_iter()
+    .flatten()
+    .unique_by(|install_spec| install_spec.local_package_spec.id().clone())
+    .collect_vec();
```
vhyrro, line 73: misc: I'm soo curious if this can be shortened down even further. It's like the code is taunting me to try and reduce it :joy:, this is fine for the time being.

rocks-lib/src/lockfile/mod.rs
```diff
@@ -115,40 +132,162 @@ impl LocalPackage {
...
+#[cfg_attr(feature = "lua", derive(FromLua,))]
+#[derive(Debug, Serialize, Deserialize, Clone)]
+struct LocalPackageIntermediate {
+    name: PackageName,
+    version: PackageVersion,
+    pinned: PinnedState,
+    dependencies: Vec<LocalPackageId>,
+    constraint: Option<String>,
+    hashes: LocalPackageHashes,
+}
```
vhyrro, line 164: question: could we make do without this intermediary struct? I think it makes the deserialization flow quite hard to follow - is there anything blocking us from doing the custom deserialization magic directly for LocalPackage? I can't pick it up just from looking at the code :smile:

vhyrro (commented):
I like the changes, very happy that package IDs are getting their own newtype now (and that we can get rid of async recursion, lol). I'm just curious if we could get away without the `LocalPackageIntermediate` struct.

rocks-lib/src/lockfile/mod.rs
vhyrro: Ah I see, this is about `FromLua`. Maybe we should just implement FromLua ourselves here to solve this issue? It's a shame the derive macro doesn't allow cheeky things like `flatten`/`skip_serializing` like you can in serde hah.

rocks-lib/src/operations/install.rs
mrcjkb: The reason this has so much boilerplate is that `join_all().await` returns a `Vec`, not an `Iterator` :disappointed:

rocks-lib/src/lockfile/mod.rs
mrcjkb: > Maybe we should just implement FromLua ourselves here to solve this issue?

That's what I'm doing (also for `Serialize` and `Deserialize`).
The nice thing about this approach is that those 3 trait implementations can reuse the `TryFrom` and `From` implementations.

```rust
impl<'de> Deserialize<'de> for LocalPackage {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: serde::Deserializer<'de>,
    {
        LocalPackage::try_from(LocalPackageIntermediate::deserialize(deserializer)?)
            .map_err(de::Error::custom)
    }
}

impl Serialize for LocalPackage {
    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
    where
        S: serde::Serializer,
    {
        LocalPackageIntermediate::from(self).serialize(serializer)
    }
}

#[cfg(feature = "lua")]
impl FromLua for LocalPackage {
    fn from_lua(value: mlua::Value, lua: &mlua::Lua) -> mlua::Result<Self> {
        LocalPackage::try_from(LocalPackageIntermediate::from_lua(value, lua)?)
            .map_err(|err| mlua::Error::DeserializeError(format!("{}", err)))
    }
}
```

vhyrro: let's make LocalPackageSpec `pub(crate)`, as you intended :)

mrcjkb: Done

PR #215, opened 2024-11-22
fix(list): show package name only once

vhyrro:
Fixes a slight annoyance where the package name would be displayed multiple times in `rocks list` output.


max397574 (approved):
why did that even work 🤔 
what happens in rust if you provide too many placeholders in format string?

vhyrro:
> why did that even work 🤔 what happens in rust if you provide too many placeholders in format string?

it errors. the old code worked, it just happened to print the package name, which is already printed in the root of the tree

mrcjkb (approved):

PR #217, opened 2024-11-24
feat: parallel installs

vhyrro:
Imma be honest after 1 hour of async recursion shenanigans I gave up and so we currently just have:
- Parallel installs
- Parallel rockspec pulling (but only for immediate dependencies, not transitive ones)

I think a full refactor of `operations/install.rs` is required to make this work as we'd like. Nevertheless, the speed gains are massive! It all works very nicely.


vhyrro:
I'll deal with the cleanup tomorrow :)

vhyrro:
Improvements since last push:
- Parallel installation of rockspecs
- Simplified code
- Possibly made the Intermediate structs redundant?

Potential issues:
- I'm fairly certain dependency loops will cause our code to run forever
- We don't do any deduplication based on ID. I'll have to reimplement that next.

vhyrro:
Integration tests seem to be failing for weird reasons (gpg-me)? If everything else succeeds then this PR is ready for proper review or merge. We can then layer other optimizations on top like parallelizing other parts of the codebase or making the code use a single `Manifest` instead of pulling it down every time.

mrcjkb (approved):

PR #223, opened 2024-11-25
feat: add shared `ManifestMetadata` state

vhyrro:
This PR shares a single `ManifestMetadata` between all invocations of `install`, preventing constant pulling of the manifest. It does make the function invocations slightly more clunky, but I leave that problem to another giga-refactor.

The speed gains are not as noticeable since multithreading was introduced, but it still amounts to a substantial amount of the course of 10 or so packages. On systems starved on threads, this will introduce massive performance gains.


max397574 (approved):

mrcjkb (approved):

mrcjkb:
Compilation of the integration tests is failing.

Issue #213, opened 2024-11-20
[QOL] If `--lua-version` is not set, get it from the installed lua version


mrcjkb:
We can parse the Lua version using the command `lua -v`.

PR #226, opened 2024-11-26
ci: integration tests with multiple lua versions on multiple platforms

mrcjkb:
Stacked on #228 (so we can run the busted integration test with multiple lua versions).


mrcjkb:
This may be blocked by 
- #213

mrcjkb:
Looks like we have a linker error on `aarch64-darwin` when installing `busted`:

```
called `Result::unwrap()` on an `Err` value: InstallError(BuildError(CommandFailure { status: ExitStatus(unix_wait_status(256)), stdout: "", stderr: "Undefined symbols for architecture arm64:\n  \"_luaL_argerror\"
# ...
```

mrcjkb:
ah, what are these jj shenanigans?

rocks-lib/src/operations/install.rs
```diff
@@ -26,11 +26,16 @@ use super::SearchAndDownloadError;
 #[derive(Error, Debug)]
 #[error(transparent)]
 pub enum InstallError {
+    #[error(transparent)]
     SearchAndDownloadError(#[from] SearchAndDownloadError),
+    #[error(transparent)]
     LuaVersionError(#[from] LuaVersionError),
+    #[error(transparent)]
     LuaVersionUnset(#[from] LuaVersionUnset),
+    #[error(transparent)]
```
vhyrro, line 35: nitpick: because of the top-level `#[error(transparent)]` these extra statements here are redundant

vhyrro (approved):

rocks-lib/src/operations/install.rs
mrcjkb: Removed the top-level one, because it doesn't apply to all constructors anymore.

PR #227, opened 2024-11-26
feat(install): deduplicate rocks

vhyrro:


vhyrro:
For the record, I think that the LocalPackageIntermediate type is redundant after these changes (if these are desirable changes to begin with).

mrcjkb (approved):

PR #229, opened 2024-11-27
fix(install): properly assign dependencies in the lockfile

vhyrro:
This PR accumulates the IDs of rocks during the initial dependency search, creating a dependency graph as it chugs along. After installing the required packages, this dependency graph is referenced to properly build up the lockfile.

To test this, first `rocks purge` your tree, which will also remove the `lock.json`. Then, install your favourite package and `cat /my/tree/5.4/lock.json` :)


vhyrro:
Am I proud of all these extra clones? No. Do they severely impact performance? No. Will they be removed in the giga-refactor? Yes.

mrcjkb (approved):

PR #230, opened 2024-11-28
refactor!: make progress indicators optional

vhyrro:
This refactors the entire library to allow progress indicators to be optional (crucial for #201).

This introduces a sealed trait called `HasProgress` and an enum called `Progress`. These allow the usage of either `MultiProgress` or `ProgressBar` in a similar fashion to how you would use an `Option<T>`.

Note: this PR doesn't implement a `--silent` flag of any kind, and continues to assume that `Progress` should always be displayed through the CLI. This PR serves as the skeleton for that feature, which may be implemented in a separate PR.

Let me know what you think @mrcjkb.

### DIFF
--- a/rocks-lib/src/progress.rs
+++ b/rocks-lib/src/progress.rs
@@ -1,5 +1,39 @@
 use std::{borrow::Cow, time::Duration};
 
+mod private {
+    pub trait HasProgress {}
+}
+
+impl private::HasProgress for () {}
+impl private::HasProgress for MultiProgress {}
+impl private::HasProgress for ProgressBar {}
+
+pub trait HasProgress: private::HasProgress {}
+impl<P: private::HasProgress> HasProgress for P {}
+
+#[derive(Clone)]
+pub enum Progress<T: HasProgress> {
+    NoProgress,
+    Progress(T),
+}
+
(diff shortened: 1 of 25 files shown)


vhyrro:
Huh? I didn't press nor push anything :skull:

rocks-lib/src/progress.rs
```diff
@@ -1,5 +1,39 @@
...
+impl private::HasProgress for ProgressBar {}
+
+pub trait HasProgress: private::HasProgress {}
+impl<P: private::HasProgress> HasProgress for P {}
+
+#[derive(Clone)]
+pub enum Progress<T: HasProgress> {
+    NoProgress,
+    Progress(T),
+}
```
mrcjkb, line 18: question: Can this be a newtype wrapper around an `Option<T>`?

vhyrro: I've been thinking of implementing `Deref` to an Option<T>, but I'm not sure we can uphold proper type constraint rules that way. I prefer having this as an enum because we can simply supply a `NoProgress` and have it be much more semantic.

mrcjkb (approved):

PR #234, opened 2024-12-02
build(nix): some flake tweaks

mrcjkb:
Just some small improvements to our flake.nix:

- Inherit systems from nixpkgs.
- Don't evaluate nixpkgs to add the overlay, extend it instead.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -23,23 +23,15 @@
     flake-parts,
     git-hooks,
     ...
-  }: let
-    overlay = import ./nix/overlay.nix {inherit self;};
-  in
+  }:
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -23,23 +23,15 @@
...
-          overlays = [
-            overlay
-          ];
-        };
+      systems = builtins.attrNames nixpkgs.legacyPackages;
+      perSystem = {
+        system,
+        pkgs,
+        ...
+      }: let
```
teto, line 33: alternatively if you dont like the `pkgs' = pkgs` (which is fine but increases the diff), 
`perSystem = { system, ... }@attrs: let pkgs= attrs.pkgs;`

PR #238, opened 2024-12-04
fix(build/builtin): properly auto-detect modules

### LINKED ISSUE #236: [Bug] auto detected lua modules are not installed if not specified in the install spec
The `BuiltinBuildSpec` auto-detects Lua modules, but they don't appear to get installed if they are not specified in the `InstallSpec`.
For the `luarocks` package, this results in a bunch of empty directories in the install tree.

mrcjkb:
Fixes #236.

### DIFF
--- a/rocks-lib/Cargo.toml
+++ b/rocks-lib/Cargo.toml
@@ -57,6 +57,7 @@ proptest = "1.4.0"
 dir-diff = "0.3.3"
 insta = { version = "1.41.0", features = ["redactions", "yaml", "json"] }
 assert_fs = "1.1.2"
+predicates = "3.1.2"
 
 [features]
 clap = ["dep:clap"]
--- a/rocks-lib/src/build/builtin.rs
+++ b/rocks-lib/src/build/builtin.rs
@@ -1,19 +1,22 @@
+use itertools::Itertools;
 use std::{
-    collections::HashMap,
+    collections::{HashMap, HashSet},
     path::{Path, PathBuf},
+    str::FromStr,
 };
+use walkdir::WalkDir;
 
 use crate::{
+    build::utils,
     config::Config,
     lua_installation::LuaInstallation,
     progress::{
         Progress::{self},
         ProgressBar,
     },
-    rockspec::{utils, Build, BuiltinBuildSpec, LuaModule, ModuleSpec},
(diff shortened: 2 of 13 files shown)


mrcjkb:
Actually, this was just a bug in `LuaModule::from_pathbuf`. No need to autodetect them in the install step.

I've also moved the `build.utils` module from `rockspec.build.utils` to `build.utils`.

mrcjkb:
We should also add a test case.

mrcjkb:
Turns out we should only auto detect modules if no modules are specified (for example, successfully auto detecting modules breaks `luassert`).

rocks-lib/src/build/builtin.rs
```diff
@@ -30,10 +32,11 @@ impl Build for BuiltinBuildSpec {
         progress: &Progress<ProgressBar>,
     ) -> Result<(), Self::Err> {
         // Detect all Lua modules
-        let modules = autodetect_modules(build_dir)
-            .into_iter()
-            .chain(self.modules)
-            .collect::<HashMap<_, _>>();
+        let modules = if self.modules.is_empty() {
```
vhyrro, line 35: question(non-blocking): is this luarocks's behaviour too? Is there no situation where a few modules are defined by the rockspec author and they "expect" the rest to be autodetected? Given the amount of wacky code we've seen it might be worth checking haha

rocks-lib/Cargo.toml
```diff
@@ -49,6 +49,7 @@ thiserror = "2.0.0"
 gpgme = "0.11.0"
 futures = "0.3.31"
 async-recursion = "1.1.1"
+predicates = "3.1.2"
```
vhyrro, line 52: issue: I only see this used in tests. Should this be part of `dev-dependencies` instead?

rocks-lib/src/build/builtin.rs
mrcjkb: I'm not sure. 
I tried installing `luassert` and it failed if using both the rockspec modules and the auto-detected modules.

mrcjkb: Done. Excluding paths from the `ModuleSpec` seems to solve the problem.

PR #248, opened 2024-12-05
tests: don't modify lockfile in sample tree

### LINKED ISSUE #245: [QOL] Don't keep modifying the lockfile in the sample tree
Something keeps modifying it (I'm guessing a test).
That causes a high potential for merge conflicts, which is a slight annoyance.

mrcjkb:
Stacked on #247.
Fixes #245.


vhyrro (dismissed):
Love the QOL improvements over the past few PRs, nice work :)

PR #253, opened 2024-12-10
feat: luarocks build backend

mrcjkb:


rocks-lib/src/luarocks_installation.rs
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    pub async fn install_build_dependencies(
+        self,
+        rockspec: &Rockspec,
+        progress: &Progress<ProgressBar>,
+    ) -> Result<(), InstallBuildDependenciesError> {
+        let mut lockfile = self.tree.lockfile()?;
+        let manifest = ManifestMetadata::from_config(&self.config).await?;
+        let build_dependencies = match rockspec.rockspec_format {
```
mrcjkb, line 126: - [ ] TODO: What if build dependencies themselves have dependencies?

(perhaps for another issue, as this is quite unlikely).

rocks-lib/src/build/utils.rs
```diff
@@ -27,6 +27,24 @@ pub(crate) fn copy_lua_to_module_path(
...
+                pathdiff::diff_paths(src.join(file.clone().into_path()), src)
+                    .expect("failed to copy directories!");
+            let filepath = file.path();
+            let target = dest.join(relative_src_path);
+            std::fs::create_dir_all(target.parent().unwrap())?;
+            std::fs::copy(filepath, target)?;
+        }
+    }
+    Ok(())
+}
```
vhyrro, line 47: nit(for a different PR): I believe we already vendor (or wanted to vendor) a library that has a recursive copy function with the ability to add globs to customize the output.

vhyrro (approved):
amazing work!

PR #264, opened 2024-12-18
feat: parallelize `update` and `remove`

vhyrro:
Converts both these functions into parallel versions, just like `install` did a week or two ago.

### DIFF
--- a/rocks-bin/src/remove.rs
+++ b/rocks-bin/src/remove.rs
@@ -1,43 +1,63 @@
+use std::io;
+
 use clap::Args;
-use eyre::Result;
+use eyre::{eyre, Result};
+use itertools::Itertools;
 use rocks_lib::{
     config::{Config, LuaVersion},
-    package::{PackageName, PackageSpec, PackageVersion},
+    package::PackageReq,
     progress::{MultiProgress, Progress},
-    remote_package_db::RemotePackageDB,
-    tree::Tree,
+    tree::{RockMatches, Tree},
 };
 
 #[derive(Args)]
 pub struct Remove {
     /// The name of the rock to remove.
-    name: PackageName,
-    /// The name of the version to remove.
-    version: Option<PackageVersion>,
+    packages: Vec<PackageReq>,
 }
 
+// TODO(vhyrro): Add `all:xyz>=2.0` JJ-like syntax.
 pub async fn remove(remove_args: Remove, config: Config) -> Result<()> {
-    let package_db = RemotePackageDB::from_config(&config).await?;
-
-    let target_version = remove_args
-        .version
-        .or(package_db.latest_version(&remove_args.name).cloned())
-        .unwrap();
-
     let tree = Tree::new(config.tree().clone(), LuaVersion::from(&config)?)?;
 
-    match tree.has_rock(
-        &PackageSpec::new(remove_args.name.clone(), target_version.clone()).into_package_req(),
-    ) {
-        Some(package) => Ok(rocks_lib::operations::remove(
-            package,
-            &config,
-            &Progress::Progress(MultiProgress::new().new_bar()),
-        )
-        .await?),
-        None => {
-            eprintln!("Could not find {}@{}", remove_args.name, target_version);
-            Ok(())
-        }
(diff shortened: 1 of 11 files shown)


rocks-lib/src/operations/remove.rs
```diff
@@ -15,27 +16,38 @@ pub enum RemoveError {
...
+        tokio::spawn(remove_impl(package, bar, config))
+    }))
+    .await;
+
+    Ok(())
 }
 
-async fn remove_impl(package: LocalPackage, config: &Config) -> Result<(), RemoveError> {
-    let tree = Tree::new(config.tree().clone(), LuaVersion::from(config)?)?;
+async fn remove_impl(
```
mrcjkb, line 41: suggestion: This could be renamed to `remove_pkg`

rocks-bin/src/remove.rs
```diff
@@ -30,9 +30,9 @@ pub async fn remove(remove_args: Remove, config: Config) -> Result<()> {
         &RemotePackage::new(remove_args.name.clone(), target_version.clone()).into_package_req(),
     ) {
         Some(package) => Ok(rocks_lib::operations::remove(
-            package,
```
mrcjkb, line 33: suggestion: let's change the CLI to take a non-empty list of packages.

mrcjkb (approved):

rocks-bin/src/remove.rs
vhyrro: sounds good, on it

mrcjkb: That's fine. I guess you  can pass in `foo@1.0 foo@2.0`.
We should filter out (and warn if found) any packages that are dependencies of other packages which aren't to be deleted, but that's out of scope for this PR.

vhyrro: Yep, we want to check to see if the package's id exists in the `entrypoints` table before deleting it.

jujutsu uses a special selector syntax whenever you want to operate on multiple values: `all:the-thing`. I find that interesting and maybe we could borrow that for our own delete command (e.g. `rocks remove all:neorg`).

mrcjkb: Hmm, wouldn't that be the same as `neorg@*`?

vhyrro: Not quite: `all:` says that you accept the fact that you're selecting many versions. `neorg@*` or something like `neorg>=7.0.0` with two versions of `neorg` installed would both error (as it could be a potentially unintended action). You'd have to do `all:neorg>=7.0.0` (borrowing jj's syntax here) to make it work.

mrcjkb: I see. Sure, why not?

vhyrro:
Implementation of `all:` is out of the scope of this PR but will be implemented for all major things like `pin`, `remove` or `info` :)

rocks-bin/src/remove.rs
```diff
@@ -1,43 +1,63 @@
...
+        |(mut p, mut n, mut d), rock_match| {
+            match rock_match {
+                RockMatches::NotFound(req) => n.push(req),
+                RockMatches::Single(package) => p.push(package),
+                RockMatches::Many(packages) => d.extend(packages),
+            };
+
+            (p, n, d)
+        },
+    );
```
vhyrro, line 40: note: I was hoping to use `partition_map` here but it only partitions into two collections :|

rocks-bin/src/remove.rs
```diff
@@ -1,43 +1,63 @@
...
     }
+
+    if !duplicate_packages.is_empty() {
+        // TODO(vhyrro): Greatly expand on this error message, notifying the user how it works and
+        // how to fix it.
+        return Err(eyre!(
+            "Multiple packages satisfying your version requirements were found. {:#?}",
+            duplicate_packages,
+        ));
+    }
```
vhyrro, line 57: I'm leaving these as-is because they can only be truly complete with the implementation of `all:`

mrcjkb:
Resolved all conflict markers.

PR #265, opened 2024-12-18
fix(build/make): pass `MakeBuildSpec::variables` to both passes

mrcjkb:


mrcjkb:
Merging, as this is trivial.

PR #266, opened 2024-12-18
feat(build): cmake build backend

mrcjkb:


mrcjkb:
weird... the `run_busted_test` failure on Darwin seems unrelated.

vhyrro (dismissed):

PR #268, opened 2024-12-20
fix: duplicated progress bars

mrcjkb:
I think cloning a `MultiProgress` could be leading to strange behaviour.
This should fix it.


vhyrro (approved):
This works and is perfectly fine, but maybe in a future refactor we could instead hold an `Arc<T>` internally for `Progress<T>` and implement `Copy` for Progress? That should make the code generally simpler, I'll make an issue for that :)

PR #270, opened 2024-12-21
fix: make CLI experience for `rocks new` much better

vhyrro:
This change forces the user to provide a directory (allowing for `rocks new .`) and optionally creates the directory and the rockspec there.

This is much more consistent with other package managers and also much more predictable. I also fixed inferring the directory name by making the path absolute first.


rocks-bin/src/project/mod.rs
```diff
@@ -1 +1,315 @@
...
+use spdx::LicenseId;
+use spinners::{Spinner, Spinners};
+
+use crate::utils::github_metadata::{self, RepoMetadata};
+use rocks_lib::{package::PackageReq, project::Project};
+
+// General notes and ideas:
+// - Create a subdirectory for the project, unless a path is explicitly passed.
+// - Automatically detect build type by inspecting the current repo (is there a Cargo.toml? is
+//   there something that tells us it's a lua project?).
```
mrcjkb, line 20: question: I suppose these notes are outdated?

mrcjkb (dismissed):

rocks-bin/src/project/mod.rs
vhyrro: Yep, the first one is for sure.

vhyrro: resolved :)

mrcjkb (approved):

PR #276, opened 2024-12-22
feat: `rocks debug project`

vhyrro:
This adds a new debug command to print out some very basic information on the current project.

This is useful for the tutorial, so a user can check if their project works. It also allows you to do basic linting on the rockspec (with errors for the invalid URL etc.)


rocks-bin/src/project/debug.rs
vhyrro, line 1: note: this is the only file worth reviewing, everything else is just file moves :)

mrcjkb (approved):

PR #278, opened 2024-12-22
feat: don't reinstall dependencies that are already installed

mrcjkb:
- In `resolve::get_all_dependencies`, we filter out dependencies that are already installed, as long as the `build_behaviour` is `NoForce`.
- We set the build behaviour to `NoForce` for dependencies.
  This means that we don't reinstall dependencies when reinstalling a rock. (Is this desirable behaviour? :thinking:).

### DIFF
--- a/rocks-lib/src/lockfile/mod.rs
+++ b/rocks-lib/src/lockfile/mod.rs
@@ -485,6 +485,26 @@ impl Lockfile {
 
         Ok(())
     }
+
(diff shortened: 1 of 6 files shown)


vhyrro:
I think the dependencies should inherit the build behaviour of the actual package, that way we forcefully reinstall all dependencies too if need be :)

mrcjkb:
> I think the dependencies should inherit the build behaviour of the actual package, that way we forcefully reinstall all dependencies too if need be :)

Sounds fine to me.

vhyrro (approved):

PR #279, opened 2024-12-23
feat: man pages and shell completions

### LINKED ISSUE #57: [Feature] man pages and shell completions
See for example: https://github.com/nix-community/nix-init/blob/main/build.rs

mrcjkb:
Closes #57.


mrcjkb:
Looks like `run_lua::test::test_run_lua` is failing in the nix build (it used to be skipped because it wasn't a `lib` test).
It succeeds locally and shouldn't need an internet connection. I'll have to investigate this 👀

vhyrro (approved):

PR #283, opened 2024-12-28
fix!: prepend paths instead of appending them

### LINKED ISSUE #281: [Bug] prepend to `PATH` environment variable, instead of appending
`rocks run` and other related commands (e.g. `rocks test`) append to the `PATH`.
We should prepend instead, so that packages managed by rocks are prioritized.

mrcjkb:
Closes #281.

Also changes the `--append` flag for the `rocks path` command to `--prepend`.


max397574 (approved):
lgtm

vhyrro (approved):

Issue #159, opened 2024-10-23
[Feature] Support searching/installing packed rocks from binary servers


mrcjkb:
e.g. 

- `rocks --server "https://nvim-neorocks.github.io/rocks-binaries/" install foo`
 - `rocks --server "https://nvim-neorocks.github.io/rocks-binaries/" --no-default-servers install foo`

PR #285, opened 2024-12-29
feat!: support installing from multiple servers

mrcjkb:
In preparation for #159.
This PR makes the `Manifest` and `ManifestMetadata` types `pub(crate)`. `rocks-lib` now only exposes the `RemotePackageDB` type (which could one day be a client to a non-luarocks server).


vhyrro (approved):
Big fan of this refactor: this makes the library interface much more intuitive for the average user, prepares luanox integration and makes things like `db.find()` look very elegant.
Good job :D

PR #289, opened 2024-12-29
feat: `rocks check`

vhyrro:
Adds a simple implementation of `rocks check`, which ensures `luacheck` is installed and runs it for the whole source directory of a given project.

Similarly to `rocks fmt`, we'll want to expand this to take in a path as an argument for finer control.
I don't believe we should let the user control `luacheck` with flags, that should be up to the luacheckrc file.

### DIFF
--- a/rocks-bin/src/check.rs
+++ b/rocks-bin/src/check.rs
@@ -1,0 +1,42 @@
+use eyre::{OptionExt, Result};
+use rocks_lib::{
+    build::BuildBehaviour,
+    config::{Config, LuaVersion},
+    lockfile::PinnedState::Pinned,
+    operations::{self, install},
+    progress::MultiProgress,
+    project::Project,
+    remote_package_db::RemotePackageDB,
+};
+
+pub async fn check(config: Config) -> Result<()> {
+    let project = Project::current()?.ok_or_eyre("Not in a project!")?;
+
+    let db = RemotePackageDB::from_config(&config).await?;
+
+    install(
+        vec![(BuildBehaviour::NoForce, "luacheck".parse()?)],
+        Pinned,
+        &db,
+        &config,
+        MultiProgress::new_arc(),
+    )
+    .await?;
+
+    operations::run(
+        "luacheck",
+        vec![
+            project.root().to_string_lossy().into(),
+            "--exclude-files".into(),
+            project
+                .tree(LuaVersion::from(&config)?)?
+                .root()
+                .to_string_lossy()
+                .to_string(),
+        ],
+        config,
+    )
+    .await?;
+
+    Ok(())
+}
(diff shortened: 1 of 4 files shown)


rocks-bin/src/check.rs
```diff
@@ -0,0 +1,38 @@
...
+    progress::MultiProgress,
+    project::Project,
+};
+
+pub async fn check(config: Config) -> Result<()> {
+    let project = Project::current()?.ok_or_eyre("Not in a project!")?;
+
+    let src_dir = project.src_dir().ok_or_eyre(
+        "No source directory found! Expected a `src/` or `lua/` directory at the project root.",
+    )?;
```
mrcjkb, line 17: question: why not the project root?
People might want to check `spec` or `test` files too.

rocks-bin/src/check.rs
```diff
@@ -0,0 +1,38 @@
...
+
+    let manifest = Manifest::from_config(config.server(), &config).await?;
```
mrcjkb, line 19: After #285, this should be a `RemotePackageDB`.

vhyrro:
Will fix the failing build in a bit.

rocks-bin/src/check.rs
vhyrro: That's a fair point, I haven't thought of that. Let's do just the root. Should I throw out the `src_dir()` function then if it's not used?

rocks-bin/src/check.rs
vhyrro: Yeah this was a rebase mistake on my end.

rocks-bin/src/check.rs
mrcjkb: > Should I throw out the src_dir() function then if it's not used?

Yeah, let's :+1:

vhyrro: Oh, found a problem. Just setting the project root will also search `.rocks` hahah. We want to limit it somehow. There should be an `--exclude` flag or whatever.

rocks-bin/src/check.rs
```diff
@@ -0,0 +1,42 @@
...
+        "luacheck",
+        vec![
+            project.root().to_string_lossy().into(),
+            "--exclude-files".into(),
+            project
+                .tree(LuaVersion::from(&config)?)?
+                .root()
+                .to_string_lossy()
+                .to_string(),
+        ],
```
mrcjkb, line 36: suggestion:

```suggestion
        vec![
            format!("{}", project.root().display()),
            "--exclude-files".into(),
            format!("{}", project
                .tree(LuaVersion::from(&config)?)?
                .root()
                .display()),
        ],
```

mrcjkb (approved):

rocks-bin/src/check.rs
mrcjkb: Not sure if that's actually better, but apparently `display()` is safer.

vhyrro: `display()` does essentially the same thing as `to_string_lossy` by replacing all invalid unicode chars with the replacement character. Amounts to the same code in the end :)

PR #291, opened 2024-12-29
feat!: support installing binary `.rock` packages

### LINKED ISSUE #159: [Feature] Support searching/installing packed rocks from binary servers

mrcjkb:
Stacked on #297.

Closes #159.

This has a few major differences compared to the luarocks implementation:

1. We ignore the file extension (as we don't support installing packed rocks by passing in a file path).
1. When constructing the manifest, we filter out entries that don't match the current platform.
    Luarocks lets you set the platform using the config. We don't. And I don't think we should.
    Instead, we should add a config option to ignore `.rock` or `.src.rock` or `.rockspec` entries when searching for rocks (#294).


mrcjkb:
TODO:

- [ ] `build::build_packed_rock` function

mrcjkb:
Marking as draft.
I'm going to rebase this on #297 and modify it so that the `RemotePackageSource` contains the information about whether the package is a `.src.rock`, a `.rock` or a `.rockspec`. We'll have to store that information in the lockfile, too.

vhyrro (approved):

PR #297, opened 2025-01-01
feat!: add remote package source to lockfile

mrcjkb:
This PR introduces a `RemoteSource` type, which contains information about what type of source (e.g. a Luarocks server) and the URL a `LocalPackage` was pulled from.

The new field has to be an `Option`, because it's possible to run `rocks build` on a local project.


mrcjkb:
I'm stuck on the fact that the `LocalPackage` is constructed in the `build` function, but the `RemoteSource` struct is `pub(crate)`.
I don't want to make it `pub`, so I'll start work on https://github.com/nvim-neorocks/rocks/issues/293.

mrcjkb:
This ended up getting a lot bigger than I originally anticipated (sorry for that :disappointed:).

Here's a summary of the changes:

- Changed `Config::server()` and `::extra_servers()` to return `Url`s instead of `String`s.
- Introduced a `RemotePackageSource` type, which provides information about where a remote package was fetched from.
  Currently, there are two variants:
    - `LuarocksServer` - with a URL
    - `RockspecContent` - for packages that were built from a rockspec on a local filesystem
- We serialize the `RemotePackageSource` and store it in the lockfile [this is a breaking change!].
  - [ ] Question: Do we want to base64 encode the `RockspecContent`, to make it more compact?
        My gut feeling says no, because it wouldn't be good for git diffs.
- The introduction of `RemotePackageSource` and its presence as a field in `LocalPackage` has resulted in a clippy `large_enum_variant` warning for `RockMatches`, which is why I have modified it to hold `LocalPackageId`s instead.
- Because `RemotePackageSource` is private, the `LocalPackage` constructor is now also private, so a bunch of public APIs that used to take a `LocalPackage` now take a `LocalPackageId` and look up the package from the `Tree` (for example, `Remove`).

vhyrro (approved):

PR #299, opened 2025-01-01
refactor!(install): use builder pattern

mrcjkb:
Stacked on #298


rocks-lib/src/operations/install.rs
```diff
@@ -20,6 +20,81 @@ use thiserror::Error;
...
+    progress: Option<Arc<Progress<MultiProgress>>>,
+}
+
+impl<'a> Install<'a> {
+    /// Construct a new installer.
+    pub fn new(config: &'a Config) -> Self {
+        Self {
+            config,
+            package_db: None,
+            packages: Vec::new(),
```
vhyrro, line 40: issue: i think the builder pattern doesn't quite fit in install's case since one can install an empty set of packages by simply running `Install::new().install()` (which makes little sense). Can we enforce at the type level that at least one package should be provided (without having to do runtime checks in `install()`)?

rocks-lib/src/operations/install.rs
```diff
@@ -20,6 +20,81 @@ use thiserror::Error;
...
 
+/// A rocks package installer, providing fine-grained control
+/// over how packages should be installed.
+/// Can install multiple packages in parallel.
+pub struct Install<'a> {
+    config: &'a Config,
+    package_db: Option<RemotePackageDB>,
+    packages: Vec<(BuildBehaviour, PackageReq)>,
+    pin: PinnedState,
+    progress: Option<Arc<Progress<MultiProgress>>>,
```
vhyrro, line 31: nitpick: I just realized this exists in the other PRs (including the earlier PR that was merged) as well: we don't need to wrap this in an option, since the whole purpose of `Progress<T>` was to allow us to pass `NoProgress` :p

mrcjkb: The reason this is wrapped in an `Option` is so that we can fall back to default behaviour if the caller doesn't pass in one.
(right now, the default behaviour is to create a new `MultiProgress`).

rocks-lib/src/operations/install.rs
mrcjkb: This is also possible without the builder pattern, by passing in an empty vec.
We could change the `new` constructor to take a [`NonEmpty<PackageReq>`](https://docs.rs/nonempty/latest/nonempty/).

I'd be in favour of that, but that would be for a separate PR.

vhyrro (approved):

rocks-lib/src/operations/install.rs
mrcjkb: Come to think of it, sometimes we want it to be possible for the list to be empty, for example when installing test dependencies.

PR #301, opened 2025-01-02
feat!(update/remove): parallelize + use builder pattern

mrcjkb:
Stacked on #300.

@vhyrro I've pulled in the changes from #264, because it keeps getting conflicted.
It's still missing the `:all` syntax, but I'd suggest you add that in a separate PR :smile:


vhyrro (approved):
Awesome, I'll close my PR after this gets merged :)

PR #304, opened 2025-01-02
refactor!(run,upload): use builder pattern

mrcjkb:
Stacked on #303


mrcjkb:
Oops, forgot to `jj new` before the upload changes.

vhyrro (approved):

PR #310, opened 2025-01-05
fix(rocks-bin): remove duplicate code

mrcjkb:


mrcjkb:
Merging without review, as this is trivial.

PR #312, opened 2025-01-05
refactor(build): move builder to `derive-builder`

vhyrro:
This moves the `build` builder to derive-builder instead, giving us full runtime checks.

Feel free to neglect or nitpick these changes, I'm not sure if this is something we want in our codebase. It seems to work very well though and integrates with our own errors fairly well.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -684,8 +684,18 @@ version = "0.14.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7b750cb3417fd1b327431a470f388520309479ab0bf5e323505daf0290cd3850"
 dependencies = [
- "darling_core",
- "darling_macro",
+ "darling_core 0.14.4",
+ "darling_macro 0.14.4",
+]
+
+[[package]]
+name = "darling"
+version = "0.20.10"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6f63b86c8a8826a49b8c21f08a2d07338eec8d900540f8630dc76284be802989"
+dependencies = [
+ "darling_core 0.20.10",
+ "darling_macro 0.20.10",
 ]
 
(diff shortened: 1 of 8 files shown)


mrcjkb (commented):
:thinking: I definitely like the derived builder, but I'm not a fan of returning `UninitializedFieldError`s at runtime instead of having compile type safety.

Can we keep the `::new` constructors with the required fields and somehow get rid of the `::default` constructors?

vhyrro:
The problem with the compile time version is you can't override the `build()` function (so we'd need a different approach). You can make the build function private and then hope to create your own, but then adding your own `build()` looks like... this:

```rust
#[allow(dead_code, non_camel_case_types, missing_docs)]
impl<
        'a,
        __progress: ::typed_builder::Optional<&'a Progress<ProgressBar>>,
        __pin: ::typed_builder::Optional<PinnedState>,
        __constraint: ::typed_builder::Optional<LockConstraint>,
        __behaviour: ::typed_builder::Optional<BuildBehaviour>,
        __source: ::typed_builder::Optional<Option<RemotePackageSource>>,
    >
    BuildBuilder<
        'a,
        (
            (&'a Rockspec,),
            (&'a Config,),
            __progress,
            __pin,
            __constraint,
            __behaviour,
            __source,
        ),
    >
{
```

`typed-builder` allows specifying an `into = ...` field, but `Into<Result<X, Y>> for Build` isn't `async`...

As for customizing the function name to be `new()`, I can't find anything in the documentation for `derive_builder`. `typed-builder` does have such a feature, but not here apparently.

I do wonder if we could make a temp fork of `typed-builder` and do a quickfix for the custom build function.

vhyrro:
I'll try making a quick fork, see how easy it is

mrcjkb:
I'd choose type safety over less boilerplate anytime.

PR #314, opened 2025-01-05
refactor(build): move custom builder syntax to `bon`

vhyrro:
Supercedes #312 with a much better (compile-time checked) implementation :)


vhyrro:
We could make `do_build` take in our struct now for a nicer API :)

mrcjkb (dismissed):

PR #321, opened 2025-01-07
refactor!: remove `Drop` for `Lockfile`

vhyrro:


vhyrro:
Can't really pinpoint the workflow errors on my end. Is it a CI issue?

mrcjkb:
> Can't really pinpoint the workflow errors on my end. Is it a CI issue?

There are some clippy warnings and it looks like the `luarocks_make` integration test is failing.

vhyrro:
one last easy fix is required for CI to pass, I'll deal with that tomorrow.

.pre-commit-config.yaml
```diff
@@ -0,0 +1 @@
+/nix/store/0qn5cx859h6b2hdpll99qzb39j7lvfc6-pre-commit-config.json
```
mrcjkb, line 1: issue: accidental symlink addition?
This should probably go in .gitignore. I don't know why it wasn't already in there.

mrcjkb (changes requested):

.pre-commit-config.yaml
vhyrro: true, let me fix that

mrcjkb (approved):

PR #325, opened 2025-01-12
feat!: `rocks.toml` support

vhyrro:
Stacked on #321.

This PR aims to replace the entire rockspec-based workflow to a TOML based one instead. The rationale behind this decision should be nicely laid out in the documentation site :)

Not complete yet:
- [ ] Add `extra.rockspec` support
- [x] Add code to convert `rocks.toml` to a rockspec
- [x] Ensure all existing code properly uses the new API


rocks-lib/src/project/rocks.rs
```diff
@@ -0,0 +1,79 @@
...
+    version: PackageVersion,
+    description: RockDescription,
+    supported_platforms: PlatformSupport,
+    dependencies: PerPlatform<Vec<PackageReq>>,
+    build_dependencies: PerPlatform<Vec<PackageReq>>,
+    external_dependencies: PerPlatform<HashMap<String, ExternalDependencySpec>>,
+    test_dependencies: PerPlatform<Vec<PackageReq>>,
+    //source: PerPlatform<RockSource>,
+    //pub build: BuildSpec,
+    //pub test: TestSpec,
```
vhyrro, line 26: @mrcjkb I'm a little lost on how to properly implement `Deserialize` for `RockSource` and the others without everything erroring. Is trying to make `RockSource` implement `Deserialize` flawed? Should some other work be done on `PerPlatform`/`PerPlatformWrapper` instead?

mrcjkb: You'll probably want to do something similar to the `impl<T> FromLua for PerPlatform<T>`.

vhyrro:
Things to fix the build failures:
- BuildSpec has no PartialOverride
- Only `Rockspec` implements the `binaries()` function, whereas it should be a part of the `Rocks` trait impl

vhyrro:
The more I further this implementation the more I feel I'm going insane. If you could do a prophylactic check of everything at some point today I'd appreciate @mrcjkb :joy:

vhyrro:
The general gist of this PR is:
- Create a generic `Rocks` trait that both `Rockspec` and `RocksToml` implement
- Make every part of the code use `impl Rocks` instead of a specific implementation
- Because the `source.url` can be optional in the TOML, I need to implement some extra serialization logic to make it work (and ensure that the conversion from TOML to rockspec is valid)

The rest of the PR is mostly accommodating these changes.

vhyrro:
Well good news the code at least builds now.

rocks-lib/src/project/mod.rs
```diff
@@ -38,23 +41,23 @@ impl Project {
         }
 
         match find_up_with(
-            "project.rockspec",
+            "rocks.toml",
```
mrcjkb, line 44: suggestion: let's put this in a constant (I think there's one already).

rocks-lib/src/rockspec/mod.rs
```diff
@@ -46,26 +47,117 @@ pub enum RockspecError {
...
+    test: PerPlatform<TestSpec>,
     /// The original content of this rockspec, needed by luarocks
-    pub raw_content: String,
+    raw_content: String,
     /// The sha256 of this rockspec
     hash: Integrity,
 }
 
+/// A trait for querying information about a project from either a rockspec or `rocks.toml` file.
+pub trait Rocks {
```
mrcjkb, line 71: suggestion: I find `Rocks` a bit confusing to read, as it doesn't convey much information about what it is.
How about we name the trait `RockSpec` and rename `Rockspec` to `LuaRockSpec`?

rocks-lib/src/project/rocks.rs
```diff
@@ -0,0 +1,221 @@
...
+                .map(|(name, version_req)| PackageReq { name, version_req })
+                .collect()
+        })
+    }))
+}
+
+#[derive(Debug, Error)]
+pub enum IntoRockspecError {
+    #[error("no source url provided")]
+    NoSourceUrl,
```
mrcjkb, line 75: suggestion: Perhaps we could define a spec for generating a source URL for releases and scm versions, with variable expansion:

```toml
[source.scm]
url = "git://github.com/nvim-neorocks/$(PACKAGE)"

[source.release]
url = "https://github.com/nvim-neorocks/$(PACKAGE)/archive/v$(VERSION).zip"
```

For building a local project, we replace the `SourceSpec` with one that points to the project root.

rocks-lib/src/project/rocks.rs
```diff
@@ -0,0 +1,221 @@
...
+    fn format(&self) -> &Option<RockspecFormat> {
+        &Some(RockspecFormat::_3_0)
+    }
+}
+
+impl HasIntegrity for RocksToml {
+    fn hash(&self) -> std::io::Result<Integrity> {
+        Ok(self.hash.to_owned())
+    }
+}
```
mrcjkb, line 221: question: is this needed, if we're only uploading rockspecs?

rocks-lib/src/rockspec/mod.rs
```diff
@@ -46,26 +47,117 @@ pub enum RockspecError {
...
+pub trait LuaVersionCompatibility {
+    fn validate_lua_version(&self, config: &Config) -> Result<(), LuaVersionError>;
+
+    fn lua_version_matches(&self, config: &Config) -> Result<LuaVersion, LuaVersionError>;
+
+    fn supports_lua_version(&self, lua_version: &LuaVersion) -> bool;
+
+    fn lua_version(&self) -> Option<LuaVersion>;
+
+    fn test_lua_version(&self) -> Option<LuaVersion>;
```
mrcjkb, line 116: suggestion: Some of these have very similar names. Let's add some doc strings.

rocks-lib/src/project/rocks.rs
```diff
@@ -0,0 +1,221 @@
...
+
+    fn build_mut(&mut self) -> &mut PerPlatform<BuildSpec> {
+        &mut self.build
+    }
+
+    fn test_mut(&mut self) -> &mut PerPlatform<TestSpec> {
+        &mut self.test
+    }
+
+    fn to_rockspec_str(&self) -> Result<String, IntoRockspecError> {
```
mrcjkb, line 208: quibble: `_str` implies it returns a `&str`, but it returns `String`.

rocks.lock
```diff
@@ -0,0 +1,35 @@
+{
```
mrcjkb, line 1: issue: accidental commit?

rocks-bin/src/build.rs
```diff
@@ -13,6 +13,7 @@ use rocks_lib::{
     progress::MultiProgress,
     project::Project,
     remote_package_db::RemotePackageDB,
+    rockspec::{LuaVersionCompatibility as _, Rocks as _},
```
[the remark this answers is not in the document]
vhyrro: that is a question for rust analyzer :sweat_smile:, I'll sed through the codebase.

vhyrro:
Damn, the scope keeps growing larger :joy:. Here are the things left to do:
- [ ] Create a struct for nicer serialization of the TOML to a rockspec
- [ ] Translation of the `build` table from TOML to rockspec (this one is going to be a blast :cry:)
- [ ] Moar tests
- [ ] `extra.rockspec`

After these are done I'll go through one massive cleanup of unused code and struct/impl reordering.

vhyrro:
If tests pass, then the code is complete. I'll now go and trim all the unnecessary code snippets (trait impls that are no longer needed) before marking as ready!

vhyrro:
@mrcjkb feel free to do a full review now! I'm excluding `extra.rockspec`, as that requires even more codebase changes, which we'd better split off into a separate pr :p

vhyrro:
oh, also. this currently *compiles*, and the tests *pass*, but I haven't done real world testing on actual projects with this fella (to see how the UX is). Kind of silly of me to forget that, actually.

rocks-lib/src/lua_rockspec/dependency.rs
```diff
@@ -0,0 +1,105 @@
...
+
+        if map.contains_key("header") == map.contains_key("library") {
+            return Err(serde::de::Error::custom(ConflictingExternalDependencySpec));
+        }
+
+        match map.into_iter().next() {
+            Some((key, value)) => match key.as_str() {
+                "header" => Ok(ExternalDependencySpec::Header(value)),
+                "library" => Ok(ExternalDependencySpec::Library(value)),
+                _ => unreachable!(),
```
mrcjkb, line 45: question: Is this really unreachable?

rocks-lib/src/lua_rockspec/serde_util.rs
```diff
@@ -68,3 +70,66 @@ where
...
+    List(Vec<Self>),
+    Table(Vec<DisplayLuaKV>),
+}
+
+pub(crate) struct DisplayLuaKV {
+    pub(crate) key: String,
+    pub(crate) value: DisplayLuaValue,
+}
+
+impl Display for DisplayLuaValue {
```
mrcjkb, line 89: suggestion: some tests would be nice.

rocks-lib/src/lua_rockspec/serde_util.rs
```diff
@@ -68,3 +70,66 @@ where
...
+            //DisplayLuaValue::Nil => write!(f, "nil"),
+            //DisplayLuaValue::Number(n) => write!(f, "{}", n),
+            DisplayLuaValue::Boolean(b) => write!(f, "{}", b),
+            DisplayLuaValue::String(s) => write!(f, "\"{}\"", s),
+            DisplayLuaValue::List(l) => {
+                writeln!(f, "{{")?;
+
+                for item in l {
+                    writeln!(f, "{},", item)?;
+                }
```
mrcjkb, line 101: suggestion: (perhaps for a follow-up PR).

We could use a crate like [`pretty`](https://docs.rs/pretty/latest/pretty/) that should be able to handle things like indentation
(not sure if this crate does that - the Haskell `pretty` package does).

rocks-lib/src/operations/install.rs
```diff
@@ -109,8 +110,8 @@ impl<'a> Install<'a> {
 pub enum InstallError {
     #[error(transparent)]
     SearchAndDownloadError(#[from] SearchAndDownloadError),
-    #[error(transparent)]
-    LuaVersionError(#[from] LuaVersionError),
+    //#[error(transparent)]
+    //LuaVersionError(#[from] LuaVersionError),
```
mrcjkb, line 114: question: can this be removed?

mrcjkb (changes requested):
Nice work :rocket: 

Just a typo and some suggestions :)

rocks-lib/src/lua_rockspec/serde_util.rs
vhyrro: Added a test for the DisplayLuaValue to make sure it's all working nicely, I think it should suffice, but do let me know if I should extend it some way.

rocks-lib/resources/test/sample-project-busted/rocks.toml
```diff
@@ -0,0 +1,9 @@
+package = "sample-project-busted"
+version = "0.1.0"
+lua = "5.1"
+
+[source]
+url = "https://github.com/nvim-neorocks/luarocks-stub"
+
```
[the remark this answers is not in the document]
vhyrro: These `[source]`s need to be here, as some test fail without them. Whenever you convert the toml into a validated toml it enforces that the `url` be provided (just like is required from a regular rockspec), and errors otherwise :)

mrcjkb (approved):

PR #331, opened 2025-01-13
feat(rocks-bin): rocks uninstall command

mrcjkb:
We already have a `remove` command.

### DIFF
--- a/rocks-bin/src/lib.rs
+++ b/rocks-bin/src/lib.rs
@@ -18,6 +18,7 @@ use run::Run;
 use run_lua::RunLua;
 use search::Search;
 use test::Test;
+use uninstall::Uninstall;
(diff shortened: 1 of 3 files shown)


vhyrro:
There is actually a reason for this, but it's unimplemented rn: `add` and `install` are different, and so is `remove` and `uninstall`. The former versions operate on a project, whereas the latter versions are supposed to work on the user-wide tree :p

vhyrro:
Currently yes, although I expect that to change right after #325 :)

mrcjkb:
> Currently yes, although I expect that to change right after #325 :)

Gotcha. Updated the PR to implement `rocks uninstall`.

vhyrro (dismissed):

PR #334, opened 2025-01-13
fix!: remove associated binaries when deleting a rock

vhyrro:
This PR keeps track of all binaries installed by a given rock and properly cleans up the binaries when running `rocks remove`.

I also made sure to sanitize the path input before reading its filename just in case to prevent any sort of DOS. Whether it's as effective as I hope should be looked into as part of a security check.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,5 +14,6 @@
 /*.rockspec
 *.snap.new
 /.rocks
+rocks.lock
 
 .pre-commit-config.yaml
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -527,6 +527,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "clean-path"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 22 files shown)


rocks.lock
vhyrro, line 1: issue: this should be added to gitignore shouldn't it :p

vhyrro:
ah, I must update the tests to use the new syntax, was wondering where the failures were coming from :joy:

mrcjkb:
todo: the `builtin` build type auto-detects and installs binaries.
We should change the signature to return a `BuildInfo` that contains additional paths that were installed.

vhyrro:
Making this a draft so i can continue work tomorrow on laptop

mrcjkb:
> todo: the `builtin` build type auto-detects and installs binaries. We should change the signature to return a `BuildInfo` that contains additional paths that were installed.

@vhyrro you marked this as ready, but this doesn't seem to be implement

vhyrro:
I assumed `todo:` meant "for a future PR", but I'll have some time in an hour to implement that :)

mrcjkb (approved):

PR #338, opened 2025-01-14
feat(ui): show spinner while downloading manifest

mrcjkb:
...at least this will prevent people from thinking rocks is dead if downloading the manifest on a slow connection.


vhyrro:
seems like you accidentally rebased this on the codefreeze demo :p

vhyrro (approved):

teto:
dope !

PR #339, opened 2025-01-14
feat(rocks-bin): `rocks doc` command

mrcjkb:


rocks-bin/src/doc.rs
```diff
@@ -0,0 +1,118 @@
...
+            }
+        }
+    } else if files.len() == 1 {
+        open::that(layout.doc.join(files.first().unwrap()))?;
+        Ok(())
+    } else {
+        let file = Select::new(
+            "Multiple documentation files found. Please select one to open.",
+            files,
+        )
```
vhyrro, line 112: praise: this is an interesting solution to the problem that existed in luarocks, nice one.

PR #341, opened 2025-01-14
feat: `rocks which` command

mrcjkb:


mrcjkb:
TODO:

- [ ] Some more test cases (e..g. searching for modules in individual packages).
- [ ] Implement `rocks which` command in `rocks-bin`.

vhyrro (approved):

PR #342, opened 2025-01-16
feat(build): apply rockspec patches

### LINKED ISSUE #99: [Feature] Apply rockspec patches
After fetching the source and unpacking it to a temp directory, we should apply any patches that are defined in the rockspec.
It would be nice if we could support more sophisticated patches than the hand-rolled unified diff patches that luarocks supports.

mrcjkb:
Closes #99.

### DIFF
--- a/rocks-lib/src/build/patch.rs
+++ b/rocks-lib/src/build/patch.rs
@@ -1,0 +1,234 @@
+use crate::progress::{Progress, ProgressBar};
+use bon::Builder;
+use diffy::{self, ApplyError, ParsePatchError};
+use std::io;
+use std::{collections::HashMap, path::PathBuf};
+use thiserror::Error;
+
+#[derive(Builder)]
+#[builder(start_fn = new, finish_fn(name = _build, vis = ""))]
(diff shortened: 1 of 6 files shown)


mrcjkb:
It would be nice to parse the patches when passing the rockspec, but that requires us to add lifetime parameters everywhere. 
Maybe for another PR...

rocks-lib/src/build/patch.rs
```diff
@@ -0,0 +1,234 @@
...
+        err: io::Error,
+    },
+}
+
+pub(crate) fn do_apply(args: Patch<'_>) -> Result<(), PatchError> {
+    for (path, patch_str) in args.patches {
+        args.progress
+            .map(|bar| bar.set_message(format!("Applying patch {}", path.display())));
+        let patch = dbg!(diffy::Patch::from_str(patch_str)
+            .map_err(|err| PatchError::Parse(path.clone(), err))?);
```
vhyrro, line 67: issue: sneaky debug call :)

vhyrro (approved):
other than that LGTM!

PR #348, opened 2025-01-20
fix: proper parsing of `PlatformSupport`

vhyrro:
This pull request includes changes to the `PlatformSupport` code - the changes improve the platform parsing logic by distinguishing between exclusive and inclusive platform matches. These were improperly handled earlier but still somehow passed our tests (maybe because we never directly compared the structs themselves to my knowledge? idk)

Improvements to platform parsing logic:

* Modified the condition to check for any negative matches instead of all negative matches to determine if platforms should be matched exclusively.
* Simplified the inclusive matching branch to only validate positive matches.

### DIFF
--- a/rocks-lib/src/rockspec/platform.rs
+++ b/rocks-lib/src/rockspec/platform.rs
@@ -191,18 +191,24 @@ impl PlatformSupport {
     }
 
     pub fn parse(platforms: &[String]) -> Result<Self, PlatformValidationError> {
+        // Platforms are matched in one of two ways: exclusively or inclusively.
+        // If only positive matches are present, then the platforms are matched inclusively (as you only support the matches that you specified).
+        // If any negative matches are present, then the platforms are matched exclusively (as you want to support any operating system *other* than the ones you negated).
         match platforms {
             [] => Ok(Self::default()),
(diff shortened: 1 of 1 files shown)


vhyrro:
Fwiw this regression case is already handled during the comparison of the rocks TOML and the rockspec itself - do we want a custom regression test for this anyway? Unsure how we'd properly check for it though...

mrcjkb:
> Fwiw this regression case is already handled during the comparison of the rocks TOML and the rockspec itself (in the rocks.toml PR) - do we want a custom regression test for this anyway? Unsure how we'd properly check for it though...

I guess a test case would be something like

```lua
platforms = {
  "!windows",
  "linux",
}
```

-> assert that macos is supported?

But I don't really think that would make a good test case, as it tests the implementation.
The example itself is undefined by the spec (perhaps it should cause a warning to be printed).
So if this regression is handled by the rocks.toml/rockspec comparison, we're good without any tests.

mrcjkb (approved):
Probably just needs a fustfmt.

vhyrro:
CI failure is unrelated to the code changes lol

mrcjkb:
> CI failure is unrelated to the code changes lol

Looks like a flake due to a network error.

PR #357, opened 2025-01-23
feat: `rocks pack`

mrcjkb:


mrcjkb:
TODO:

- [ ] Rename `<name>-<version>.rockspec` to `package.rockspec` and vice-versa, when packing/unpacking.
- [ ] Roundtrip test (pack -> unpack)
- [ ] Test case that packs a rock and uses a `LuaRocksInstallation` to install it.

vhyrro:
Code looks great to me: I guess I have a UX question - having to provide a path to a `rocks.toml` is inconsistent with the rest of the CLI where you can just omit the path, which means "operate on the current project". We probably want to support that here as well.

mrcjkb:
> Code looks great to me: I guess I have a UX question - having to provide a path to a `rocks.toml` is inconsistent with the rest of the CLI where you can just omit the path, which means "operate on the current project". We probably want to support that here as well.

Done.

vhyrro (approved):

vhyrro:
Once tests succeed, let's merge and  rebase the `rocks add` and `extra.rockspec` PRs on top of this.

PR #362, opened 2025-01-27
feat: `rocks add`

vhyrro:
Summary: rocks add and refactors existing code to enforce the handling of unsupplied package version requirements.

### DIFF
--- a/rocks-bin/src/add.rs
+++ b/rocks-bin/src/add.rs
@@ -1,0 +1,86 @@
+use eyre::{OptionExt, Result};
+use rocks_lib::{
+    config::{Config, LuaVersion},
+    lockfile::PinnedState,
+    operations,
+    package::PackageReq,
+    progress::MultiProgress,
+    project::Project,
+};
+
+use crate::utils::install::apply_build_behaviour;
+
+// TODO: Make `rocks add thing --build lots of stuff here --test more stuff here` work
+
+#[derive(clap::Args)]
+pub struct Add {
+    /// Package or list of packages to install.
+    package_req: Vec<PackageReq>,
+
+    /// Pin the package so that it doesn't get updated.
+    #[arg(long)]
+    pin: bool,
+
+    /// Reinstall without prompt if a package is already installed.
+    #[arg(long)]
+    force: bool,
+
+    /// Install the package as a development dependency.
+    /// Also called `dev`.
+    #[arg(short, long, alias = "dev", visible_short_aliases = ['d', 'b'])]
+    build: Option<Vec<PackageReq>>,
+
+    /// Install the package as a test dependency.
+    #[arg(short, long)]
+    test: Option<Vec<PackageReq>>,
+}
+
+pub async fn add(data: Add, config: Config) -> Result<()> {
+    let mut project = Project::current()?.ok_or_eyre("No project found")?;
+
+    let pin = PinnedState::from(data.pin);
+    let lua_version = LuaVersion::from(&config)?;
+    let tree = project.tree(lua_version)?;
+
+    project
+        .add(rocks_lib::project::DependencyType::Regular(
+            data.package_req.clone(),
+        ))
+        .await?;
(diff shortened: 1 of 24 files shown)


vhyrro:
Things to flesh out:
- [ ] When adding a new dependency without a version constraint, do we want to set it as `dep = "*"` (like the code currently does), or bind it to the latest version by default? (gut says bind to latest version, which is the approach that Cargo takes)
- [ ] Fix a small annoyance where `toml-edit` creates inline tables instead of proper TOML tables

mrcjkb:
>  When adding a new dependency without a version constraint, do we want to set it as dep = "*" (like the code currently does), or bind it to the latest version by default? (gut says bind to latest version, which is the approach that Cargo takes)

Let's add the latest version.

vhyrro:
TODO:
- Prevent empty tables from being created if they have no content (there's a flag for this in toml_edit)
- Ensure that the rocks.toml is never overwritten during an error case.

vhyrro:
Summary: `rocks add` and refactors existing code to enforce the handling of unsupplied package version requirements.

vhyrro:
Soo how on earth did this magically get conflicted? :p

rocks-lib/src/package/version.rs
```diff
@@ -226,9 +226,15 @@ pub enum PackageVersionReq {
 }
 
 impl PackageVersionReq {
+    /// Returns a `PackageVersionReq` that matches any version.
+    pub fn any() -> Self {
+        PackageVersionReq::SemVer(VersionReq::default())
```
mrcjkb, line 231: thought: when we add support for dev manifests, we might want to add a new `PackageVersionReq::Any` variant.
This is fine for now.

rocks-lib/src/project/mod.rs
```diff
@@ -121,6 +139,221 @@ impl Project {
...
+        let mut rocks =
+            toml_edit::DocumentMut::from_str(&tokio::fs::read_to_string(self.rocks_path()).await?)?;
+
+        if !rocks.contains_table("dependencies") {
+            let mut table = toml_edit::table().into_table().unwrap();
+            table.set_implicit(true);
+
+            rocks["dependencies"] = toml_edit::Item::Table(table);
+        }
+        if !rocks.contains_table("build_dependencies") {
```
mrcjkb, line 157: question: will this add an empty table if we don't specify any build dependencies to add?

rocks-bin/src/add.rs
```diff
@@ -0,0 +1,96 @@
...
+            .add(
+                rocks_lib::project::DependencyType::Test(
+                    test_packages.iter().map(|(_, req)| req.clone()).collect(),
+                ),
+                &db,
+            )
+            .await?;
+    }
+
+    operations::Install::new(&config)
```
mrcjkb, line 86: issue: if `rocks add` also installs the packages, we're also going to have to add them to the project lockfile (see `rocks build`).

mrcjkb (changes requested):

rocks-lib/src/project/mod.rs
vhyrro: Nope, the tables are marked as `implicit` to prevent this :)

rocks-lib/src/package/version.rs
vhyrro: That sounds reasonable. Initially I tried implementing a `Latest` variant, but that requires some impossible refactors (as you need to have access to a packagedb at all times to query the latest version if need be). `Any` sounds easier to implement.

rocks-bin/src/add.rs
vhyrro: note: this uses a different implementation to the way `Build` does it. I think the `build` code could be simplified to match the more convenient API?

vhyrro:
Tests are failing because Neovim changed its binaries, unrelated to the PR :)

mrcjkb (approved):

PR #369, opened 2025-01-31
feat: support dev manifests

mrcjkb:


rocks-lib/src/package/version.rs
```diff
@@ -234,7 +237,16 @@ impl PackageVersionReq {
...
+            (PackageVersionReq::SemVer(version_req), PackageVersion::DevVer(..)) => {
+                let large_version = Version {
+                    major: u64::MAX,
+                    minor: u64::MAX,
+                    patch: u64::MAX,
+                    pre: Prerelease::EMPTY,
+                    build: BuildMetadata::EMPTY,
+                };
+                version_req.matches(&large_version)
+            }
```
vhyrro, line 249: suggestion: it's quite hard figuring out what this exactly does, maybe a comment could be useful?

mrcjkb, line 249: I will remove this again in the follow-up PR that implements `PackageVersionReq::Any`,
because I think this (how luarocks does it) is an anti-feature.
We can have `<package>` without a version constraint parse as `Any`, and it will still be ergonomic to install dev packages with `--dev`.

vhyrro (approved):

PR #373, opened 2025-01-31
chore: Add `PackageVersionReq::Any` variant

### LINKED ISSUE #364: [Refactor] Add `PackageVersionReq::Any` variant
thought: when we add support for dev manifests, we might want to add a new `PackageVersionReq::Any` variant.
This is fine for now.

_Originally posted by @mrcjkb in https://github.com/nvim-neorocks/rocks/pull/362#discussion_r1933323550_

mrcjkb:
Stacked on #369.
Closes #364.

This also changes the behaviour when installing with a `--dev` flag.
`foo >= <semver>` without an upper constraint no longer matches `<dev>`.
Only `Any` and `<dev>` can match `<dev>`, which I think will result in less surprises for users.
For example, `rocks --dev update` won't update semver packages to dev versions.

### DIFF
--- a/rocks-lib/src/package/version.rs
+++ b/rocks-lib/src/package/version.rs
@@ -3,7 +3,7 @@ use std::{cmp::Ordering, fmt::Display, str::FromStr};
 use html_escape::decode_html_entities;
 use itertools::Itertools;
 use mlua::FromLua;
-use semver::{BuildMetadata, Comparator, Error, Op, Prerelease, Version, VersionReq};
+use semver::{Comparator, Error, Op, Version, VersionReq};
 use serde::{de, Deserialize, Deserializer, Serialize};
 use thiserror::Error;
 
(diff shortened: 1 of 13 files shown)


rocks-lib/src/package/version.rs
```diff
@@ -243,29 +244,26 @@ impl PackageVersionReq {
...
-                    build: BuildMetadata::EMPTY,
-                };
-                version_req.matches(&large_version)
-            }
+            (PackageVersionReq::SemVer(..), PackageVersion::DevVer(..)) => false,
             (PackageVersionReq::Dev(..), PackageVersion::SemVer(..)) => false,
             (PackageVersionReq::Dev(name_req), PackageVersion::DevVer(devver)) => {
                 name_req.ends_with(&devver.modrev)
             }
+            (PackageVersionReq::Any, _) => true,
```
vhyrro, line 252: question: should `(_, Any)` also be handled here?

mrcjkb: there's no `PackageVersion::Any`.

vhyrro (approved):

PR #378, opened 2025-02-02
fix: make `rocks lua` also provide project paths

### LINKED ISSUE #377: [Issue] `rocks lua` should provide paths to the current project too
Since the latest changes in #375 we should ensure that other subcommands properly interact with the project state now that the project tree isn't automatically selected by the config.

Commands like `rocks list` can be left alone, since `rocks list` should always operate on the user-wide tree, but commands like `rocks lua` should definitely also consider the current project.

Repro steps:
- Create new project
- `rocks add argparse@0.7`
- `rocks lua`
- `require("argparse").version`

vhyrro:
Fixes #377

### DIFF
--- a/rocks-bin/src/run_lua.rs
+++ b/rocks-bin/src/run_lua.rs
@@ -55,7 +55,17 @@ pub async fn run_lua(run_lua: RunLua, config: Config) -> Result<()> {
         }
     }
     let tree = config.tree(lua_version)?;
-    let paths = Paths::from_tree(tree)?;
+
+    let paths = if let Some(project) = project {
+        let mut paths = Paths::new(tree)?;
+
+        paths.prepend(&Paths::new(project.tree(&config)?)?);
+
+        paths
+    } else {
+        Paths::new(tree)?
+    };
+
     let status = match Command::new(&lua_cmd)
         .args(run_lua.args.unwrap_or_default())
         .env("PATH", paths.path_prepended().joined())
--- a/rocks-bin/src/path.rs
+++ b/rocks-bin/src/path.rs
@@ -74,7 +74,7 @@ impl Default for Shell {
 
 pub async fn path(path_data: Path, config: Config) -> Result<()> {
     let tree = config.tree(LuaVersion::from(&config)?)?;
-    let paths = Paths::from_tree(tree)?;
+    let paths = Paths::new(tree)?;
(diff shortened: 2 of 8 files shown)


rocks-bin/src/run_lua.rs
```diff
@@ -55,7 +55,17 @@ pub async fn run_lua(run_lua: RunLua, config: Config) -> Result<()> {
...
     }
     let tree = config.tree(lua_version)?;
-    let paths = Paths::from_tree(tree)?;
+
+    let paths = if let Some(project) = project {
+        let mut paths = Paths::new(tree)?;
+
+        paths.prepend(&Paths::new(project.tree(&config)?)?);
+
+        paths
```
mrcjkb, line 64: question: If I understand this correctly, this makes the user-installed packages available on the paths.
Is that the behaviour we want? I think we should only include the local project tree.

vhyrro: I do think we want to include all packages. The ones in the project have higher precedence by being earlier in the path.

I'd expect `rocks lua` to be a wrapper around Lua with all the relevant packages available to the REPL. The ones I installed in the user-wide tree also count as relevant, I'd say :p

mrcjkb: scenario:

- I have package `foo` installed in my user tree
- My project depends on `foo`, but it's not listed in the rocks.toml dependencies
- I'm too dumb to write a test suite and I only use `rocks lua` to test my package manually
- My manual tests succeed
- I publish my package

I think we should default to not including all packages, but perhaps could add an `--impure` flag that changes the behaviour.
But I really can't think of a scenario where you'd actually want that.

vhyrro: A `--impure` flag sounds nice, but shouldn't the user be running their code through the (hopefully) upcoming `rocks run` anyway, which will be a pure environment?

Another option would be to differentiate script-running mode (pure by default) and REPL mode (impure by default), which I think sounds pretty sensible without being too "magical". Wdyt?

mrcjkb: Sure, that sounds fine to me.

mrcjkb (approved):

rocks-bin/src/run_lua.rs
vhyrro: Let me make a tracking issue for that.

PR #385, opened 2025-02-02
feat: `rocks sync` command

### LINKED ISSUE #273: [Feature] `rocks sync` command
- Use the `rocks.lock` as the source of truth to sync rocks in a given tree.
- Fail if there is no lockfile.
- Fail if there are rockspec or source integrity mismatches
  - Should have an option to only warn on failed integrity checks (to handle force-uploaded scm rockspecs).

mrcjkb:
Closes #273.

Bonus:

- ~Fixes the `HasIntegrity` implementation for `PathBuf`.~ -> revamped again in #401.
- Removes the `.git` directory when checking out a repo with libgit2, to make sure the hash is reproducible.
  ~I don't know if this is compatible with nix, but I don't know a better way to solve this right now.~
- Rewires `rocks build` to use `Sync` if a lockfile is present, otherwise `Install`, passing in the project reference.


mrcjkb:
issue: `rocks build` needs to sync the dependencies in rocks.toml before syncing the lockfile, in case dependencies were added or removed manually.

vhyrro (commented):
looks great. I like that we got rid of the LockfileUpdate :+1: just that one typo which i won't be able to live with :)

vhyrro (approved):

PR #386, opened 2025-02-02
fix: recursively unpack nested archives

mrcjkb:
Scenario:

An older rockspec that does not specify a `source.dir` and for which the `source.url` no longer exists:

```lua
source = {
    url = "http://www.kyne.com.au/~mark/software/download/lua-cjson-2.1.0.zip",
}
```

The luarocks `src.rock` looks like this:

- `lua-cjson-2.1.0-1.src.rock`
  - `lua-cjson-2.1.0-1.rockspec`
  -  `lua-cjson-2.1.0.zip` -> contains the directory: `lua-cjson-2.1.0` (not specified in the `source.dir`).

To account for this, we need to recursively unpack the nested source archive and infer the source directory.

### DIFF
--- a/rocks-lib/src/build/mod.rs
+++ b/rocks-lib/src/build/mod.rs
@@ -21,6 +21,7 @@ use command::CommandError;
 use external_dependency::{ExternalDependencyError, ExternalDependencyInfo};
 
 use indicatif::style::TemplateError;
+use itertools::Itertools;
 use luarocks::LuarocksBuildError;
 use make::MakeError;
 use patch::{Patch, PatchError};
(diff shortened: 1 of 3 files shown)


vhyrro:
Code looks great, but this definitely needs a rigorous test harness. I remember implementing this functionality at one point and breaking half the other luarocks packages which *didn't* want recursive unpacking (I think it was either kong or some openresty package that didn't like this) :/

How hard would it be for us to make a harness that tests rocks compatibility against a large sample of luarocks.org rockspecs? Preferably without bombarding them with get requests every day :sweat_smile:

mrcjkb:
> I remember implementing this functionality at one point and breaking half the other luarocks packages which didn't want recursive unpacking (I think it was either kong or some openresty package that didn't like this) :/

I just realised your check was already in place, for `tar.gz` archives only though.
I have modified my PR to use the same flag prevent the recursive unpack call if `source.file` is set.
Note that the flag is always `false` when unpacking a `.src.rock`.

PR #397, opened 2025-02-08
fix: make libgit2 checkout hash compatible with nix

### LINKED ISSUE #396: [Issue] libgit2 checkout hash is not compatible with nix
When checking out a repo with libgit2, it modifies the `.git` directory, making the source hash we compute random and incompatible with nix.
As a workaround (implemented in #385), we delete the `.git` directory before computing the hash.
This makes it reproducible, but we will not be able to use the hash to create a fixed-output-derivation.

To solve this, we need to:

- Figure out how to prevent libgit2 from modifying the `.git` directory.
- No longer delete the unmodified `.git` directory before computing the source hash.

mrcjkb:
Stacked on #385.

Fixes #396.


mrcjkb:
> It probably won't be possible to create fixed-output derivations using lockfiles that contain packages with a git source

Perhaps for this reason, it might actually be better to prioritise .src.rock source archives :thinking: 
What do you think @vhyrro?

mrcjkb:
I've worked out to compute nix-compatible source hashes for paths in #400.
Once I work out how to make sure the source hash of a package we fetch with libgit2 is nix-compatible, I will stack another PR on top of that.
I will probably end up reimplementing the cleanup logic in [`nix-prefetch-git`](https://github.com/NixOS/nixpkgs/blob/5156904c711d340e303ad800b94369b3a7085e42/pkgs/build-support/fetchgit/nix-prefetch-git#L255).

mrcjkb:
Actually, [`nix-prefetch-git` just deletes the `.git` directory by default](https://github.com/NixOS/nixpkgs/blob/9aed71348bff9c6e142cc3f64206a128388494b9/pkgs/build-support/fetchgit/nix-prefetch-git#L331).

PR #401, opened 2025-02-10
fix(hash): ensure path hashing compatibility with nix-hash

mrcjkb:
Supercedes #400, which was closed because of a `jj abandon` of the PR it was stacked on.

Stacked on #385.

### DIFF
--- a/rocks-lib/src/luarocks/install_binary_rock.rs
+++ b/rocks-lib/src/luarocks/install_binary_rock.rs
@@ -255,6 +255,7 @@ mod test {
         .unwrap();
         let installed_rock_layout = tree.rock_layout(&local_package);
         let orig_install_tree_integrity = installed_rock_layout.rock_path.hash().unwrap();
+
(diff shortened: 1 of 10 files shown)


rocks-lib/src/luarocks/install_binary_rock.rs
```diff
@@ -230,7 +230,7 @@ mod test {
             file_name: packed_rock_file_name.clone(),
         };
         let rockspec = unpack_rockspec(&rock).await.unwrap();
-        let install_root = assert_fs::TempDir::new().unwrap();
+        let install_root = PathBuf::from("/tmp/old/"); //assert_fs::TempDir::new().unwrap();
```
vhyrro, line 233: question: is this a leftover?

vhyrro (dismissed):

rocks-lib/src/luarocks/install_binary_rock.rs
mrcjkb: Yes, good catch

PR #405, opened 2025-02-12
refactor(lockfile): extract `LocalPackageLock` type

mrcjkb:
This is in preparation for #395.

PR #407, opened 2025-02-12
feat!: add build/test dependencies to rocks.lock

### LINKED ISSUE #395: [Feature] add build/test dependencies to rocks.lock
The project lockfile should probably include build/test dependencies.
This makes me think we might want to introduce a type like:

```rust
struct RocksLock<P: LockfilePermissions> {
  tree: DepsLock,
  test_tree: Option<DepsLock>,
  build_tree: Option<DepsLock>,
}
```

where `DepsLock` is equivalent to the current `Lockfile` struct, but without the `filepath` fields.

mrcjkb:
Stacked on #405 

Closes #395.

### DIFF
--- a/rocks-bin/src/add.rs
+++ b/rocks-bin/src/add.rs
@@ -1,15 +1,14 @@
-use eyre::{OptionExt, Result};
+use eyre::{Context, OptionExt, Result};
 use rocks_lib::{
     config::Config,
     lockfile::PinnedState,
-    operations,
+    luarocks::luarocks_installation::LuaRocksInstallation,
+    operations::Sync,
     package::PackageReq,
     progress::{MultiProgress, Progress, ProgressBar},
     project::Project,
     remote_package_db::RemotePackageDB,
 };
-
-use crate::utils::install::apply_build_behaviour;
 
 #[derive(clap::Args)]
 pub struct Add {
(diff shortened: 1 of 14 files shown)


mrcjkb:
todo:

- [ ] Remove `project_lockfile` from `operations::install` and use `operations::sync` instead.
- [ ] Sync `build_dependencies` in `rocks build` (#406).
- [ ] Sync `test_dependencies` in `rocks test`.

mrcjkb:
No idea why the integration tests are failing on Ubuntu 🥲

vhyrro:
Not sure I understand the following:

> That would mean you have to declare a dependency both in the regular dependencies and in the test dependencies if you want to use them in tests

If they're in the test dependencies then that means you want to use them in tests, right? :p

mrcjkb:
> If they're in the test dependencies then that means you want to use them in tests, right? 

The question is, do you want to be able to `require` a regular dependency in tests?
Intuitively, I would say yes.

mrcjkb:
Okay now it looks like it might be luarocks.org having issues.

mrcjkb:
Ha! I figured it out.
It was two `#[tokio::test]`s installing busted into the same test tree. That explains the flakiness :hand_over_mouth:

vhyrro:
> > If they're in the test dependencies then that means you want to use them in tests, right?
> 
> The question is, do you want to be able to `require` a regular dependency in tests? Intuitively, I would say yes. And that's how I'm trying to implement it in this PR. But afaik, that's not how it works with luarocks.

Yep, if you're testing functionality, and the regular dependencies are part of the functionality, then you definitely want them included too :p

PR #409, opened 2025-02-13
feat(lockfile): add source_url info

### LINKED ISSUE #408: [feature] include source type in lockfile
To make sure we know to use `fetchGit` or `fetchUrl` when creating a fixed-output-derivation from our lockfile,
we should include information about the source type.

mrcjkb:
Stacked on #407 

Closes #408

### DIFF
--- a/rocks-lib/resources/test/rocks.lock
+++ b/rocks-lib/resources/test/rocks.lock
@@ -2,22 +2,6 @@
   "version": "1.0.0",
   "dependencies": {
     "rocks": {
-      "028a2b47c550e0e17bd2952c7444b6bf64687b3cd6b59573a066a21f545466b8": {
-        "name": "lua-cjson",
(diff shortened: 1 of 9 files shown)


vhyrro:
question: what do you say that we make the representation an [internally tagged enum](https://serde.rs/enum-representations.html#internally-tagged) instead (`{ "type": "url", "uri": "..." }`)? Should make it generally easier to parse for external programs without having to do `nil` checks :)

mrcjkb:
> question: what do you say that we make the representation an [internally tagged enum](https://serde.rs/enum-representations.html#internally-tagged) instead (`{ "type": "url", "uri": "..." }`)? Should make it generally easier to parse for external programs without having to do `nil` checks :)

Good idea.

vhyrro (approved):

PR #410, opened 2025-02-14
refactor!: rename project to `lux`

vhyrro:
I probably missed some segments. Hope the tests pass :crossed_fingers:


mrcjkb (changes requested):
We probably need to rename rocks.lock in our test resources, too.

vhyrro:
Looks like the tests are going to succeed!

mrcjkb (approved):

Issue #380, opened 2025-02-02
[Discussion] Make `rocks test`/`rocks run` local


vhyrro:
If executed without arguments, `rocks test` and `rocks run` should try to operate on the current project.
The design for this is simpler for `rocks test` than it is for `rocks run`, which needs further design considerations in the dedicated issue.

vhyrro:
(by local I mean without having to specify a `url` in the rockspec)

PR #411, opened 2025-02-15
refactor!: break rockspec apart into `LocalRockspec` and `RemoteRockspec`

vhyrro:
In preparation of #380

### DIFF
--- a/lux-lib/src/build/mod.rs
+++ b/lux-lib/src/build/mod.rs
@@ -1,6 +1,5 @@
 use crate::lockfile::RemotePackageSourceUrl;
 use crate::lua_rockspec::LuaVersionError;
-use crate::project::project_toml::LocalProjectTomlValidationError;
 use crate::rockspec::{LuaVersionCompatibility, Rockspec};
 use std::{io, path::Path, process::ExitStatus};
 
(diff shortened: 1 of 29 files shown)


lux-lib/src/build/mod.rs
```diff
@@ -46,7 +46,7 @@ pub mod variables;
 /// over how a package should be built.
 #[derive(Builder)]
 #[builder(start_fn = new, finish_fn(name = _build, vis = ""))]
-pub struct Build<'a, R: Rockspec + HasIntegrity> {
+pub struct Build<'a, R: RemoteRockspec + HasIntegrity> {
```
mrcjkb, line 49: question: wouldn't we want to be able to build both local and remote rockspecs?
I suppose this would require local rockspecs to be able to provide the path to the project.

mrcjkb (approved):

lux-lib/src/build/mod.rs
vhyrro: Anything that's a RemoteRockspec has to be a LocalRockspec already, so we'll maybe have to add a special `.local()` function to the builder to differentiate? That's for a different pull request for sure :)

vhyrro: oh I see what you mean, you want to be able to pass just a localrockspec. We might need a special refactor for this case indeed.

mrcjkb (approved):

vhyrro:
Once tests pass I'll merge with rebase so the commits are preserved :)

vhyrro:
slightly suspicious that only a single version of ubuntu with a single lua version failed in one spot?

PR #417, opened 2025-02-15
refactor!: allow building of local rockspecs

vhyrro:


vhyrro:
I'll merge this down into the other pull request I have because some changes here overwrite the ones made previously.

PR #421, opened 2025-02-16
fix!: multiple sync operations overwrite lockfile

### LINKED ISSUE #414: [Feature] Consistent behaviour on existing packages
We should establish the correct behaviour for the following cases:
- An entry for a rock is already part of the lux.toml (manually added) but not installed yet and someone runs `lux add <rock>`
- A rock is already installed and someone runs `lux add <rock>`

### LINKED ISSUE #415: [Bug] `busted` is reinstalled on every `lux test`

mrcjkb:
Fixes #415


lux-lib/src/lockfile/mod.rs
```diff
@@ -880,7 +899,19 @@ impl Lockfile<ReadOnly> {
     }
 
     /// Creates a temporary, writeable lockfile which can never flush.
-    pub fn into_temporary(self) -> Lockfile<ReadWrite> {
+    #[cfg(feature = "lua")]
+    pub(crate) fn into_temporary(self) -> Lockfile<ReadWrite> {
```
vhyrro, line 903: question(rationale): I see that we're doing an override to make `into_temporary` public only to the crate here, but why on `feature = "lua"` specifically? Can't quite grasp what difference it makes :sweat_smile:

vhyrro (approved):

lux-lib/src/lockfile/mod.rs
mrcjkb: We only need it to be `pub(crate)` for a `UserData` impl in another module.
Without the `lua` feature, this function is private to this module, because we should typically use `write_guard`.

PR #422, opened 2025-02-16
feat(lockfile): add git `ref`

### LINKED ISSUE #419: [Feature] Add git `ref` to lockfile

mrcjkb:
Stacked on #421 

Closes #419


mrcjkb:
Come to think of it, the `ref` in the lockfile should not be optional. We should write the current revision if it's not specified in the rockspec source and use that when using the lockfile as a `RemotePackageDB` (for another PR).

vhyrro (approved):

PR #423, opened 2025-02-16
feat!: rename binary to `lx`

mrcjkb:


vhyrro (approved):

teto:
for posterity, what's the rationale :) ? is `lux` already taken by a famous executable ?

Issue #426, opened 2025-02-16
[Feature] cargo-like `lx update`


mrcjkb:
Currently, `lx update` only updates packages in the install tree.

We probably want to make the behaviour more like `cargo update`, where we sync the changes back to the project lockfile and update version constraints in the `lux.toml` where applicable.

PR #428, opened 2025-02-17
feat(update): operate on lux.toml and lux.lock if in a project

mrcjkb:
Stacked on #425.

Addresses #426 (partially).

Also fixes `Sync` so that it keeps only packages that have the same constraints as the `Vec<PackageReq>` constraints we pass in from the lux.toml dependencies.
This ensures that if we update a version constraint in `lux.toml`, it is reflected in the lockfile.

### DIFF
--- a/lux-lib/src/operations/update.rs
+++ b/lux-lib/src/operations/update.rs
@@ -1,4 +1,4 @@
-use std::sync::Arc;
+use std::{io, sync::Arc};
 
 use bon::Builder;
 use itertools::Itertools;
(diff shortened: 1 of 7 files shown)


lux-lib/src/operations/update.rs
```diff
@@ -1,92 +1,199 @@
...
+
+    let dependencies = toml
+        .dependencies()
+        .current_platform()
+        .iter()
+        .filter(|package| !package.name().eq(&PackageName::new("lua".into())))
+        .cloned()
+        .collect_vec();
+    let dep_report = super::Sync::new(&tree, &mut project_lockfile, args.config)
+        .validate_integrity(false)
```
mrcjkb, line 106: tode:

- [ ] this should default to `true`

mrcjkb:
> I see some accidental commits of `.rocks` directories, might be worth looking into that ;)

Looks like `rocks-lib` snuck itself back into the repo

vhyrro (approved):
:fire:

vhyrro (approved):

PR #433, opened 2025-02-18
feat!(cli): remove `sync` command

mrcjkb:
Stacked on #432.

With the ability to run `lx --tree <project_tree_override> build` within any directory containing a `lux.toml`, there is no need for `lx sync` anymore.


vhyrro (changes requested):
might also want to remove `lux-bin/src/sync.rs` itself? :)

mrcjkb:
> might also want to remove `lux-bin/src/sync.rs` itself? :)

Must have slipped back in during a rebase.

vhyrro (approved):

PR #436, opened 2025-02-21
feat: Lua API

vhyrro:
Heavily WIP, let's get this thing finally to completion.


vhyrro:
I was hoping to get this done today, but I've had to do a rewrite of the loader to account for more edge cases. An MVP is definitely close now!

What's interesting is that the loader can be as intelligent as we need to be, including being able to search through test and build dependencies by detecting the mode lux is running in. Not to mention the various optimizations we could perform. Sky's the limit :)

lux-lib/src/lockfile/mod.rs
```diff
@@ -221,14 +247,34 @@ pub(crate) enum RemotePackageSourceUrl {
...
+            Ok(this.spec.dependencies.clone())
+        });
+        methods.add_method("constraint", |_, this, _: ()| {
+            Ok(this.spec.constraint.clone())
+        });
+        methods.add_method("hashes", |_, this, _: ()| Ok(this.hashes.clone()));
+        methods.add_method("to_package", |_, this, _: ()| Ok(this.to_package()));
+        methods.add_method("into_package_req", |_, this, _: ()| {
+            Ok(this.clone().into_package_req())
+        });
```
mrcjkb, line 274: suggestion: since Lua doesn't have a concept of ownership, perhaps this should be

```suggestion
        methods.add_method("to_package_req", |_, this, _: ()| {
            Ok(this.clone().into_package_req())
        });
```

lux-lib/src/lockfile/mod.rs
```diff
@@ -500,7 +556,7 @@ impl LockfilePermissions for ReadOnly {}
 impl LockfilePermissions for ReadWrite {}
 
 #[derive(Clone, Debug, Serialize, Deserialize, Default)]
-pub(crate) struct LocalPackageLock {
+pub struct LocalPackageLock {
```
mrcjkb, line 559: suggestion: Instead of exposing this type, can we add delegate methods?

lux-lib/src/path.rs
```diff
@@ -2,23 +2,39 @@ use itertools::Itertools;
...
 
-use crate::{build::utils::lua_lib_extension, tree::Tree};
+use crate::{
+    build::utils::lua_lib_extension,
+    config::{Config, LuaVersion},
+    tree::Tree,
+};
 
 const LUA_PATH_SEPARATOR: &str = ";";
+const LUA_INIT: &str = "require('lux').loader()";
```
mrcjkb, line 12: question: :thinking: [should we name this module something else](https://github.com/IgorMael/Lux/blob/master/src/lux.lua)?
`lx` isn't taken.

lux-lib/src/project/project_toml.rs
```diff
@@ -105,6 +107,23 @@ pub struct PartialProjectToml {
...
+impl UserData for PartialProjectToml {
+    fn add_methods<M: mlua::UserDataMethods<Self>>(methods: &mut M) {
+        methods.add_method("package", |_, this, _: ()| Ok(this.package().clone()));
+        methods.add_method("version", |_, this, _: ()| Ok(this.version().clone()));
+        methods.add_method("into_local", |_, this, _: ()| {
+            this.into_local().into_lua_err()
+        });
+        methods.add_method("into_remote", |_, this, _: ()| {
+            this.into_remote().into_lua_err()
+        });
```
mrcjkb, line 119: ditto: `into_`

```suggestion
        methods.add_method("to_local", |_, this, _: ()| {
            this.into_local().into_lua_err()
        });
        methods.add_method("to_remote", |_, this, _: ()| {
            this.into_remote().into_lua_err()
        });
```

lux-lua/loader-notes.txt
```diff
@@ -0,0 +1,10 @@
+Currently, we assume a single version of the loader is used everywhere.
+The `lx path` command should not only generate the code to load the loader, but should also compare against `_VERSION` to see if it even *can*.
+
+Currently, support for searching through dev and build dependencies is unsupported, and we should have some way of detecting when we're in those modes
+(probably through environment variables that lux sets?)
+
+Current, the loader does not isolate dependencies from potentially impure places, as it delegates anything it can't find to further searchers.
+This is a tricky topic - we probably don't want to overwrite the existing searchers as there may be some important ones (?), but we also want to ensure
+that dependencies that shouldn't be found aren't found. I am considering hijacking the `package.searchers` entirely and just making the lux
+loader do all the work.
```
mrcjkb, line 10: note: With the luarocks loader, some rocks.nvim users have expressed the concern about not being able to set `vim.loader.enable()`. This would be the case if the lux loader hijacks `package.searchers`.
Personally, I think it's fine to fall back to other searchers. If the lux loader can ensure the correct dependency is loaded, it's done its job.

lux-lua/src/loader.rs
```diff
@@ -0,0 +1,190 @@
...
+}
+
+fn load_file(lua: &Lua, module: String, path: PathBuf) -> mlua::Result<mlua::Function> {
+    lua.create_function(move |lua, ()| {
+        // We need to handle 3 possible cases:
+        // - `src/?.lua`
+        // - `src/?/init.lua`
+        // - `src/?.so`
+
+        let module_path = module.replace('.', "/");
```
mrcjkb, line 41: suggestion:

```suggestion
        let module_path = module.replace('.', std::path::MAIN_SEPARATOR_STR);
```

lux-lua/src/loader.rs
```diff
@@ -0,0 +1,190 @@
...
+    let current_file = current_file.absolutize().into_lua_err()?;
+
+    match lets_find_up::find_up_with(
+        "lux.lock",
+        lets_find_up::FindUpOptions {
+            cwd: current_file.parent().unwrap(),
+            kind: lets_find_up::FindUpKind::File,
+        },
+    )
+    .into_lua_err()?
```
mrcjkb, line 104: question: could searching for the lockfile on every `require` be expensive?
Perhaps we should cache the results for `current_file`. I suppose since we're in the Lua space, we would have to use a Lua global.

lux-lib/src/lockfile/mod.rs
```diff
@@ -1336,7 +1437,7 @@ mod tests {
 
     fn get_test_lockfile() -> Lockfile<ReadOnly> {
         let sample_tree = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
-            .join("resources/test/sample-tree/5.1/lock.json");
+            .join("resources/test/sample-tree/5.1/lux.lock");
```
mrcjkb, line 1440: todo:

- [ ] update/rename snapshot files (I think this is why the tests are failing).

lux-lib/src/lockfile/mod.rs
vhyrro: Sure, I'll get to that near the end of this PR :+1:

lux-lib/src/path.rs
vhyrro: I think lux is fine. I'm all for preventing conflicts, but none of the existing rocks are large enough for us to realistically have to worry imo

lux-lua/src/loader.rs
vhyrro: This implementation was a proof of concept. I'll definitely be spending time optimizing this further with a caching mechanism and with some heuristics to prevent having to traverse the filesystem itself.

nix/overlay.nix
```diff
@@ -51,35 +54,96 @@
...
+    nativeBuildInputs = with final; [
+      cargo-hakari
+    ];
+  };
+
   lux-nextest = craneLib.cargoNextest (commonArgs
     // {
-      src = self;
+      inherit (luxCliCargo) pname version;
+      src = cleanCargoSrc;
```
mrcjkb, line 146: issue: `cleanCargoSrc` removes the test resources.

```suggestion
      src = self;
```

lux-lua/src/loader.rs
vhyrro: New implementation completely omits any filesystem reads for lookup and simply loops over the path in-memory to try and find the correct path :)

lux-lib/src/lockfile/mod.rs
vhyrro: With the refactor we don't even need to delegate anything anymore, so I'll just revert these parts of the changes.

vhyrro:
I'm personally in favour of merging this in its current state. The loader is finished and functional, the workspace hack crate is complete and the most important components of the Lua API are in place. I'm fond of supplementing the API as we need things during our rocks.nvim rewrite. What do you think @mrcjkb? Or should I finish off all the work in this PR? Either is fine.

mrcjkb (approved):
> > I'm personally in favour of merging this in its current state. The loader is finished and functional, the workspace hack crate is complete and the most important components of the Lua API are in place. I'm fond of supplementing the API as we need things during our rocks.nvim rewrite. What do you think @mrcjkb? Or should I finish off all the work in this PR? Either is fine.

Agreed.

PR #453, opened 2025-03-03
feat!: optional packages

mrcjkb:
Stacked on #436 (to avoid conflicting with it).

This PR introduces the following new concepts:

- `lockfile::OptState`, which can be `Required` and `Optional`.
  For now, this doesn't have much meaning to lux, but it influences the `LocalPackageId` hash
  and will be useful later.
- `PackageInstallSpec`, which replaces the `(BuildBehaviour, PackageReq)` tuple
  and includes the `PinnedState` and `OptState`.

The reason for this is so that when building a local project, we can support stuff like this:

```toml
[dependencies.foo]
version = "1.0.0"
opt = true
pin = true
```

This will require a follow-up PR that introduces a `Dependency` type with an `into_package_req` method.
The `*ProjectToml` types will then have lists of `Vec<Dependency>` instead of `Vec<PackageReq>`.

The reason for introducing the `OptState` is so that we can later use it to manipulate the `RockLayout` to support Neovim `packpath` structures.

We could potentially also exclude `opt` dependencies from the `lx path` result, unless they are a dependency of another non-opt dependency (yagni?).

### DIFF
--- a/lux-cli/src/build.rs
+++ b/lux-cli/src/build.rs
@@ -7,9 +7,9 @@ use eyre::Result;
 use lux_lib::{
     build::{self, BuildBehaviour},
     config::Config,
-    lockfile::PinnedState,
+    lockfile::{OptState, PinnedState},
     luarocks::luarocks_installation::LuaRocksInstallation,
-    operations::{Install, Sync},
+    operations::{Install, PackageInstallSpec, Sync},
     package::PackageName,
     progress::MultiProgress,
     project::Project,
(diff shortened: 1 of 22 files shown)


vhyrro (approved):

vhyrro:
I'd say YAGNI for the path related operations for now. `opt` could be a really interesting way for us to implement something akin to features from cargo, all while making it general enough to be supported in neovim.

PR #456, opened 2025-03-06
feat!(toml): `opt` and `pin` fields

mrcjkb:
Stacked on #454.

This adds the ability to set the `opt` and `pin` fields for a Lua dependency via the `lux.toml` entries.


vhyrro (approved):
i was certain i approved this one

PR #457, opened 2025-03-07
feat!: allow overriding `etc` tree

### LINKED ISSUE #463: [bug] lockfile entrypoints
https://github.com/nvim-neorocks/lux/blob/2da6e424a0bee18df9b4faa03f9e43642a455ba1/lux-lib/src/lockfile/mod.rs#L785

If I install a package and then install another package with that package and the same constraint as a dependency, the current logic for determining the entrypoints will exclude that package.

mrcjkb:
This is a draft for allowing us to configure the install tree so that it can be made compatible with Neovim's `packpath` without any symlink hacks on the rocks.nvim side.

I'd like to make this generic, so that we can potentially serve other use cases in the future.

Also fixes #463.

### DIFF
--- a/lux-lib/src/build/mod.rs
+++ b/lux-lib/src/build/mod.rs
@@ -72,6 +72,10 @@ pub struct Build<'a, R: Rockspec + HasIntegrity> {
     #[builder(default)]
     behaviour: BuildBehaviour,
 
+    /// Whether to install `etc` and `doc` directories.
+    #[builder(default = true)]
+    install_etc: bool,
+
     // TODO(vhyrro): Remove this and enforce that this is provided at a type level.
     source: Option<RemotePackageSource>,
 }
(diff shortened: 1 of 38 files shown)


mrcjkb:
todo:

- [ ] only entrypoints should be installed into the `etc_root`.

lux-lib/src/build/mod.rs
```diff
@@ -55,6 +55,8 @@ pub struct Build<'a, R: Rockspec + HasIntegrity> {
     #[builder(start_fn)]
     tree: &'a Tree,
     #[builder(start_fn)]
+    is_entrypoint: bool,
+    #[builder(start_fn)]
```
vhyrro, line 59: suggestion(non-blocking): we probably want to make this an enum

vhyrro (dismissed):

lux-lib/src/build/mod.rs
mrcjkb: Yeah, makes sense.
I've added

```rust
struct EntryType {
  Entrypoint,
  DependencyOnly,
}
```

to the `tree` module.

PR #509, opened 2025-04-01
feat: enable vim mode for `lx new` selections

vhyrro:
This activates the `hjkl` keys and makes the experience easier when selecting from a dropdown in the terminal.


mrcjkb:
question: does this disable arrows for selection (or whatever was the default)? If so, we might want to make vim mode configurable, in case non vim users don't know how to navigate it.

mrcjkb (approved):

PR #512, opened 2025-04-02
fix: use compilation target to get platform identifier

mrcjkb:
We've been using the target that lux was built with to get the target platform identifier.
Instead, we need to use the target for the compiler toolchain we're using.


mrcjkb:
- [ ] Need some test cases for platform overrides in a rockspec.

vhyrro (approved):

PR #519, opened 2025-04-06
ci: move to `release-plz`

vhyrro:
`release-plz` has much more features and should work with workspaces out of the box.


vhyrro:
I hope it'll easily allow us to override the version to release with (e.g. `0.2.0` instead of `1.0.0`). I'll go scour the docs for that.

PR #520, opened 2025-04-06
fix: don't publish `lux-lua`

vhyrro:
that crate should only be built manually through xtask


vhyrro:
merging because it's trivial (already ran `nix flake check` beforehand locally)

PR #533, opened 2025-04-07
fix(upload): fix upload searching for old `project.rockspec` file

vhyrro:
This fixes `lx upload` failing in all cases (how was this not caught earlier? at least one test should have caught this)


vhyrro:
due to this being a pretty critical and simple fix I'll merge without review once the checks pass

PR #536, opened 2025-04-07
chore: add logo

mrcjkb:


vhyrro:
looking great - although I think we should retain the original structure (title above and catchphrase below). i took that from a template which was supposed to be a set of "best readme standards". also, max-width doesn't seem to be respected by github at all, we might need to just set that to `width` :p

PR #545, opened 2025-04-07
fix!: incompatible generated rockspec dependencies

### LINKED ISSUE #544: [Bug] version requirements in generated rockspecs not compatible with luarocks
- `lx new .`
- `lx add argparss`
- `lx build`
- Look at the `package.rockspec` in the built package's root

Expected: `argparse == 0.7.1` in dependencies
Actual: `argparse ^0.7.1` in dependencies

mrcjkb:
This is a breaking change because it changes the hashes in the lockfile.

Note: releaze-plz won't detect these kinds of breaking changes.

Fixes #544

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2177,6 +2177,7 @@ dependencies = [
  "open",
  "path-absolutize",
  "serde_json",
+ "serial_test",
  "spdx",
  "spinners",
  "strum 0.27.0",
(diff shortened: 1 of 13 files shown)


mrcjkb:
I've also reduced our sample lux.lock (removed neorg 8.0) because it is a pain to update.

vhyrro:
Just to be clear: all the hashes have changed because the constraint is now no longer `^xyz` but `=xyz`, correct?

mrcjkb:
> Just to be clear: all the hashes have changed because the constraint is now no longer `^xyz` but `=xyz`, correct?

Yes

vhyrro (approved):

PR #557, opened 2025-04-08
fix(build): properly handle legacy rockspecs

### LINKED ISSUE #540: [lx run]: project files not added to `LUA_PATH`
If I follow the guide for setting up a sample project, and then try and run it, even with the exact same file structure and code inside the files, running `lx run` doesn't seem to be able to find the has_hello.lua file. Even though it is in the src file with main.lua.

mrcjkb:
Fixes #540 (again)


vhyrro (dismissed):
Code looks good - we ought to make a large test harness with a bunch of rockspecs soon to check our compatibility percentage.

Issue #532, opened 2025-04-07
`PermissionDenied` error in Termux Android


xronin01:
![Image](https://github.com/user-attachments/assets/f63758d0-742e-42c9-9055-da9d5dcaa5aa)

```
thread 'main' panicked at /data/data/com.termux/files/home/.local/share/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/lux-cli-0.2.1/src/bin/lx.rs:43:41:
called `Result::unwrap()` on an `Err` value: Project(Io(Os { code: 13, kind: PermissionDenied, message: "Permission denied" }))
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
```
thread 'main' panicked at /data/data/com.termux/files/home/.local/share/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/lux-cli-0.2.1/src/bin/lx.rs:43:41:
called `Result::unwrap()` on an `Err` value: Project(Io(Os { code: 13, kind: PermissionDenied, message: "Permission denied" }))
stack backtrace:
   0:       0x64caf3c558 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::hf3165a890b3f99a7
   1:       0x64caf7b8d4 - core::fmt::write::h508e32b073fc24c7
   2:       0x64caf3a2b4 - std::io::Write::write_fmt::hec73da2da261f297
   3:       0x64caf3c458 - std::sys::backtrace::BacktraceLock::print::hb683b8d74ca6376a
   4:       0x64caf5e3bc - std::panicking::default_hook::{{closure}}::h8dd0451103712394
   5:       0x64caf5e21c - std::panicking::default_hook::hedb963aceca4abf9
   6:       0x64caf5edd0 - std::panicking::rust_panic_with_hook::hf43d358058db0558
   7:       0x64caf3ce94 - std::panicking::begin_panic_handler::{{closure}}::hc765cd251c62ee61
   8:       0x64caf3ca14 - std::sys::backtrace::__rust_end_short_backtrace::h28b56892d3980e5c
   9:       0x64caf5e864 - rust_begin_unwind
  10:       0x64caf7a114 - core::panicking::panic_fmt::he5d7e6884ef83b14
  11:       0x64caf840b0 - core::result::unwrap_failed::hbc28b18647150a07
  12:       0x64ca37dc38 - lx::main::{{closure}}::h14fb5ab43aab93ff
[...]

PR #567, opened 2025-04-10
fix: stop searching to lux.toml on filesystem error

mrcjkb:
See #532

Hopefully this fixes it? We'll have to see.


vhyrro (approved):
this looks like it should indeed do the trick. I wonder if assert_fs would have some utilities to test this

mrcjkb:
> this looks like it should indeed do the trick. I wonder if assert_fs would have some utilities to test this

I can't find anything :(
This is where it would be nice to have a Haskell effects system.

PR #592, opened 2025-04-14
perf: enable LTO + set `codegen-units = 1` in release profile

### LINKED ISSUE #531: Enable Link-Time Optimization (LTO) and codegen-units = 1
Hi!

I noticed that in the `Cargo.toml` file Link-Time Optimization (LTO) for the project is not enabled. I suggest switching it on since it will reduce the binary size (always a good thing to have) and will likely improve the application's performance a bit. If you want to read more about LTO and its possible modes, I recommend starting from [this](https://doc.rust-lang.org/rustc/codegen-options/index.html#lto) Rustc documentation.

I think you can enable LTO only for the Release builds so as not to sacrifice the developers' experience while working on the project, since LTO consumes an additional amount of time to finish the compilation routine. In this case, we can create a dedicated `[profile.optimized-dev]` profile where LTO will be disabled (so developers experience will not be affected). If we enable it on the Cargo profile level for the Release profile, users, who install the application with `cargo install`, will get the LTO-optimized version of the app "automatically". E.g., check `cargo-outdated` Release [profile](https://github.com/kbknapp/cargo-outdated/blob/master/Cargo.toml#L48). You also could be interested in other optimization options like `codegen-units = 1` - it also brings improvements over the current defaults.

Basically, it can be enabled with the following lines to the root Cargo.toml file:

```
[profile.release]
codegen-units = 1
lto = true
```

I have made quick tests (AMD Ryzen 5900x, Fedora 41, Rust 1.86, the latest version of the project at the moment, `cargo build --release` command) - here are the results for `lx`:

* Release: 29 Mib, clean build time: 1m 03s
* Release + `codegen-units = 1` + Fat LTO: 20 Mib, clean build time: 2m 26s

Thank you.

mrcjkb:
Closes #531

>[!IMPORTANT]
>
>This PR has two commits and should be rebase-merged.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,3 +5,4 @@ resolver = "2"
 
 [profile.release]
 lto = true
+codegen-units = 1
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,3 +2,6 @@
 members = ["lux-lib", "lux-cli", "lux-workspace-hack", "xtask", "lux-lua"]
 default-members = ["lux-cli", "lux-lib"]
 resolver = "2"
+
+[profile.release]
+lto = true


vhyrro (approved):

vhyrro:
should get better with caching I hope :p

PR #597, opened 2025-04-14
revert: use compilation target to get platform identifier

mrcjkb:
We have too many places where we use `cfg` to determine things that are specific to the target (for example, file extensions).
So using `cc::Build` to determine the platform identifier isn't helpful at this point.
For now, let's just expect users to use a version of lux that was packaged for the target they want to build for.
On Windows, that means we will have to distribute separate lux packages for MSVC and Cygwin.


vhyrro:
LGTM, I like this change. Should we rename the function to `for_current_platform()` or just back to `current_platform()`?

mrcjkb:
> Should we rename the function to `for_current_platform()` or just back to `current_platform()`?

Sure

vhyrro (approved):

PR #599, opened 2025-04-14
fix(pack): regression in manifest creation

mrcjkb:
Introduced in #579 (we don't run this test in CI - which is bad).


mrcjkb:
Merging as this is trivial

PR #605, opened 2025-04-15
fix(build/tree-sitter): race condition when building many parsers simultaneously

vhyrro:
I can't test directly through the devshell since I'm in a region with poor internet currently, but a quick run of `parallel` with 4 separate treesitter installations has stopped throwing an error, so I assume this has fixed the problem. Further testing will be necessary.


mrcjkb:
suggestion: we have a test case that builds a tree-sitter parser. I suppose we could change it to spawn multiple builds in parallel.

Issue #329, opened 2025-01-12
[Feature] git dependencies


mrcjkb:
This would likely require us to create a luarocks build backend that is capable of handling it.

- Should this be its own `git_dependencies` table?
- Or a `[dependencies].<foo>.git = github:owner/repo`
- Or something that allows to embed a rockspec:

```lua
dependency_rockspecs = {
[==[
package = "foo",
version = "1.0.0",
source.url = "bla",
build = {
  type = "builtin",
},
]==],
}
```

Even though it's more verbose, I'm leaning towards the third option, because it's very flexible (not limited to git dependencies) and we can already encode rockspecs in the lockfile.

vhyrro:
I guess my only gripe with all of these approaches is that they aren't compatible with the rockspec format. I see two possible approaches here:

1. We disallow a rockspec with git dependencies to be uploaded to `luarocks.org` (`cargo` does this for `crates.io`)
2. We instead opt for an adapter architecture (`luarocks-build-git-adapter`), which allows you to specify git dependencies in the build step, all while being able to specify a different build step that runs as a subcommand of sorts (e.g. `build.type = "git-adapter"` and `build.subtype = "builtin"`).

mrcjkb:
Ah crap, luarocks doesn't support any off-spec fields. That means we'd have to change it to

```lua
build = {
  dependency_rockspecs = {
[==[
package = "foo",
version = "1.0.0",
source.url = "bla",
build = {
  type = "builtin",
},
]==],
  }
}
```

which is kind of ugly.
But with that, luarocks won't complain (even if the rockspec uses the `builtin` build backend).
It'll just ignore the off-spec fields.
So I don't think it would be strictly necessary to disallow uploading rockspecs with off-spec fields to luarocks.org.

vhyrro:
I'm of the opinion that a given rock should provide the same output no matter what plugin manager consumes it. Doing off-spec fields is dangerous in this way because it becomes an anarchy of "this package manager can build this rockspec but this one can't". I think disallowing is safer (while still making the functionality cool).

mrcjkb:
With out [luarocks-build-embedded-rockspecs](https://github.com/nvim-neorocks/luarocks-build-embedded-rockspecs) build backend, we now should have everything we need to be able to implement git dependencies in our `rocks.toml` spec, and translate them to a rockspec that can also be used by luarocks.

mrcjkb:
[...]

PR #623, opened 2025-04-23
feat(lux-lib): more lenient dev version parsing

mrcjkb:
This is in preparation for #329.
It will allow us to encode things like git commit hashes and arbitrary tags as `DevVer` versions.


mrcjkb:
> Rather than this, wouldn't we prefer to go the cargo way and make git dependencies inline tables? `my_dep = { git = "https://example.com", rev = "xxx" }`? A little more work, but should make it more extensible for future expansions.

That's the plan. But by reusing the `version` field, there's no possibility for `version` and `rev` to conflict with each other.

vhyrro (approved):

PR #624, opened 2025-04-23
feat(lib/install): support installing from alternate sources

mrcjkb:
Stacked on #623.
A preparation for #329.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2321,7 +2321,7 @@ dependencies = [
  "futures-core",
  "futures-sink",
  "futures-util",
- "getrandom 0.3.1",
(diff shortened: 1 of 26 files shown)


vhyrro (approved):

vhyrro:
worth fixing the build before merging :p

mrcjkb:
@vhyrro I've made a new change:

- We now build alternate sources using the build backend from the project toml or rockspec (closes #647)
  If the project rockspec/toml specifies a custom luarocks build backend, we fail, because we don't know if we can provide the build dependencies.
- Each build backend now has its own error type.

Issue #631, opened 2025-04-27
Running lx add fails with external dependency specification


cootshk:
Brand new install of lux, on a NixOS system

When I run `lx add lapis` on a brand new lux project, I get this error:
```
thread 'main' panicked at lux-cli/src/bin/lx.rs:88:69:
called `Result::unwrap()` on an `Err` value: syncing dependencies with the project lockfile failed.

Caused by:
   0: could not parse rockspec: deserialize error: conflicting external dependency specification
   1: deserialize error: conflicting external dependency specification

Location:
    /build/source/lux-cli/src/utils/project.rs:34:14
stack backtrace:
   0:     0x562e714ab3be - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::h9edbd6e38a8b0805
   1:     0x562e70ea5073 - core::fmt::write::h7b1248e5e0c79c78
   2:     0x562e71470a42 - std::io::Write::write_fmt::h5e301665499081bf
   3:     0x562e714a87f5 - std::panicking::rust_panic_with_hook::h100726ba9570b85a
   4:     0x562e714aba48 - std::panicking::begin_panic_handler::{{closure}}::h141712493bfacf0c
   5:     0x562e714ab9a9 - std::sys::backtrace::__rust_end_short_backtrace::h891003731531c924
   6:     0x562e714a807c - rust_begin_unwind
   7:     0x562e70bcab7f - core::panicking::panic_fmt::h1df68d570cb2382b
   8:     0x562e70bcb1d5 - core::result::unwrap_failed::h846e3c141fa2fc46
   9:     0x562e70dd161a - lx::main::{{closure}}::h583fd553e590d7db
  10:     0x562e70daf7fe - lx::main::hab71232452e8dfb9
  11:     0x562e70c33e13 - std::sys::backtrace::__rust_begin_short_backtrace::h7901ad2b8faacc5e
  12:     0x562e70dddc23 - main
  13:     0x7f1f1e22a47e - __libc_start_call_main
  14:     0x7f1f1e22a539 - __libc_start_main_impl
  15:     0x562e70c038b5 - _start
  16:                0x0 - <unknown>
```

cootshk:
program works perfectly fine after running `lx remove lapis`.

Running `lx add argparse` works perfectly fine but `lx add luaossl` doesn't.

mrcjkb:
Thanks for reporting 🙏 
Fix is on the way.

PR #635, opened 2025-04-27
fix(cli): suggest `--no-lock` instead of `--ignore-lockfile`

cootshk:
`--ignore-lockfile` isn't a valid argument.

I noticed this in #631.


mrcjkb (approved):
Thanks 😄🙏

PR #650, opened 2025-04-29
fix(build): transitive dependencies added as dependencies of main rock

mrcjkb:
When running `lx build`, the entry for the built rock in the install tree's lux.lock has all transitive dependencies listed as dependencies.


mrcjkb:
Merging, as this is trivial.

PR #651, opened 2025-04-29
fix(build): dependencies added as install tree entrypoints

mrcjkb:
Stacked on #650.

Because we install dependencies before building a local project, they get added to the install tree's lux.toml as entrypoints.
When installing the local project as a rock, we want it to be the only entrypoint.


mrcjkb:
Merging, as this is trivial.

PR #665, opened 2025-05-09
feat: use pkg-config to probe lux-lua

### LINKED ISSUE #655: [Feature] If `LUX_LIB_DIR` isn't defined, use pkg-config to find lux-lua
This would simplify the Nix derivation, as we wouldn't have to create a wrapper that provides a lux-lua bundle for all Lua versions.
Instead, Nix users could enter a devShell that has the required `lux-lua` version, and lux should find it automatically.

mrcjkb:
Closes #655 

This implementation prioritizes pkg-config over `LUX_LIB_DIR`.


lux-lib/src/path.rs
```diff
@@ -63,9 +63,19 @@ impl Paths {
...
+            .cargo_metadata(false)
+            .env_metadata(false)
+            .probe(&lib_name)
+            .ok()
+            .and_then(|library| library.link_paths.first().cloned())
+            .unwrap_or(
+                option_env!("LUX_LIB_DIR")
+                    .map(PathBuf::from)
+                    .unwrap_or_else(|| Config::get_default_data_path().unwrap()),
+            )
```
vhyrro, line 78: question: shouldn't we prioritise `LUX_LIB_DIR`, since that's a compile time constant that's injected into the binary at build time? If the packager decided to set it, they probably would want it to be respected at all times.

vhyrro (approved):
awesome work, this will make things so much simpler, just a question

lux-lib/src/path.rs
mrcjkb: Good point.

PR #666, opened 2025-05-09
refactor!: unify `Sync` by making it take in a `Project`

vhyrro:
this unifies the API and further prevents mistakes.

one extra issue: we always build a luarocks installation for build dependencies - shouldn't we incorporate this deeper into the API (e.g. as part of `Sync`), instead of hoping that users provide the correct tree? That should be for a different PR, and I make an issue for that if you'd wish :)

### DIFF
--- a/lux-lib/src/operations/sync.rs
+++ b/lux-lib/src/operations/sync.rs
@@ -3,14 +3,14 @@ use std::{io, sync::Arc};
 use crate::{
     build::BuildBehaviour,
     config::Config,
-    lockfile::{
-        LocalPackage, LocalPackageLockType, LockfileIntegrityError, PackageSyncSpec,
-        ProjectLockfile, ReadWrite,
-    },
+    lockfile::{LocalPackage, LocalPackageLockType, LockfileIntegrityError},
     luarocks::luarocks_installation::LUAROCKS_VERSION,
     package::{PackageName, PackageReq},
     progress::{MultiProgress, Progress},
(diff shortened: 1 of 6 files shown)


mrcjkb:
> shouldn't we incorporate this deeper into the API (e.g. as part of Sync), instead of hoping that users provide the correct tree? That should be for a different PR, and I make an issue for that if you'd wish :)

:thinking: Makes sense to me.

lux-lib/src/operations/sync.rs
```diff
@@ -238,123 +246,122 @@ async fn do_sync(
...
+//             .tree(Some(temp_dir.path().into()))
+//             .build()
+//             .unwrap();
+//         let dest_tree = config.tree(LuaVersion::Lua51).unwrap();
+//         let report = Sync::new(&dest_tree, &mut source_lockfile, &config)
+//             .sync_dependencies()
+//             .await
+//             .unwrap();
+//         assert!(report.removed.is_empty());
+//         assert!(!report.added.is_empty());
```
mrcjkb, line 289: question: This is all commented out. Are we losing important test coverage here?

vhyrro: ... this is from a messed up jj evolog restoration. well done on catching this, silly me.

vhyrro:
all should be fixed now, waiting for CI

mrcjkb (approved):

vhyrro:
the CI errors are unrelated but still worth fixing in a follow-up PR :)

PR #667, opened 2025-05-09
feat(cli): `lx add` for git dependencies

### LINKED ISSUE #645: `lx add` and `lx remove` for git dependencies
#644 introduces basic support for git dependencies.
It would be nice to be able to `lx add` and `lx remove` them.

If specying a dependency without a version requirement, e.g. `lx add "github:owner/repo"`, then lux will need to checkout the repository and get the current revision or latest semver tag.

mrcjkb:
Closes #645.

There's no need to implement `lx remove` for git dependencies, as we remove them by name.

### DIFF
--- a/
+++ b/
@@ -68,14 +68,10 @@ pub enum IntoRemoteRockspecError {
 }
 
 #[derive(Error, Debug)]
+#[error(transparent)]
 pub enum ProjectEditError {
-    #[error(transparent)]
     Io(#[from] tokio::io::Error),
-    #[error(transparent)]
     Toml(#[from] toml_edit::TomlError),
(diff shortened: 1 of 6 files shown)


vhyrro:
question: didn't we want to separate git repositories with a `--git` flag, just like we do `--build` and `--test`? feels a little dirty to be parsing both at once :sweat_smile:

mrcjkb:
> question: didn't we want to separate git repositories with a `--git` flag, just like we do `--build` and `--test`? feels a little dirty to be parsing both at once 😅

I thought of that, but then we would need --git-test and --git-build too.
As long as we only support specific host prefixes, I think this makes for a cleaner API.

vhyrro:
right, this sounds like a good compromise for now then.

PR #671, opened 2025-05-10
feat(cli): `lx update` for git dependencies

### LINKED ISSUE #646: `lx update` for git dependencies
With #644, we can install git dependencies. But we need to make sure `lx update` (`Project::upgrade`) correctly updates them to the latest revision or semver tag.

Currently, `lx update` will overwrite the version of a git dependency if a newer one is found on luarocks.org, ignoring the `rev` (if set). This can be problematic, because without a `rev` field, lux treats the `version` field as the `checkout_ref`.

mrcjkb:
Stacked on #670.

This PR completes lux's support for git dependencies.

Closes #646.


vhyrro (approved):
just a small conflict that needs to be resolved

PR #674, opened 2025-05-10
ci: add lux-lua to release-plz

mrcjkb:


mrcjkb:
Hmm, come to think of it, I'm not sure if we want to use release-plz for lux-lua, as it's a Lua library, not a Rust library.
@vhyrro what do you think?

vhyrro:
yeah i don't think it's a good idea to add lux-lua to release-plz. It won't even be able to detect API changes properly because it's all happening on the Lua side. We can bump the lux-lua versions manually - I'm currently fine with it being frozen at 0.1, since it's not even complete yet.

mrcjkb:
Technically, we should be bumping minor versions if we update lux-cli.
But for now, I'm fine with doing that manually,

PR #679, opened 2025-05-11
refactor!: unify `Install` tree operations

vhyrro:
WIP - similar in vain to the `Sync` refator, aiming to limit mistakes that can be made as a result of mismatched trees.

### DIFF
--- a/lux-cli/src/build.rs
+++ b/lux-cli/src/build.rs
@@ -9,7 +9,6 @@ use lux_lib::{
     config::Config,
     luarocks::luarocks_installation::LuaRocksInstallation,
     operations::{Install, PackageInstallSpec, Sync},
-    package::PackageName,
     progress::MultiProgress,
     project::Project,
     rockspec::Rockspec,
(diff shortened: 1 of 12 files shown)


lux-cli/src/build.rs
```diff
@@ -93,9 +93,11 @@ pub async fn build(data: Build, config: Config) -> Result<()> {
         if !build_dependencies_to_install.is_empty() {
             let bar = progress.map(|p| p.new_bar());
             luarocks.ensure_installed(&bar).await?;
-            Install::new(luarocks.tree(), luarocks.config())
+            // NOTE: Is this doing what we are expecting? Should it just
+            // install into the luarocks tree or should it install into the project?
```
vhyrro, line 97: cc @mrcjkb i think i found a bug through this. The old code supplies the luarocks tree but then also supplies `.project()`, but all that ended up doing is using the project's tree :shrug:

mrcjkb: I think what we actually want here is a separate build-dependencies tree for the project (and maybe luarocks too?), like we have for test dependencies.

But we definitely shouldn't be installing build dependencies in the project tree.

mrcjkb (approved):

lux-cli/src/build.rs
mrcjkb: Resolved in #692.

PR #680, opened 2025-05-11
fix(build): separate xtask-lua crate for lux-lua

mrcjkb:
Becauses our `xtask` crate depends on lux-cli:

- The nix build for a non-luajit lux-lua depends on luajit.
- We have to build lux twice (once for xtask with the luajit feature, once for lux-lua).

Solution: A separate `xtask-lua` crate for lux-lua, which does not depend on lux-cli.


vhyrro (approved):
fantastic change. This should allow us to unify the `lux-lib` and `lux-lua` features (so they both build for the same lua target all the time) and we should be able to readd `lux-lua` back to the workspace by default so `cargo build` can build it :D

PR #684, opened 2025-05-11
feat: Autogenerate a .gitignore file

### LINKED ISSUE #682: [Suggestion] Alter `.gitignore` when creating a project
When making a project, we should create a `.gitignore` which ignores the `.lux` directory. If one already exists, we should add to the file (I'm fairly sure a Rust crate exists for this purpose).

cootshk:
Closes #682

Basically just automatically add a `.gitignore` file containing `.lux/` to the project root

### DIFF
--- a/lux-cli/src/project/new.rs
+++ b/lux-cli/src/project/new.rs
@@ -358,10 +358,7 @@ type = "builtin"
     } else {
         std::fs::create_dir(&main_dir)?;
         std::fs::write(main_dir.join("main.lua"), r#"print("Hello world!")"#)?;
-        std::fs::write(
-            main_dir.join(".gitignore"),
-            ".lux/"
-        )?;
+        std::fs::write(main_dir.join(".gitignore"), ".lux/")?;
     }
 
     println!("All done!");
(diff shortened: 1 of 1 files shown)


cootshk:
Also maybe make a question "do you want to initialize a git repository?" that runs `git init` in the new project questions?

lux-cli/src/project/new.rs
```diff
@@ -358,6 +358,7 @@ type = "builtin"
     } else {
         std::fs::create_dir(&main_dir)?;
         std::fs::write(main_dir.join("main.lua"), r#"print("Hello world!")"#)?;
+        std::fs::write(main_dir.join(".gitignore"), ".lux/")?;
```
mrcjkb, line 361: issue: this will overwrite the content if there's an existing .gitignore.
We need to append to existing .gitignore files if there's no existing .lux entry.

mrcjkb (changes requested):
Thanks 🙏

lux-cli/src/project/new.rs
cootshk: This only gets run if `main_dir` doesn't already exist, right? (line 353)

mrcjkb: Oh, right. Fair point.

mrcjkb (approved):

PR #689, opened 2025-05-12
fix!: treat unknown string versions as `< SemVer`

### LINKED ISSUE #688: [Issue] Potential problem in version comparison
See snippet:
```
╰─❯ lx outdated
fzy
└── 1.0.3-1 => main-1

fidget.nvim
└── 1.6.0-1 => main-1
```

I assume this is undesirable behaviour - we shouldn't compare versions that consist of just words.

mrcjkb:
Fixes #688.

This PR introduces a `StringVer` version variant, which can be an arbitrary string.
It differs from `DevVer` as follows:

- `DevVer` can only be `"dev"` or `"scm"`
- `StringVer` is treated as `< SemVer`
- `DevVer` is treated as `> SemVer`

Background info: luarocks-site will add `"dev"` and `"scm"` versions to the dev manifest, so you need a `--dev` flag to install them.
But it will happily add an arbitrary string version to the root manifest.


mrcjkb:
Todo:

- [ ] We should refuse to upload `StringVer` packages.

mrcjkb:
hakari always wants to revert the getrandom dependency :thinking:

vhyrro (approved):

vhyrro:
> hakari always wants to revert the getrandom dependency 🤔

yeah it appears something is slightly bugged or misconfigured, as is our changelog generation

PR #694, opened 2025-05-13
ci(cd): don't build for windows targets

vhyrro:
This should make the build hopefully succeed now and allow us to vendor valid builds for `cargo binstall`. In the near future, I also want to extend this to building lux-lua so that lux.nvim can also pull it down `blink.cmp` style.


vhyrro:
merging, since trivial

PR #697, opened 2025-05-13
fix: remove hash field from `RockSourceInternal`

mrcjkb:
Back when I implemented this, I wasn't aware that luarocks would error if the source table has any fields it doesn't know about.


mrcjkb:
Merging as trivial.

Issue #700, opened 2025-05-14
Error: failed to parse manifest from configuration: failed to unzip manifest file


kuator:
I'm on ubuntu 24, I had a bunch of errors (I figured out it was due to missing prerequisites), so I had to install `sudo apt-get -y update && sudo apt-get install -y libgpgme-dev libgpg-error-dev luajit libluajit-5.1-dev`. After that Lux finally installed
```bash
$ lx --version
lux-cli 0.4.5
```
Now I'm trying to follow the tutorial and when I try to add a dependency, there's an error

```bash
$ lx add argparse@0.7     
Error: failed to parse manifest from configuration: failed to unzip manifest file

Caused by:
   0: failed to unzip manifest file
   1: invalid Zip archive: Could not find EOCD

Location:
    ~/.local/share/cargo/registry/src/index.crates.io-6f17d22bba15001f/lux-cli-0.4.5/src/add.rs:46:14
```

kuator:
> lx config show

HI, the output of the command:
```bash
$ lx config show
server = "https://luarocks.org/"
extra_servers = []
tree = "/tmp/tmp/my-lua-project/.lux"
luarocks_tree = "/tmp/tmp/my-lua-project/.lux/build_dependencies"
cache_dir = "/home/evakuator/.cache/lux"
data_dir = "/home/evakuator/.local/share/lux"
no_project = false
enable_development_packages = false
verbose = false

[timeout]
secs = 30
nanos = 0

[variables]
LIBFLAG = "-shared"
CMAKE = "cmake"
LIB_EXTENSION = "so"
OBJ_EXTENSION = "o"
MAKE = "make"
CFLAGS = "-O2"

[external_deps]
bin_patterns = ["?"]
include_patterns = ["?.h"]
lib_patterns = ["lib?.so", "lib?.a"]
bin_subdir = "bin"
include_subdir = "include"
lib_subdirs = ["lib", "lib64"]
search_prefixes = ["/usr", "/usr/local", "/opt/local", "/opt"]

[external_deps.prefixes]

[entrypoint_layout]
etc = "etc"
opt_etc = "etc"
conf = "conf"
doc = "doc"
```
I can't find a line where it show a lua version, I think I messed up something during installation

mrcjkb:
> I can't find a line where it shows a lua version

Ah, I guess it doesn't output it if it's inferred from your lua installation. What's the output of `lua -v`?

[...]

PR #701, opened 2025-05-14
feat: more detailed zip error messages

mrcjkb:
This ought to help troubleshoot what is actually going on in #700, before we implement a plaintext mainfest fallback as a workaround.


mrcjkb:
Trivial - merging.

PR #726, opened 2025-05-21
fix: unable to parse large luarocks manifest

### LINKED ISSUE #725: Unable to load large luarocks manifest
The luarocks `manifest-5.3` has reached a size that can no longer be loaded. It fails with:

```
Error: error initialising remote package DB: failed to parse manifest from configuration: failed to parse manifest: syntax error: [string "lux-lib/src/manifest/mod.rs:200:13"]:201129: main function has more than 65536 constants

Caused by:
   0: failed to parse manifest from configuration: failed to parse manifest: syntax error: [string "lux-lib/src/manifest/mod.rs:200:13"]:201129: main function has more than 65536 constants
   1: failed to parse manifest: syntax error: [string "lux-lib/src/manifest/mod.rs:200:13"]:201129: main function has more than 65536 constants
   2: syntax error: [string "lux-lib/src/manifest/mod.rs:200:13"]:201129: main function has more than 65536 constants

Location:
    /home/mrcjk/git/github/nvim-neorocks/lux/lux-cli/src/install.rs:36:5
```

Not sure if this is a lux issue or a luarocks-site issue.
And I'm not sure if there's a way for us to work around it.

mrcjkb:
Fixes #725.

We build Lux with luajit and #725 is a luajit limitation.
Solution: Build Lux with Lua 5.4.

@vhyrro fyi, in lux.nvim, we'll have to use lux-lua built against Lua 5.1 (not luajit) if we want to prevent it from running into that issue.
(We should do that anyway, as not all Neovim builds ship with luajit).

I've also enabled the `vendored-lua` feature by default, so that we can build lux and run tests in a devShell with a different lua version.

### DIFF
--- a/lux-cli/Cargo.toml
+++ b/lux-cli/Cargo.toml
@@ -58,7 +58,7 @@ path = "../lux-lib/"
 features = ["clap"]
 
 [features]
-default = ["luajit"]
+default = ["lua54", "vendored-lua"]
 lua51 = ["lux-lib/lua51"]
 lua52 = ["lux-lib/lua52"]
 lua53 = ["lux-lib/lua53"]
--- a/lux-lib/Cargo.toml
+++ b/lux-lib/Cargo.toml
@@ -96,7 +96,7 @@ assert_fs = "1.1.3"
 predicates = "3.1.3"
 
 [features]
-default = ["luajit"]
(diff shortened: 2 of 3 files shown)


mrcjkb:
Merging, as this is trivial.

PR #730, opened 2025-05-22
refactor!(lua-rockspec): split out lua from dependencies

### LINKED ISSUE #427: [Refactor/LuaRockspec] split lua version out from dependencies when parsing
Having to filter out `lua` when iterating over dependencies is really annoying.
Let's split it out when parsing the rockspec.

mrcjkb:
Closes #427.


mrcjkb:
> great refactor, but we should change make sure that when converting back to a rockspec we merge the lua key back into the dependencies.

When we create a lua rockspec from toml, we already do that.
Afaik, we never convert a parsed Lua rockspec back to Lua (we just store the string content).

vhyrro (approved):

PR #749, opened 2025-05-27
feat!(cli/check): respect ignore files by default

mrcjkb:
cc @cootshk


cootshk (approved):
LGTM (I don't know that much rust but I think I understand)

PR #751, opened 2025-05-28
feat: keep lux-cli and lux-lua versions in sync

mrcjkb:
We don't have a way to update lux-lua with release-plz.
So let's just keep it synced with lux-cli. This means that lux-lua releases will have to be linked with lux-cli releases. But I that's probably a good thing, because it helps ensure compatibility.


mrcjkb:
Need to update the versions in the nix overlay.

PR #752, opened 2025-05-28
Add: .gitignore inside .lux to ignore all files in that directory Close #750

### LINKED ISSUE #750: [Suggestion] what if we just put the .gitignore into the .lux folder?
Just make it say `*` so that nothing gets included

That way we don't have to worry about overwriting an existing file and it's also less clogged

(It's 2 am, sorry if my engrish makes no sense rn, I need sleep)

Feelfeel20088:
see #750

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,4 +19,9 @@
 /*.rockspec
 *.snap.new
 
+# Lux project files
+/.lux/*
+/lux.lock
+/lux.toml
+
 .pre-commit-config.yaml
--- a/.gitignore
+++ b/.gitignore
@@ -19,9 +19,4 @@
 /*.rockspec
 *.snap.new
 
-# Lux project files
-/.lux/*
-/lux.lock
-/lux.toml
-
 .pre-commit-config.yaml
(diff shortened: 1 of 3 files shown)


cootshk (changes requested):
Change it in the code (in lux-cli/project/new) pls

mrcjkb (commented):
Thanks 🙏 

After some consideration, I don't think this is what we want, because it means people have to check the `.lux` directory into vcs.

cootshk:
> Thanks 🙏 
> 
> 
> 
> After some consideration, I don't think this is what we want, because it means people have to check the `.lux` directory into vcs.

Git automatically ignores directories with no tracked files

having a directory structure of:

- .lux/
  -> .gitignore (contains "*")
  -> lua5.4/
    -> ...

won't track the .lux folder

cootshk:
image
notice there's no project-wide `.gitignore`, just the one in the folder

Feelfeel20088:
Yeah what he said :()

mrcjkb:
Superseded by #753.

PR #757, opened 2025-05-30
fix!: more robust lua binary detection

### LINKED ISSUE #745: `lx --lua-version=jit run` tries to use lua 5.4 and fails
I'm running on Mac, with all of my lua versions installed via `homebrew`.

Image

Image

Maybe check for the existence of a `luajit` or a `lua5.1` or `lua5.4` binary? `dnf` makes a luajit binary but not a `lua5.4` one, but `brew` makes both

mrcjkb:
Fixes #745.

Our `lua` binary detection currently doesn't take the specified Lua version into account.


mrcjkb:
This has decent enough test coverage. Merging.

PR #763, opened 2025-06-01
fix: don't set `LUA_INIT` if lux-lua not present

### LINKED ISSUE #755: `lx path`: don't set `LUA_INIT` if lux-lua isn't installed
Image

Image

mrcjkb:
Closes #755

### DIFF
--- a/lux-lib/src/path.rs
+++ b/lux-lib/src/path.rs
@@ -6,7 +6,7 @@ use thiserror::Error;
 
 use crate::{
     build::utils::c_dylib_extension,
-    config::{Config, LuaVersion},
+    config::LuaVersion,
     tree::{Tree, TreeError},
 };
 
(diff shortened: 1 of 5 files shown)


lux-lib/src/path.rs
```diff
@@ -63,28 +63,14 @@ impl Paths {
...
-            .map(PathBuf::from)
-            .unwrap_or_else(|| {
-                pkg_config::Config::new()
-                    .print_system_libs(false)
-                    .cargo_metadata(false)
-                    .env_metadata(false)
-                    .probe(&lib_name)
-                    .ok()
-                    .and_then(|library| library.link_paths.first().cloned())
-                    .unwrap_or_else(|| Config::get_default_data_path().unwrap())
```
mrcjkb, line 77: @vhyrro I've removed the `.unwrap_or_else(|| Config::get_default_data_path().unwrap())` line, because we can't determine if lux-lua is available with it.
Do you know if this could break something?

vhyrro: I can't see an immediate problem with that, so I'd say it's fine

vhyrro (dismissed):

PR #767, opened 2025-06-02
fix: properly quote complex keys when generating rockspec

vhyrro:


mrcjkb (approved):

mrcjkb:
Looks like a test needs to be updated

mrcjkb (approved):

PR #781, opened 2025-06-10
feat(lux-lua): state functions and search functionality

vhyrro:
lays out the foundations for lux.nvim operations

I'm torn between calling this "state" (since we're pulling from global state - user-local trees and manifests) and "operations" (even though not all of the functions will be operations, i.e. `installed_packages`). Thoughts?


mrcjkb:
Why not have a `state` module and an `operations` module? 
Search sounds more like an operation to me.

lux-lua/src/lib.rs
```diff
@@ -31,6 +32,7 @@ fn lux(lua: &Lua) -> LuaResult<LuaTable> {
     )?;
     exports.set("config", config::config(lua)?)?;
     exports.set("project", project::project(lua)?)?;
+    exports.set("state", operations::operations(lua)?)?;
```
mrcjkb:
```suggestion
    exports.set("operations", operations::operations(lua)?)?;
```

mrcjkb (changes requested):

PR #791, opened 2025-06-15
feat(repl): alias `exit` to `os.exit()`

### LINKED ISSUE #526: feat: alias `exit` to `os.exit` in repl
`exit` is a commonly used command. it would be nice if we had that instead of `os.exit()`.

cootshk:
Fixes #526 
Succeeds #790 (I renamed the branch)

You can set `exit` equal to `os.exit` so that you can run `exit()` in a lua repl

### DIFF
--- a/lux-lib/src/path.rs
+++ b/lux-lib/src/path.rs
@@ -94,9 +94,9 @@ impl Paths {
     pub fn init(&self) -> String {
         format!(
             r#"
-        if _VERSION:find('{}') then {LUA_INIT} end
-        exit = os.exit
-        setmetatable(exit, {__tostring = function(...) return os.exit() end})
+            exit = os.exit
+            setmetatable(exit, {__tostring = function(...) return os.exit() end})
+            if _VERSION:find('{}') then {LUA_INIT} end
         "#,
             self.version
         )
--- a/lux-lib/src/path.rs
+++ b/lux-lib/src/path.rs
@@ -95,7 +95,7 @@ impl Paths {
         format!(
             r#"
             exit = os.exit
-            setmetatable(exit, {__tostring = function(...) return os.exit() end})
+            setmetatable(exit, {{__tostring = function(...) return os.exit() end}})
             if _VERSION:find('{}') then {LUA_INIT} end
         "#,
             self.version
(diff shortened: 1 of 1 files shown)


cootshk:
The clippy build is failing because it wants a double brace. If I use a double brace, lua doesn't set the metatable correctly and the macro won't work

cootshk:
(maybe use `include_str!()` to load a .lua file at compile time?)

lux-lib/src/path.rs
```diff
@@ -92,7 +92,14 @@ impl Paths {
 
     /// Get `$LUA_INIT`
     pub fn init(&self) -> String {
-        format!("if _VERSION:find('{}') then {LUA_INIT} end", self.version)
+        format!(
+            r#"
+            exit = os.exit
+            setmetatable(exit, {__tostring = function(...) return os.exit() end})
```
mrcjkb: issue: `{` and `}` need to be escaped in a format string.
suggestion: I don't think it needs to be wrapped in a function?

```suggestion
            setmetatable(exit, {{__tostring = os.exit}})
```

cootshk: It does, because it gets the `self` argument passed in, which we need to discard

mrcjkb:
> The clippy build is failing because it wants a double brace. If I use a double brace, lua doesn't set the metatable correctly and the macro won't work

I don't think the final formatted string will have a double brace

mrcjkb (approved):

cootshk:
> > The clippy build is failing because it wants a double brace. If I use a double brace, lua doesn't set the metatable correctly and the macro won't work
> 
> I don't think the final formatted string will have a double brace

I've been testing through just `LUA_INIT="..." lx lua` and made a quick script to do that. If rust removes the brace, then that's a rust thing (I'm not a rust user)

mrcjkb:
> > > The clippy build is failing because it wants a double brace. If I use a double brace, lua doesn't set the metatable correctly and the macro won't work
> > 
> > 
> > I don't think the final formatted string will have a double brace
> 
> I've been testing through just `LUA_INIT="..." lx lua` and made a quick script to do that. If rust removes the brace, then that's a rust thing (I'm not a rust user)

Yep, it removes the extra brace. What you pass to the `format!` macro isn't a regular string, but a format string, in which braces have special meaning and need to be escaped.

PR #796, opened 2025-06-15
feat!: break apart `LuaVersion::from` into two subfunctions for more fine-grained behaviour control

vhyrro:
This PR changes the `LuaVersion::from` function into two distinct functions - one that gets the current version just based off of the currently installed Lua binary, the other one which checks the current project's lux.toml in order to determine the expected version.

Setting this as draft, I still need to test if the CLI has properly benefitted from these changes.

### DIFF
--- a/lux-lib/src/config/mod.rs
+++ b/lux-lib/src/config/mod.rs
@@ -12,6 +12,7 @@ use thiserror::Error;
 use tree::RockLayoutConfig;
 use url::Url;
 
+use crate::project::Project;
 use crate::tree::{Tree, TreeError};
 use crate::variables::GetVariableError;
 use crate::{
@@ -175,7 +176,7 @@ fn lux_lib_resource_dir() -> Option<PathBuf> {
 
 #[derive(Error, Debug)]
 #[error(
-    r#"lua version not set.
+    r#"lua version couldn't be inferred!
 Please provide a version through `lx --lua-version <ver> <cmd>`
 Valid versions are: '5.1', '5.2', '5.3', '5.4', 'jit' and 'jit52'.
 "#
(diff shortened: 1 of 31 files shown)


vhyrro:
the changes in this PR were generally done on a best-guess basis - commands which aren't project dependent use `LuaVersion::from_config()`, and commands that do contextually use information from the project use `LuaVersion::from_current_project_or_config()`. Not sure if there's a comprehensive way of writing tests for this without writing 20 of them for each case.

lux.toml
vhyrro, line 1: issue: temporary (for quick testing), should be reverted before merge

vhyrro:
inside of the nix environment, this code works great. as soon as I try to build a project in my arch install, it fails to recognize that i have lua 5.1 installed on my system and instead tries to force lua 5.4, so that's one last fix that I need to implement before this can be fully merged.

mrcjkb:
> inside of the nix environment, this code works great. as soon as I try to build a project in my arch install, it fails to recognize that i have lua 5.1 installed on my system and instead tries to force lua 5.4, so that's one last fix that I need to implement before this can be fully merged.

And it works fine on `master`?

vhyrro:
No, this is a consistent error. I've managed to locate the source of the issue - as a fallback, if no version is supplied, we only parse the output of executing the `lua` binary. On nix this makes sense, because `lua` should be symlinked to the lua version declared in the environment. However, we never handle the case where multiple Lua versions are installed on a given system with binary names like `lua5.1`.

In general, the way we handle version checking is currently flawed, I noticed. `LuaVersion::from_config` shouldn't be a part of the enum and instead we should make something like `SupportedLuaVersions::from_config` which returns a *list* of Lua versions the current system supports. This should let us handle the case of enumerating all available versions with pkg-config or whatever.

mrcjkb:
Have you rebased?

- https://github.com/lumen-oss/lux/pull/1219

mrcjkb:
> In general, the way we handle version checking is currently flawed, I noticed. LuaVersion::from_config shouldn't be a part of the enum and instead we should make something like SupportedLuaVersions::from_config which returns a list of Lua versions the current system supports. This should let us handle the case of enumerating all available versions with pkg-config or whatever.

That sounds like a good idea.

lux-lib/src/config/mod.rs
```diff
@@ -175,16 +176,34 @@ fn lux_lib_resource_dir() -> Option<PathBuf> {
...
+
+    pub fn from_current_project_or_config(config: &Config) -> Result<Self, LuaVersionUnset> {
+        match config.lua_version.clone() {
+            Some(lua_version) => Ok(lua_version),
+            None => match Project::current_or_err() {
+                Ok(project) => project.lua_version(config).map_err(|_| LuaVersionUnset),
+                Err(_) => LuaVersion::from_detected(),
+            },
+        }
     }
```
mrcjkb, line 207: issue: This introduces a cyclic dependency between the `config` and `project` modules.
Perhaps it's better to have a `Project::lua_version_from_current_or_config` method?

vhyrro: in general having `LuaVersion` in `config.rs` makes little sense. It'd probably be wiser to extract this to a separate module and then this cycle is no longer as bad? Keeping this API in the enum is much better for discoverability I think :thinking:

vhyrro: optionally, I could turn this into `from_project_or_config(&Project, &Config)` where a project has to be provided

mrcjkb: > in general having LuaVersion in config.rs makes little sense. 

Agreed, it should probably go in it's own module. But because `config` depends on it, it wouldn't solve the dependency cycle issue.

Cycles are always bad, as they get in the way of refactoring and lead to architectural issues. I'm a bit disappointed that rust/cargo allows them and would like to add some checks to CI to prevent them.

> Keeping this API in the enum is much better for discoverability I think 🤔

That's a fair point. How about we define it via a trait that we implement in the `project` module?

vhyrro:
I'll rethink this PR and try to implement a better solution :)

PR #801, opened 2025-06-16
[troubleshooting] lua-src upgrade

mrcjkb:
Ignore this PR. Troubleshooting the build failure on Windows after upgrading lua-src


vhyrro:
best of luck :sweat_smile:

PR #815, opened 2025-06-24
Some fixes and improvements to how we handle `LUA_INIT` [do not squash]

mrcjkb:
See individual commits


cootshk:
```lua
exit = {} 
setmetatable(exit, {
    __call=(
        function(...) return os.exit() end
    ),__tostring=(
        function(...) return "To exit, type 'exit()' or press <C-d>." end)
})
```
`exit` will print the message on lua5.3 and 5.4, `print(exit)` will print the message on all versions (including JIT), and `exit()` also works on all versions.

This is similar to how python handles it.
image

cootshk (approved):
Other than that, LGTM

mrcjkb:
> ```lua
> exit = {} 
> setmetatable(exit, {
>     __call=(
>         function(...) return os.exit() end
>     ),__tostring=(
>         function(...) return "To exit, type 'exit()' or press <C-d>." end)
> })
> ```
> 
> `exit` will print the message on lua5.3 and 5.4, `print(exit)` will print the message on all versions (including JIT), and `exit()` also works on all versions.

Thanks for the suggestion.
I'd rather not use the `setmetatable` hack. I had to revert it (in a previous PR) because the behaviour was inconsistent and it errored for some Lua versions.
We print how to exit in the repl's welcome message. That's good enough.

PR #817, opened 2025-06-28
feat(cli): lx shell

### LINKED ISSUE #719: [Feature] `lx shell`
Similar to `lx repl`, we could add a `lx shell` command, which sets up the `PATH`, `LUA_PATH` and `LUA_CPATH` environment and enters a shell (which we may want to infer from the parent shell if possible, falling back to bash on linux, zsh on mac and cmd.exe on windows).

Feelfeel20088:
This is by no means done yet this is more of a template so i can confirm what i should do. since this is just a shell with lua dependency in and is not really executing a lua program, LUA_INIT wont work here and we may have to use something else like a --rcfile to show the welcome message and how to exit. 

Closes #719

### DIFF
--- a/lux-cli/src/lib.rs
+++ b/lux-cli/src/lib.rs
@@ -70,14 +70,11 @@ pub mod which;
 
 /// A luxurious package manager for Lua.
 #[derive(Parser)]
-#[command(author, about, long_about = None, arg_required_else_help = true)]
+#[command(author, version, about, long_about = None, arg_required_else_help = true)]
 pub struct Cli {
     /// Enable the sub-repositories in luarocks servers forrockspecs of in-development versions.
     #[arg(long)]
     pub dev: bool,
-
-    #[arg(short = 'v', long = "version")]
-    pub version: bool,
 
     /// Fetch rocks/rockspecs from this server (takes priority over config file).
     #[arg(long, value_name = "server")]
@@ -119,7 +116,7 @@ pub struct Cli {
 
     /// Override config variables.{n}
     /// Example: `lx -v "LUA=/path/to/lua" ...`
-    #[arg(long, value_name = "variable", value_parser = parse_key_val::<String, String>)]
+    #[arg(long, value_name = "variable", visible_short_aliases = ['v'], value_parser = parse_key_val::<String, String>)]
     pub variables: Option<Vec<(String, String)>>,
 
     /// Display verbose output of commands executed.
--- a/lux-cli/src/lib.rs
+++ b/lux-cli/src/lib.rs
@@ -25,6 +25,7 @@ use remove::Remove;
 use run::Run;
 use run_lua::RunLua;
 use search::Search;
+use shell::Shell;
 use test::Test;
 use uninstall::Uninstall;
 use update::Update;
@@ -58,6 +59,7 @@ pub mod remove;
 pub mod run;
 pub mod run_lua;
 pub mod search;
+pub mod shell;
 pub mod test;
 pub mod uninstall;
 pub mod unpack;
@@ -289,6 +291,8 @@ pub enum Commands {
     Upload(Upload),
     /// Tell which file corresponds to a given module name.
     Which(Which),
+    /// Makes a subshell that sets up PATH, LUA_PATH and LUA_CPATH.
+    Shell(Shell),
 }
 
 /// Parse a key=value pair.
--- a/lux-cli/src/lib.rs
+++ b/lux-cli/src/lib.rs
@@ -288,7 +288,7 @@ pub enum Commands {
     Upload(Upload),
     /// Tell which file corresponds to a given module name.
     Which(Which),
-    /// Makes a subshell that sets up PATH, LUA_PATH and LUA_CPATH.
+    /// Spawns an interactive shell with PATH, LUA_PATH, LUA_CPATH and LUA_INIT set.
     Shell(Shell),
 }
 
--- a/lux-cli/src/lib.rs
+++ b/lux-cli/src/lib.rs
@@ -68,11 +68,14 @@ pub mod which;
 
 /// A luxurious package manager for Lua.
 #[derive(Parser)]
-#[command(author, version, about, long_about = None, arg_required_else_help = true)]
+#[command(author, about, long_about = None, arg_required_else_help = true)]
 pub struct Cli {
     /// Enable the sub-repositories in luarocks servers forrockspecs of in-development versions.
     #[arg(long)]
     pub dev: bool,
+
+    #[arg(short = 'v', long = "version")]
+    pub version: bool,
 
     /// Fetch rocks/rockspecs from this server (takes priority over config file).
     #[arg(long, value_name = "server")]
@@ -114,7 +117,7 @@ pub struct Cli {
 
     /// Override config variables.{n}
     /// Example: `lx -v "LUA=/path/to/lua" ...`
-    #[arg(long, value_name = "variable", visible_short_aliases = ['v'], value_parser = parse_key_val::<String, String>)]
+    #[arg(long, value_name = "variable", value_parser = parse_key_val::<String, String>)]
     pub variables: Option<Vec<(String, String)>>,
 
     /// Display verbose output of commands executed.
--- a/lux-cli/src/shell.rs
+++ b/lux-cli/src/shell.rs
@@ -1,12 +1,13 @@
 use clap::Args;
 use eyre::{eyre, Result, WrapErr};
-use lux_lib::{config::Config, path::Paths};
+use lux_lib::{config::Config, path::PackagePath, path::Paths};
 use which::which;
 
 use std::{env, path::PathBuf};
 use tokio::process::Command;
 
 use std::process::Stdio;
+use std::str::FromStr;
 
 use super::utils::project::current_project_or_user_tree;
 
@@ -64,6 +65,14 @@ pub async fn shell(data: Shell, config: Config) -> Result<()> {
         path.prepend(&build_path);
     }
 
+    let mut lua_path = PackagePath::from_str(env::var("LUA_PATH").unwrap_or_default().as_str())
+        .unwrap_or_default();
+    lua_path.prepend(path.package_path());
+
+    let lua_cpath = PackagePath::from_str(env::var("LUA_CPATH").unwrap_or_default().as_str())
+        .unwrap_or_default();
+    lua_path.prepend(path.package_cpath());
+
     let lua_init = if data.no_loader {
         None
     } else if tree.version().lux_lib_dir().is_none() {
@@ -80,8 +89,8 @@ pub async fn shell(data: Shell, config: Config) -> Result<()> {
 
     let _ = Command::new(&shell)
         .env("PATH", path.path_prepended().joined())
-        .env("LUA_PATH", path.package_path().joined())
-        .env("LUA_CPATH", path.package_cpath().joined())
+        .env("LUA_PATH", lua_path.joined())
+        .env("LUA_CPATH", lua_cpath.joined())
         .env("LUA_INIT", lua_init.unwrap_or_default())
         .env("LUX_SHELL", "")
         .stdin(Stdio::inherit())
--- a/lux-cli/src/shell.rs
+++ b/lux-cli/src/shell.rs
@@ -12,15 +12,17 @@ use super::utils::project::current_project_or_user_tree;
 
 #[derive(Args)]
 pub struct Shell {
-    /// Whether to load test dependencies into the new shell
+    /// Add test dependencies to the shell's paths
     #[arg(long)]
     test: bool,
 
-    /// Whether to load build dependencies into the new shell
+    /// Add build dependencies to the shell's paths
     #[arg(long)]
     build: bool,
 
-    /// Suppresses the warning for checking if the lux-lua lib exists
+    /// Disable the Lux loader.
+    /// If a rock has conflicting transitive dependencies,
+    /// disabling the Lux loader may result in the wrong modules being loaded.
     #[arg(long)]
     no_loader: bool,
 }
--- a/lux-cli/src/shell.rs
+++ b/lux-cli/src/shell.rs
@@ -60,18 +60,26 @@ pub async fn shell(data: Shell, config: Config) -> Result<()> {
         let build_path = Paths::new(&build_tree_path)?;
         path.prepend(&build_path);
     }
-    if !data.no_loader {
-        if tree.version().lux_lib_dir().is_none() {
-            eprintln!("⚠️ WARNING: lux-lua library not found.\nCannot use the `lux.loader`.");
-            eprintln!("To suppress this warning, set the --no-loader option.");
-            std::process::exit(1);
-        }
-    }
+
+    let lua_init = if data.no_loader.unwrap_or(false) {
+        None
+    } else if tree.version().lux_lib_dir().is_none() {
+        eprintln!(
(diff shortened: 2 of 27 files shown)


cootshk:
> This is by no means done yet this is more of a template so i can confirm what i should do. since this is just a shell with lua dependency in and is not really executing a lua program, LUA_INIT wont work here and we may have to use something else like a --rcfile to show the welcome message and how to exit.
> 
> Closes #719

if it's not done, mark it as a draft so that CI/CD won't run (top right, under the reviewers section)

Feelfeel20088:
@mrcjkb Thanks for marking it for me. should I make a custom rc file for the shell? This is how other shells work like nix shell in think. Also where should I put the file.

mrcjkb:
> should I make a custom rc file for the shell?

No, there's no need for an rc file. The shell should just drop you into the same shell as your parent shell, with the environment variables set.

Feelfeel20088:
> > should I make a custom rc file for the shell?
> 
> 
> 
> No, there's no need for an rc file. The shell should just drop you into the same shell as your parent shell[^1], with the environment variables set.
> 
> 
> 
> [^1]: If the parent shell is unknown, we can default to `bash` on Linux, `zsh` on macOS and `cmd.exe` on Windows.

Alright thanks. I thought we were doing something like nix shell. If you look in shell.rs I have already made the back ups if the parent shell is not found. Should I display something before the new shell starts?

mrcjkb:
> I thought we were doing something like nix shell

I'd say YAGNI - we can always do that later if the need arises.

> Should I display something before the new shell starts?

No need for now. I suppose we'll want to display the same warning we display in `lx lua` if `lux-lua` can't be found.

lux-cli/src/shell.rs
```diff
@@ -0,0 +1,72 @@
...
+
+    let mut path = Paths::new(&project_tree)?;
+
+    let shell = env::var("SHELL").unwrap_or_else(|_| {
+        #[cfg(target_os = "linux")]
+        return "/bin/bash".to_string();
+        #[cfg(target_os = "windows")]
+        return "cmd.exe".to_string();
+        #[cfg(target_os = "macos")]
+        return "/bin/zsh".to_string();
```
mrcjkb: issue: not all operating systems have a `/bin` directory.
suggestion: use the [`which` crate](https://docs.rs/which/latest/which/) to detect the location of the binary. If not available, return an `Err`.

lux-cli/src/shell.rs
```diff
@@ -0,0 +1,72 @@
...
+
+    if data.build {
+        let build_tree_path = project_tree.build_tree(&config)?;
+        let build_path = Paths::new(&build_tree_path)?;
+        path.prepend(&build_path);
+    }
+
+    if project_tree.version().lux_lib_dir().is_none() {
+        eprintln!("⚠️ WARNING: lux-lua library not found.\nCannot use the `lux.loader`.");
+        std::process::exit(1);
```
mrcjkb: todo: add a `--no-loader` flag to the `lx shell` command and add a hint to the warning message: `To suppress this warning, set the --no-loader option.`
(See the other warning messages).

lux-cli/src/lib.rs
```diff
@@ -286,6 +288,8 @@ pub enum Commands {
     Upload(Upload),
     /// Tell which file corresponds to a given module name.
     Which(Which),
+    /// Makes a subshell that sets up PATH, LUA_PATH and LUA_CPATH.
```
mrcjkb:
```suggestion
    /// Spawns an interactive shell with PATH, LUA_PATH, LUA_CPATH and LUA_INIT set.
```

lux-cli/src/shell.rs
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    let _ = Command::new(&shell)
```
mrcjkb, line 90: todo: We also need to set `LUA_INIT` (see the implementation `lx lua` for example).

lux-cli/src/shell.rs
```diff
@@ -0,0 +1,72 @@
...
+        .env(
+            "PATH",
+            format!(
+                "{}:{}",
```
mrcjkb: issue: This is not platform independent.
Use `BinPath::path_prepended` instead.

lux-cli/src/shell.rs
```diff
@@ -0,0 +1,72 @@
...
+use lux_lib::{path::Paths, project::Project};
+
+use std::process::Stdio;
+
+#[derive(Args)]
+pub struct Shell {
+    #[arg(long)]
+    test: bool,
+    #[arg(long)]
+    build: bool,
```
mrcjkb: todo: These flags need documentation.

lux-cli/src/shell.rs
```diff
@@ -0,0 +1,72 @@
...
+}
+
+pub async fn shell(data: Shell, config: Config) -> Result<()> {
+    let project =
+        Project::current()?.ok_or_else(|| eyre::eyre!("Not in a Lux project directory"))?;
+    let project_tree = project.tree(&config)?;
```
mrcjkb: todo: `lx shell` should also work if there's no project in the cwd.
Use `utils::current_project_or_user_tree` instead.

mrcjkb (changes requested):

Feelfeel20088:
> Thanks for the PR.
> 
> @Feelfeel20088 have you tested this?
> 
> I just tried it using Nushell and Zsh on NixOS and it didn't work.
> 
> The command didn't drop me into an interactive shell.

I did test it. I think my shell environment var was set so it worked for me.

lux-cli/src/shell.rs
mrcjkb: issue: We don't want to exit here (and we never just `std::process::exit` in Lux - that makes things hard to troubleshoot).

See [my other comment](https://github.com/nvim-neorocks/lux/pull/817#discussion_r2176724895) about setting `LUA_INIT`.

In this case, we want something like we do in the `lux-lib::operations::exec` module:

https://github.com/nvim-neorocks/lux/blob/ddc38640486f0562ee3d9ea5aa972902b465b6e5/lux-lib/src/operations/exec.rs#L99-L111

You can copy/paste that for now. I will do a refactor to extract duplicate code after merging.


note: This was why it was failing when I tested it.

lux-cli/src/shell.rs
mrcjkb: issue: Why did you mark this as resolved?
There's still no documentation.

We should always add comments documenting what the flags mean.
In this case:

```rust
    /// Add test dependencies to the `LUA_PATH` and `LUA_CPATH.
    #[arg(long)]
    test: bool,

    /// Add build dependencies to the `LUA_PATH` and `LUA_CPATH.
    #[arg(long)]
    build: bool,
```

lux-cli/src/shell.rs
mrcjkb: issue: Why did you mark this as resolved?

lux-cli/src/shell.rs
mrcjkb: issue: Why did you mark this as resolved?

lux-cli/src/shell.rs
mrcjkb: `path.path()` returns a `BinPath`, which has a `path_prepended` method. Please use that.

lux-cli/src/shell.rs
mrcjkb: See https://github.com/nvim-neorocks/lux/pull/817#discussion_r2178501682

lux-cli/src/shell.rs
Feelfeel20088: as im making the changes im marking them as resolved. am i supposed to only mark these after i fix and commit it? sorry.

lux-cli/src/shell.rs
Feelfeel20088: but this is not exiting anyways. how should i gracefully exit?

Feelfeel20088:
i think its building a prior version

lux-cli/src/shell.rs
mrcjkb: Yes.

lux-cli/src/shell.rs
mrcjkb: There is no need to exit here. We're only printing a warning message. And if `lua_init` is `None`, we simply don't set `LUA_INIT`.

Feelfeel20088: got it ty.

lux-cli/src/shell.rs
```diff
@@ -0,0 +1,91 @@
...
+        None
+    } else {
+        Some(path.init())
+    };
+
+    let _ = Command::new(&shell)
+        .env("PATH", path.path_prepended().joined())
+        .env("LUA_PATH", path.package_path().joined())
+        .env("LUA_CPATH", path.package_cpath().joined())
+        .env("LUA_INIT", lua_init)
```
mrcjkb:
```suggestion
        .env("LUA_INIT", lua_init.unwrap_or_default())
```

lux-cli/src/shell.rs
```diff
@@ -0,0 +1,91 @@
...
+    let tree = current_project_or_user_tree(&config).unwrap();
+
+    let mut path = Paths::new(&tree)?;
+
+    let shell: PathBuf = match env::var("SHELL") {
+        Ok(val) => PathBuf::from(val),
+        Err(_) => {
+            #[cfg(target_os = "linux")]
+            let fallback = which("bash")
+                .map_err(|_| eyre::eyre!("Cannot find shell `bash` on your system!"))?;
```
mrcjkb:
```suggestion
                .wrap_err("Cannot find `bash` on your system!")?;
```

(requires an import of the `eyre::Context` trait)

Ditto for the other `which` calls.

lux-cli/src/shell.rs
```diff
@@ -0,0 +1,91 @@
...
+        None
+    } else {
+        Some(path.init())
+    };
+
+    let _ = Command::new(&shell)
+        .env("PATH", path.path_prepended().joined())
+        .env("LUA_PATH", path.package_path().joined())
+        .env("LUA_CPATH", path.package_cpath().joined())
+        .env("LUA_INIT", lua_init.unwrap_or_else(|| "".to_string()))
```
mrcjkb:
```suggestion
        .env("LUA_INIT", lua_init.unwrap_or_default())
```

Feelfeel20088: good catch

Feelfeel20088:
i will remember next time to commit changes before resolving a issue. Thanks for your patience with this.

Feelfeel20088 (commented):
looks good. whats the requested change?

lux-cli/src/shell.rs
```diff
@@ -0,0 +1,88 @@
...
+pub struct Shell {
+    /// Whether to load test dependencies into the new shell
+    #[arg(long)]
+    test: bool,
+
+    /// Whether to load build dependencies into the new shell
+    #[arg(long)]
+    build: bool,
+
+    /// Suppresses the warning for checking if the lux-lua lib exists
```
mrcjkb:
```suggestion
    /// Disable the Lux loader.
    /// If a rock has conflicting transitive dependencies,
    /// disabling the Lux loader may result in the wrong modules being loaded.
```

lux-cli/src/shell.rs
```diff
@@ -0,0 +1,88 @@
...
+use std::{env, path::PathBuf};
+use tokio::process::Command;
+
+use std::process::Stdio;
+
+use super::utils::project::current_project_or_user_tree;
+
+#[derive(Args)]
+pub struct Shell {
+    /// Whether to load test dependencies into the new shell
```
mrcjkb:
```suggestion
    /// Add test dependencies to the shell's paths
```

lux-cli/src/shell.rs
```diff
@@ -0,0 +1,88 @@
...
+    #[arg(long)]
+    test: bool,
+
+    /// Whether to load build dependencies into the new shell
```
mrcjkb:
```suggestion
    /// Add build dependencies to the shell's paths
```

mrcjkb:
> whats the requested change?

GitHub will keep displaying the status as "changes requested" until I review again.

mrcjkb:
One more thing just occurred to me. 
The way this is implemented right now is not idempotent: entering a `lx shell` twice will keep prepending to the `PATH`.

Perhaps we want to set a `LUX_SHELL` environment variable and add a check like this: 

```rust
if env::var("LUX_SHELL").is_ok() {
  return Err(eyre!("Already in a Lux shell."));
}
```

We should also prepend our `LUA_PATH`s and `LUA_CPATH`s:

```rust
let mut lua_path = PackagePath::from_str(env::var("LUA_PATH").unwrap_or_default().as_str()).unwrap_or_default();
lua_path.prepend(path.package_path());

let mut lua_cpath = PackagePath::from_str(env::var("LUA_CPATH").unwrap_or_default().as_str()).unwrap_or_default();
lua_path.prepend(path.package_cpath());
```

For now, there's no need for any flags to change this behaviour. Users can unset the environment variables before entering an `lx shell` if they need to.

Feelfeel20088:
I thought setting luapath and luacpath was the entire point of the shell why would we be prepending these if they haven't been set yet?

Feelfeel20088:
> One more thing just occurred to me. The way this is implemented right now is not idempotent: entering a `lx shell` twice will keep prepending to the `PATH`.
> 
> Perhaps we want to set a `LUX_SHELL` environment variable and add a check like this:
> 
> ```rust
> if env::var("LUX_SHELL").is_ok() {
>   return Err(eyre!("Already in a Lux shell."));
> }
> ```
> 
> We should also prepend our `LUA_PATH`s and `LUA_CPATH`s:
> 
> ```rust
> let mut lua_path = PackagePath::from_str(env::var("LUA_PATH").unwrap_or_default().as_str()).unwrap_or_default();
> lua_path.prepend(path.package_path());
> 
> let mut lua_cpath = PackagePath::from_str(env::var("LUA_CPATH").unwrap_or_default().as_str()).unwrap_or_default();
> lua_path.prepend(path.package_cpath());
> ```
> 
> For now, there's no need for any flags to change this behaviour. Users can unset the environment variables before entering an `lx shell` if they need to.

cant we just get these from calling something like package_path_prepended or some function like that? sorry im on my phone i cant check rn.

lux-cli/src/shell.rs
```diff
@@ -0,0 +1,104 @@
...
+
+    let _ = Command::new(&shell)
+        .env("PATH", path.path_prepended().joined())
+        .env("LUA_PATH", lua_path.joined())
+        .env("LUA_CPATH", lua_cpath.joined())
+        .env("LUA_INIT", lua_init.unwrap_or_default())
+        .env("LUX_SHELL", "")
+        .stdin(Stdio::inherit())
+        .stdout(Stdio::inherit())
+        .stderr(Stdio::inherit())
```
mrcjkb, line 98: suggestion: I believe these are redundant. From the docs:

```
    /// Sets configuration for the child process's standard error (stderr) handle.
    ///
    /// Defaults to [`inherit`] when used with `spawn` or `status`, and
    /// defaults to [`piped`] when used with `output`.
```

```suggestion
```

The `Stdio` import will need to be removed, too.

mrcjkb:
> cant we just get these from calling something like package_path_prepended or some function like that?

No, I don't think we have an API for that.

mrcjkb:
Note for the future @Feelfeel20088: Please use conventional commits (see CONTRIBUTING.md) for the commit messages and PR title.

PR #837, opened 2025-07-08
feat: expose shell completions in main binary

rcasia:
Hi! 👋 

As an user that installed lux via cargo install, I wanted to have an option to generate autocompletions right from the CLI. That allows me to easily do:

```bash
lx completion zsh > ~/.zsh/completions/_lx
```

## Result

```
➜  lux git:(feature/add-shell-completions) ./target/debug/lx completion --help
Generate autocompletion scripts for the shell

Usage: lx completion [SHELL]

Arguments:
  [SHELL]  The shell to generate the completion script for [default: bash] [possible values: bash, elvish, fish, powershell, zsh]

Options:
  -h, --help  Print help
```

I later noticed that there’s already some related logic in [xtask/src/main.rs](https://github.com/nvim-neorocks/lux/blob/7ca3daec6c5ea237253846abb1f06bbf85f94ceb/xtask/src/main.rs), though it seems intended for a different use case. I’m not sure if or how these should be unified, so open to feedback on that!

### DIFF
--- a/lux-cli/src/lib.rs
+++ b/lux-cli/src/lib.rs
@@ -150,7 +150,9 @@ pub enum Commands {
     Config(ConfigCmd),
     /// Generate autocompletion scripts for the shell.
     Completion {
-        /// The shell to generate the completion script for.
+        /// The shell to generate the completion script for.{n}
+        /// If not set, Lux will try to detect the current shell.{n}
+        /// Possible values: "bash", "elvish", "fish", "powershell", "zsh"{n}
         #[arg(value_enum, default_value = "bash")]
         shell: clap_complete::Shell,
     },
--- a/lux-cli/src/lib.rs
+++ b/lux-cli/src/lib.rs
@@ -36,6 +36,7 @@ use which::Which;
 pub mod add;
 pub mod build;
 pub mod check;
+pub mod completion;
 pub mod config;
 pub mod debug;
 pub mod doc;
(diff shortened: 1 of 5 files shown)


lux-cli/src/lib.rs
```diff
@@ -147,6 +148,12 @@ pub enum Commands {
     /// Interact with the lux configuration.
     #[command(subcommand, arg_required_else_help = true)]
     Config(ConfigCmd),
+    /// Generate autocompletion scripts for the shell.
+    Completion {
+        /// The shell to generate the completion script for.
```
mrcjkb: suggestion:

- More detailed documentation.
- Make this optional and try to detect the current shell if not set
  (e.g. using `std::env::get_var("SHELL")`) - I can also do this later in a follow-up PR if you'd prefer.

```suggestion
        /// The shell to generate the completion script for.{n}
        /// If not set, Lux will try to detect the current shell.{n}
        /// Possible values: "bash", "elvish", "fish", "powershell", "zsh"{n}
```

lux-cli/src/lib.rs
```diff
@@ -147,6 +148,12 @@ pub enum Commands {
     /// Interact with the lux configuration.
     #[command(subcommand, arg_required_else_help = true)]
     Config(ConfigCmd),
+    /// Generate autocompletion scripts for the shell.
```
mrcjkb: suggestion: add an example to the documentation.

```suggestion
    /// Generate autocompletion scripts for the shell.{n}
    /// Example: `lx completion zsh > ~/.zsh/completions/_lx`
```

mrcjkb (changes requested):
Thanks :pray: 

LGTM. Just some comments on the documentation.

> I later noticed that there’s already some related logic in [xtask/src/main.rs](https://github.com/nvim-neorocks/lux/blob/7ca3daec6c5ea237253846abb1f06bbf85f94ceb/xtask/src/main.rs), though it seems intended for a different use case. I’m not sure if or how these should be unified, so open to feedback on that!

~I can look into that later.~ Edit: There related logic is only a few LOC. Not worth unifying.

mrcjkb (approved):

rcasia:
Apologies for not catching the comments yesterday. Thanks a lot for following up and addressing them!

Issue #63, opened 2024-09-14
autogenerate .luarc for dependencies


max397574:
`rocks new` should generate a .luarc which can be picked up by lua-ls to recognize the dependencies in `.rocks`

mrcjkb:
we might also want a command to update it if the dependencies change.

PR #852, opened 2025-07-11
feat: autogenerate .luarc.json

rcasia:
Solves #63

- [x] ~Read directories such `./lux/<lua-version>/*/src`~
- [x] use the ProjectLockfile (dependencies and test_dependencies) as the source of truth for constructing the workspace.
- [x] Keep lua version into account
- [x] Generate file if not exists
- [x] Sort libraries
- [x] Keep preexisting json fields
- [x] add .luarc generation to the operations::sync module
- [x] check it does generate on `lx add`
- [x] check it does generate on `lx remove`
- [x] check it does generate on `lx build`
- [x] check it does generate on `lx test`
- [x] There should be a Config option to enable this feature. It can be enabled by default for now

### DIFF
--- a/lux-cli/src/lib.rs
+++ b/lux-cli/src/lib.rs
@@ -134,6 +134,10 @@ pub struct Cli {
     #[arg(long, value_name = "seconds")]
     pub timeout: Option<usize>,
 
+    /// Whether to generate or update a `.luarc.json` file for the project.
+    #[arg(long)]
+    pub generate_luarc: bool,
+
     #[command(subcommand)]
     pub command: Commands,
 }
--- a/lux-lib/src/luarc/mod.rs
+++ b/lux-lib/src/luarc/mod.rs
@@ -31,12 +31,23 @@ pub fn update_luarc(config: &Config) -> Result<(), ()> {
     }
     let project = Project::current_or_err().expect("failed to get current project");
     let tree = project.tree(&config).expect("failed to get project tree");
-    let luarc_path = project.luarc_path();
-
-    let luarc_content = fs::read_to_string(&luarc_path).unwrap_or_else(|_| String::from("{}"));
     let lockfile = project.lockfile().expect("should have a lockfile");
+    let luarc_path = project.luarc_path();
+    let relative_tree_root_path = tree
+        .root()
+        .strip_prefix(&project.root())
+        .expect("tree root should be a subpath of project root")
+        .to_path_buf();
 
-    let dependency_dirs = find_dependency_dirs(&lockfile, tree.root());
+    // read the existing .luarc file or create a new one if it doesn't exist
+    let luarc_content = fs::read_to_string(&luarc_path).unwrap_or_else(|_| String::from("{}"));
+
+    let dependency_dirs = find_dependency_dirs(&lockfile, relative_tree_root_path)
+        .into_iter()
+        // make sure the paths actually exist
+        .filter(|path| fs::exists(path).is_ok_and(|exists| exists))
+        .collect();
+
     let file = generate_luarc(luarc_content.as_str(), dependency_dirs);
 
     fs::write(&luarc_path, file)
@@ -55,7 +66,7 @@ fn find_dependency_dirs(
 
     let directories: Vec<PathBuf> = rocks
         .iter()
-        .map(|t| lux_tree_base_dir.join(format!("/{}-{}@{}/src", t.0, t.1.name(), t.1.version())))
+        .map(|t| lux_tree_base_dir.join(format!("{}-{}@{}/src", t.0, t.1.name(), t.1.version())))
         .collect();
 
     let test_rocks = lockfile.local_pkg_lock(&LocalPackageLockType::Test).rocks();
@@ -64,7 +75,7 @@ fn find_dependency_dirs(
         .iter()
         .map(|t| {
             lux_tree_base_dir.join(format!(
-                "/test-dependencies/{}-{}@{}/src",
+                "test-dependencies/{}-{}@{}/src",
                 t.0,
                 t.1.name(),
                 t.1.version()
--- a/lux-lib/src/luarc/mod.rs
+++ b/lux-lib/src/luarc/mod.rs
@@ -24,26 +24,26 @@ struct Workspace {
     library: Vec<String>,
 }
 
-pub fn update_luarc(config: &Config) -> Result<(), ()> {
+pub fn update_luarc(config: &Config) {
     if !config.generate_luarc() {
-        return Ok(());
+        return; // do nothing
     }
     let project = Project::current_or_err().expect("failed to get current project");
     let lockfile = project.lockfile().expect("should have a lockfile");
     let luarc_path = project.luarc_path();
-    let dependency_tree = project.tree(&config).expect("failed to get project tree");
+    let dependency_tree = project.tree(config).expect("failed to get project tree");
     let dependency_tree_root_relative_path = dependency_tree
         .root()
-        .strip_prefix(&project.root())
+        .strip_prefix(project.root())
         .expect("tree root should be a subpath of project root")
         .to_path_buf();
 
     let test_dependency_tree = project
-        .test_tree(&config)
+        .test_tree(config)
         .expect("failed to get project test tree");
     let test_dependency_tree_root_relative_path = test_dependency_tree
         .root()
-        .strip_prefix(&project.root())
+        .strip_prefix(project.root())
         .expect("test tree root should be a subpath of project root")
         .to_path_buf();
 
@@ -64,7 +64,7 @@ pub fn update_luarc(config: &Config) -> Result<(), ()> {
 
     let all_dependecy_dirs: Vec<PathBuf> = dependency_dirs
         .into_iter()
-        .chain(test_dependency_dirs.into_iter())
+        .chain(test_dependency_dirs)
         // make sure the paths actually exist
         .filter(|path| fs::exists(path).is_ok_and(|exists| exists))
         .collect();
@@ -72,9 +72,7 @@ pub fn update_luarc(config: &Config) -> Result<(), ()> {
     let file = generate_luarc(luarc_content.as_str(), all_dependecy_dirs);
 
     fs::write(&luarc_path, file)
-        .expect(format!("failed to write {} file", luarc_path.display()).as_str());
-
-    Ok(())
+        .unwrap_or_else(|_| panic!("failed to write {} file", luarc_path.display()));
 }
 
 fn find_dependency_dirs(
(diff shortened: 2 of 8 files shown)


mrcjkb:
Thanks for the PR 🙏

Some pointers: 

- [ ] We shouldn't read directories in `.lux`. Instead, we can use the `ProjectLockfile` (`dependencies` and `test_dependencies`) as the source of truth for constructing the workspace. 
- [ ] There should be a `Config` option to enable this feature. It can be enabled by default for now.
- [ ] We should also generate on 
  - `lx test`, as that's when test dependencies are added to the lockfile 
- [ ] Perhaps it makes sense to just add `.luarc` generation to the `operations::sync` module?

rcasia:
>  There should be a Config option to enable this feature. It can be enabled by default for now.

I just added a `--generate-luarc` flag for the config. But as it is true by default there is no way to opt out. Not sure if I should add another flag for that.

rcasia:
PR is ready for review. I’m unsure if the current error handling is sufficient, so open to suggestions there. Let me know if you see anything else that could be improved.

rcasia:
> I’m unsure if the current error handling is sufficient

After reviewing the test failures, I think the simplest approach is to propagate the errors and print to stderr explaining why the .luarc.json couldn’t be generated. I’ll keep working on it tomorrow, but let me know if you see a better alternative in the meantime.

lux-cli/src/lib.rs
```diff
@@ -134,6 +134,10 @@ pub struct Cli {
     #[arg(long, value_name = "seconds")]
     pub timeout: Option<usize>,
 
+    /// Whether to generate or update a `.luarc.json` file for the project.
+    #[arg(long)]
+    pub generate_luarc: bool,
+
```
mrcjkb, line 140: issue:

This results in a `--generate-luarc` CLI flag that *enbales* the `generate_luarc` config, which means it is impossible to *disable* it from the CLI.
We should keep `generate_luarc` config option, but the CLI flag should be something like `--no-luarc`. And we should invert it when passing it to the `ConfigBuilder`.

```rust
    /// Do not generate or update a `.luarc.json` file when building{n}
    /// a project.
    #[arg(long)]
    pub no_luarc: bool,
```

Then, in `lux-cli/src/bin/lx.rs`:

```rust
    let mut config_builder = ConfigBuilder::new()
        .unwrap()
        .dev(Some(cli.dev))
        .lua_dir(cli.lua_dir)
        .lua_version(cli.lua_version)
        // ...
        .generate_luarc(Some(!cli.no_luarc)) // <- Add this
```

(sorry, this is on me - I had suggested for the config option to default to `true` without clarifying the CLI option).

lux-lib/src/luarc/mod.rs
```diff
@@ -0,0 +1,179 @@
...
+
+    #[serde(default)]
+    workspace: Workspace,
+}
+
+#[derive(Serialize, Deserialize, Default, PartialEq, Debug)]
+struct Workspace {
+    #[serde(default)]
+    library: Vec<String>,
+}
```
mrcjkb, line 25: suggestion: I believe `workspace` can also have fields other than `library`, so this should probably also have a `BTreeMap` to capture unknown keys.

```suggestion
#[derive(Serialize, Deserialize, Default, PartialEq, Debug)]
struct Workspace {
    #[serde(flatten)] // <-- capture any unknown keys here
    other: BTreeMap<String, serde_json::Value>,
    
    #[serde(default)]
    library: Vec<String>,
}
```

lux-lib/src/luarc/mod.rs
```diff
@@ -0,0 +1,179 @@
...
+
+pub fn update_luarc(config: &Config) {
+    if !config.generate_luarc() {
+        return; // do nothing
```
mrcjkb, line 29: quibble: redundant comment

```suggestion
        return;
```

lux-lib/src/luarc/mod.rs
```diff
@@ -0,0 +1,179 @@
...
+    }
+    let project = Project::current_or_err().expect("failed to get current project");
+    let lockfile = project.lockfile().expect("should have a lockfile");
+    let luarc_path = project.luarc_path();
+    let dependency_tree = project.tree(config).expect("failed to get project tree");
```
mrcjkb, line 34: issue: We avoid `expect`, `unwrap` or `panic!` in lux-lib, unless we know something can't fail, but the compiler doesn't.

In this case, we would introduce a `pub UpdateLuaRcError` enum that uses `thiserror`.
I recommend you look at examples in `lux-lib::operations` modules (e.g. `lux-lib::operations::install`).

Then this can return a `Result<(), UpdateLuaRcError`>.

note: This will also solve [your comment here](https://github.com/nvim-neorocks/lux/pull/852#issuecomment-3067423008).

lux-lib/src/luarc/mod.rs
mrcjkb, line 1: issue: This should be in `lux-lib/src/operations/luarc.rs` (it will need to be re-exported in `operations/mod.rs`.

lux-lib/src/luarc/mod.rs
```diff
@@ -0,0 +1,179 @@
...
+}
+
+#[derive(Serialize, Deserialize, Default, PartialEq, Debug)]
+struct Workspace {
+    #[serde(default)]
+    library: Vec<String>,
+}
+
+pub fn update_luarc(config: &Config) {
+    if !config.generate_luarc() {
```
mrcjkb, line 28: issue: We are refactoring the `lux-lib` codebase to use `bon::Builder`s (because it's easier to add arguments without introducing a breaking change).

You can see an example of how we use `Builder`s in the `operations::run_lua` module.

Also, the functions should be `async` and should use `tokio::fs` instead of `std::fs`.

```rust
#[derive(Builder)]
#[builder(start_fn = new, finish_fn(name = _build, vis = ""))]
pub struct UpdateLuaRc<'a> {
    config: &'a Config,
}

impl<State> UpdateLuaRcBuilder<'_, State>
where
    State: update_lua_rc_builder::State + update_lua_rc_builder::IsComplete,
{
    pub async fn update(self) -> Result<(), UpdateLuaRcError> {
        do_update_lua_rc(self._build()).await
    }
}

async fn do_update_luarc(args: UpdateLuaRc) -> Result<(), UpdateLuaRcError> {
    // Implementation in this private function
}
```

Alternately, we can just make the `update_luarc` function `pub(crate)`.
By the looks of it, it's not used by lux-cli.

```rust
pub(crate) fn update_luarc(config: &Config) {
    // ...
}
```

lux-lib/src/luarc/mod.rs
```diff
@@ -0,0 +1,179 @@
...
+        return; // do nothing
+    }
+    let project = Project::current_or_err().expect("failed to get current project");
+    let lockfile = project.lockfile().expect("should have a lockfile");
+    let luarc_path = project.luarc_path();
```
mrcjkb, line 33: suggestion: We probably want to check for both `.luarc.json` and `.emmyrc.json` and then create a `.luarc.json` if neither exists.

Both formats use `workspace.library` fields.

I think for now there's no need to add a `luarc_path` method to `Project`, as it's only used in this module and it's not a core component of a Lux project.

lux-lib/src/luarc/mod.rs
```diff
@@ -0,0 +1,179 @@
...
+fn find_dependency_dirs(
+    lockfile: &ProjectLockfile<ReadOnly>,
+    lux_tree_base_dir: PathBuf,
+    local_package_lock_type: &LocalPackageLockType,
+) -> Vec<PathBuf> {
+    let rocks = lockfile.local_pkg_lock(local_package_lock_type).rocks();
+
+    rocks
+        .iter()
+        .map(|t| lux_tree_base_dir.join(format!("{}-{}@{}/src", t.0, t.1.name(), t.1.version())))
```
mrcjkb, line 87: issue: Hand-rolling the construction of the directory is prone to breakage.
Instead, use `Tree::test_tree` to get the `Tree` for the test dependencies and use
`Tree::entrypoint_layout`[^1] or `Tree::dependency_layout`[^1] to get a `RockLayout` struct, which has a `src: PathBuf` field.

[^1]: You can use `LocalPackageLock::is_entrypoint` to determine which one to use.

lux-lib/src/luarc/mod.rs
```diff
@@ -0,0 +1,179 @@
...
+}
+
+fn generate_luarc(prev_contents: &str, extra_paths: Vec<PathBuf>) -> String {
+    let mut luarc: LuaRC = serde_json::from_str(prev_contents).unwrap();
+
+    // remove any preexisting lux library paths
+    luarc
+        .workspace
+        .library
+        .retain(|path| !path.starts_with(".lux/"));
```
mrcjkb, line 98: todo: use `Project::default_tree_root_dir` instead of `".lux`.

lux-lib/src/luarc/mod.rs
```diff
@@ -0,0 +1,179 @@
...
+
+    for p in extra_paths {
+        let path = p.clone().into_os_string().into_string();
```
mrcjkb, line 101: issue: Use `to_slash_lossy` (requires a `use path_slash::PathBufExt;`).

Lua has issues with Windows path separators. `to_slash_lossy` ensures that we use unix style path separators.

lux-lib/src/luarc/mod.rs
```diff
@@ -0,0 +1,179 @@
...
+        if let Ok(path_str) = path {
+            luarc.workspace.library.push(path_str);
+        }
+    }
+
+    luarc.workspace.library.sort();
```
mrcjkb, line 107: quibble: I'm not sure if we want to sort all paths here.
Some people might be annoyed by that. Perhaps it's best to only sort the `extra_paths` before adding them to the `workspace.library`?

lux-lib/src/operations/sync.rs
```diff
@@ -267,6 +268,8 @@ async fn do_sync(
         project_lockfile.sync(dest_lockfile.local_pkg_lock(), lock_type);
     }
 
+    luarc::update_luarc(args.config);
+
```
mrcjkb, line 286: todo: We should probably add some snapshot tests.

If you prefer, I can do this as a follow-up after merging - let me know :smile:

lux-lib/src/operations/sync.rs
```diff
@@ -267,6 +268,8 @@ async fn do_sync(
         project_lockfile.sync(dest_lockfile.local_pkg_lock(), lock_type);
     }
 
+    luarc::update_luarc(args.config);
```
mrcjkb, line 285: todo: move this to the `sync_dependencies` and `sync_test_dependencies` methods. We don't need to update the luarc when syncing build dependencies.

mrcjkb (changes requested):
praise: Great work :rocket:

PR #918, opened 2025-07-24
fix(build): support multiline scripts in `command` backend

### LINKED ISSUE #914: [bug] `command` build backend does not support multiline strings
`lux.toml`:

```toml
[build]
type = "command"
build_command = """
ls
ls
"""
```

```
status: exit status: 2
stdout:
stderr: ls: cannot access 'ls': No such file or directory
```

mrcjkb:
Fixes #914


mrcjkb:
Doesn't seem to be working on msvc. Could be the unix commands though. Disabling the test on msvc for now.

PR #958, opened 2025-08-07
fix(build): copy `install.lib` files instead of compiling

### LINKED ISSUE #957: [bug] incorrect behavior when handling `build.install.lib` from rockspec
The rockspec format says this about the `build.install` field:

```
For packages which don't provide means to install modules and expect the user to copy the .lua or library files by hand to the proper locations. This table contains categories of files. Each category is itself a table, where the array part is a list of filenames to be copied. For module directories only, in the hash part, other keys are identifiers in Lua module format, to indicate which subdirectory the file should be copied to. For example, build.install.lua = {["foo.bar"] = "src/bar.lua"} will copy src/bar.lua to the foo directory under the rock's Lua files directory.
```

My understanding is that things under `build.install` should not be compiled, but instead merely copied to the package's directory. This behavior is implemented here: https://github.com/luarocks/luarocks/blob/99c57c8b2464550d6659cce43f84db83b17c4c15/src/luarocks/build/builtin.lua#L386-L394

In my library `luyoga` I need to copy the original library's header files, as well as the compiled shared libraries files, and I do so successfully with the following rockspec:

<details>
<summary>luyoga-1.3-3.rockspec</summary>

```
rockspec_format = "3.0"
package = "Luyoga"
version = "1.3-3"
source = {
   url = "git://github.com/SkyLeite/luyoga",
   tag = "v1.3-3"
}
description = {
   summary = "Lua bindings for facebook/yoga, a render-agnostic layouting engine",
   detailed = [[
      Lua bindings for the facebook/yoga layout library.
   ]],
   homepage = "https://skyleite.github.io/luyoga/",
   license = "LGPLv2.1"
}
dependencies = {
   "lua >= 5.1"
}
build = {
    type = "builtin",
    modules = {
        ["luyoga"] = "luyoga/init.lua",
        ["luyoga.layout"] = "luyoga/layout.lua",
        ["luyoga.enums"] = "luyoga/enums.lua",
        ["luyoga.node"] = "luyoga/node.lua",
        ["luyoga.style"] = "luyoga/style.lua",
        ["luyoga.util"] = "luyoga/util.lua",
        ["luyoga.value"] = "luyoga/value.lua",
    },
    install = {
      lib = {
        "dist/libyogacore.so",
        "dist/libyogacore.dylib",
        "dist/Yoga.h",
      },
    }
}
```

</details>

However, when attempting to install the library using `lx add luyoga`, I get the following error:

```
sky@Skys-MacBook-Air luyoga-test % cargo lx add luyoga
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 7.54s
     Running `/Users/sky/Projects/lux/target/debug/lx add luyoga`
⚠️ WARNING: Failed to fetch source for luyoga 1.3-3: failed to clone rock source: authentication required but no callback set; class=Ssh (23); code=Auth (-16)
⚠️ Falling back to .src.rock archive
Error: syncing dependencies with the project lockfile failed.

Caused by:
    failed to build luyoga: failed to compile intermediates from C files: ToolExecError: command did not execute successfully (status code exit status: 1): env -u IPHONEOS_DEPLOYMENT_TARGET LC_ALL="C" "clang" "-O3" "-ffunction-sections" "-fdata-sections" "-fPIC" "--target=arm64-apple-macosx" "-mmacosx-version-min=11.3" "-I" "/nix/store/lxli6sbwhkvsr81mwvwy0lpvmm3w4d6w-lua-5.1.5/include" "-w" "-o" "/var/folders/6d/fdn4gm611wv1z32z_qt6nghw0000gn/T/nix-shell.Lxuj1y/Yoga.qvCxtZRmENdl/62b861da44fe2efa-Yoga.o" "-c" "/var/folders/6d/fdn4gm611wv1z32z_qt6nghw0000gn/T/nix-shell.Lxuj1y/luyoga.lFJDcCBIhoz3/luyoga/dist/Yoga.h"

Location:
    /Users/sky/Projects/lux/lux-cli/src/utils/project.rs:58:10
```

If the team agrees that this behavior is incorrect, I can go ahead and implement a fix in a PR. Just let me know :)

SkyLeite:
Closes #957

### DIFF
--- a/lux-lib/src/lua_rockspec/mod.rs
+++ b/lux-lib/src/lua_rockspec/mod.rs
@@ -1059,7 +1059,7 @@ mod tests {
         build = {\n
             type = 'make',\n
             install = {\n
-                lib = {['foo.bar'] = 'lib/bar.so'},\n
+                lib = {['foo.so'] = 'lib/bar.so'},\n
             },\n
             copy_directories = {\n
                 'plugin',\n
@@ -1086,7 +1086,7 @@ mod tests {
             .default
             .install
             .lib
-            .get(&LuaModule::from_str("foo.bar").unwrap())
+            .get("foo.so")
             .unwrap();
         assert_eq!(*foo_bar_path, PathBuf::from("lib/bar.so"));
         let copy_directories = rockspec.local.build.default.copy_directories;
@@ -1654,7 +1654,7 @@ mod tests {
 
         assert_eq!(
             rockspec.build().current_platform().install.bin,
-            HashMap::from([("wsapi".into(), PathBuf::from("src/launcher/wsapi.cgi"))])
+            HashMap::from([("wsapi.cgi".into(), PathBuf::from("src/launcher/wsapi.cgi"))])
         );
     }
 
--- a/lux-lib/tests/build.rs
+++ b/lux-lib/tests/build.rs
@@ -2,6 +2,7 @@ use std::path::PathBuf;
 
 use assert_fs::prelude::PathCopy;
 use assert_fs::TempDir;
+use lux_lib::rockspec::Rockspec;
 use lux_lib::{
     build::{Build, BuildBehaviour::Force},
     config::{ConfigBuilder, LuaVersion},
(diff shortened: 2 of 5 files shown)


SkyLeite:
The use-case I outlined in the original issue is fixed, but unfortunately it broke the following test:

```
lux-tests-nextest>     thread 'lua_rockspec::tests::regression_off_spec_install_binaries' panicked at lux-lib/src/lua_rockspec/mod.rs:1655:9:
lux-tests-nextest>     assertion `left == right` failed
lux-tests-nextest>       left: {"wsapi.cgi": "src/launcher/wsapi.cgi"}
lux-tests-nextest>      right: {"wsapi": "src/launcher/wsapi.cgi"}
```

Is it intended that we're removing the file extension from binaries as well? It doesn't sound right, but I'm not too familiar with the spec

mrcjkb:
> The use-case I outlined in the original issue is fixed, but unfortunately it broke the following test:
> 
> ```
> lux-tests-nextest>     thread 'lua_rockspec::tests::regression_off_spec_install_binaries' panicked at lux-lib/src/lua_rockspec/mod.rs:1655:9:
> lux-tests-nextest>     assertion `left == right` failed
> lux-tests-nextest>       left: {"wsapi.cgi": "src/launcher/wsapi.cgi"}
> lux-tests-nextest>      right: {"wsapi": "src/launcher/wsapi.cgi"}
> ```
> 
> Is it intended that we're removing the file extension from binaries as well? It doesn't sound right, but I'm not too familiar with the spec

That's okay, you can update the expected value in the test case.
I just ran a test and luarocks keeps the file extension.

lux-lib/src/lua_rockspec/mod.rs
```diff
@@ -1086,7 +1086,7 @@ mod tests {
             .default
             .install
             .lib
-            .get(&LuaModule::from_str("foo.bar").unwrap())
+            .get("foo.bar")
```
mrcjkb: issue: This might be confusing, as it looks like a Lua module. Let's change the field in the rockspec from `["foo.bar"]` to `["foo.so"]`.

mrcjkb:
todo: Please use conventional commits (see `CONTRIBUTING.md`).

mrcjkb (changes requested):
Clippy lints are failing

lux-lib/tests/build.rs
```diff
@@ -418,3 +419,58 @@ async fn test_multiline_command_build() {
...
+        .progress(&bar)
+        .behaviour(Force)
+        .build()
+        .await
+        .unwrap();
+
+    let mut install_path = tokio::fs::read_dir(dir.path().join("5.1")).await.unwrap();
+    let mut luyoga_path = None;
+    while let Some(entry) = install_path.next_entry().await.unwrap() {
+        dbg!(&entry);
```
mrcjkb: todo: delete debug print

mrcjkb:
note: Windows build failure is unrelated (#960).

lux-lib/tests/build.rs
SkyLeite: Weird, I thought this would get caught by clippy

lux-lib/tests/build.rs
```diff
@@ -418,3 +419,57 @@ async fn test_multiline_command_build() {
...
+        .tree(&tree)
+        .entry_type(tree::EntryType::Entrypoint)
+        .config(&config)
+        .progress(&bar)
+        .behaviour(Force)
+        .build()
+        .await
+        .unwrap();
+
+    let mut install_path = tokio::fs::read_dir(dir.path().join("5.1")).await.unwrap();
```
mrcjkb: issue: test is failing here

```
thread 'builtin_build_install_include' panicked at lux-lib/tests/build.rs:460:78:
    called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
    note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

mrcjkb: I'd suggest using `assert_fs` assertions instead.

SkyLeite: Oh right, this is assuming the test is run with Lua 5.1. Is it possible to run these integration tests locally, by any chance?

mrcjkb: Yes, there's a `cargo tt` alias for it.

PR #979, opened 2025-08-14
fix(macos): archive .app bundle

mrcjkb:


mrcjkb:
The .dmg created by cargo-packager actually contains the .app bundle, so there's no need for this.

PR #1057, opened 2025-09-11
chore(docs): rename nvim-neorocks -> lumen-oss

cootshk:
It builds in my codespace™


cootshk:
Also see https://github.com/lumen-oss/nvim-neorocks.github.io/issues/13 in the docs

NTBBloodbath:
The CI that compiles the project with Nix is failing on Git hooks because the source needs reformatting, could you take a look at it? Thank you for accommodating the organization's name changes btw 🙂

PR #1073, opened 2025-09-17
feat!: don't expose git-url-parse types

### LINKED ISSUE #1067: Remove or upgrade git-url-parse dependency
`git-url-parse-rs` has introduced a breaking change that adds a lifetime to the `GitUrl` struct: https://github.com/tjtelan/git-url-parse-rs/pull/61

This change breaks our structs that depend on it, and I don't want to have to add lifetimes to our structs just to support it.

It also exposes another issue: We shouldn't let the crate leak into our public API.

We should either remove the dependency or upgrade, remove it from our structs and defer parsing (which breaks the parse, don't validate principle 🥲)

mrcjkb:
Closes #1067.


lux-lib/src/git/url.rs
```diff
@@ -0,0 +1,85 @@
...
+    #[error("not a remote git URL: {0}")]
+    NotARemoteGitUrl(String),
+}
+
+impl FromStr for RemoteGitUrl {
+    type Err = RemoteGitUrlParseError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        // XXX: git_url_parse fails to parse the provider info if the URL doesn't end with .git
+        // TODO(mrcjkb): remove this if https://github.com/tjtelan/git-url-parse-rs/pull/70 gets merged
```
mrcjkb: - [x] depends on https://github.com/tjtelan/git-url-parse-rs/pull/70

PR #1152, opened 2025-10-30
feat: add ssh auth callback

ALameLlama:
Adds support for ssh auth callback, without my change you'll get errors like this
```
{ code: -16, klass: 23, message: "authentication required but no callback set" }
```

I can't see a nice way to see if the protocol is ssh and only add it if it's ssh.

I wasn't sure what the best way to test this so I created a quick file
```
use git2::{AutotagOption, Cred, FetchOptions, RemoteCallbacks, Repository};
use tempfile::tempdir;

fn main() {
    let url = "git+ssh://git@git.llamacorp.au:22/umbra-framework/contracts.git";

    let temp_dir = tempdir().unwrap();
    let repo = Repository::init_bare(&temp_dir).unwrap();

    let mut callbacks = RemoteCallbacks::new();
    callbacks.credentials(|_url, username_from_url, _allowed_types| {
        Cred::ssh_key_from_agent(username_from_url.unwrap_or("git"))
    });

    // Attach the callbacks
    let mut fetch_opts = FetchOptions::new();
    fetch_opts.remote_callbacks(callbacks);
    fetch_opts.download_tags(AutotagOption::All);

    let mut remote = repo.remote_anonymous(url).unwrap();
    println!("Fetching tags from {url} ...");

    if let Err(e) = remote.fetch(&[] as &[&str], Some(&mut fetch_opts), None) {
        eprintln!("etch failed: {e:?}");
        std::process::exit(1);
    }

    println!("Fetch succeeded!");
}
```

when I run `cargo tt` I get 4 errors both on master and my branch sooo yeah
```
        FAIL [   2.050s] lux-lib operations::sync::tests::regression_sync_nonexistent_lock
        FAIL [   1.994s] lux-lib operations::sync::tests::test_sync_add_rocks
        FAIL [   2.047s] lux-lib operations::sync::tests::test_sync_add_rocks_with_new_package
        FAIL [   2.021s] lux-lib operations::sync::tests::test_sync_remove_rocks
```

### DIFF
--- a/lux-lib/src/git/utils.rs
+++ b/lux-lib/src/git/utils.rs
@@ -131,8 +131,6 @@ mod tests {
     #[tokio::test]
     #[cfg(feature = "ssh-tests")]
     async fn test_latest_semver_tag_ssh_schema() {
-        println!("Skipping test until upstream git_url_parse is updated from 0.6.0");
-        return;
         let url = "ssh://github.com/lumen-oss/lux.git".parse().unwrap();
         assert!(latest_semver_tag(&url).unwrap().is_some());
     }
@@ -157,8 +155,6 @@ mod tests {
     #[tokio::test]
     #[cfg(feature = "ssh-tests")]
     async fn test_latest_commit_sha_ssh_schema() {
-        println!("Skipping test until upstream git_url_parse is updated from 0.6.0");
-        return;
         let url = "ssh://github.com/lumen-oss/lux.git".parse().unwrap();
         assert!(latest_commit_sha(&url).unwrap().is_some());
     }


mrcjkb:
Thanks for the PR :pray: 
The code looks good to me.

> I wasn't sure what the best way to this

I suppose it requires an SSH agent to be running, so we can't really add a test for it to the Lux test suite.

There is one strange thing:

I tried adding

```rust
let url = "git@github.com:lumen-oss/lux.git".parse().unwrap();
assert!(latest_commit_sha(&url).unwrap().is_some());
```

to the `test_latest_commit_sha` test case (I have an agent running locally, so I'd expect it to work).
It panics on the `.unwrap()` with `invalid or unknown remote ssh hostkey`.

To troubleshoot, I added this to the callback:

```rust
callbacks.credentials(|_url, username_from_url, _allowed_types| {
    panic!("{:?}", username_from_url)
});
```

but it was never invoked :thinking: 
So it appears that something might still be missing.

> when I run cargo tt I get 4 errors both on master and my branch sooo yeah

Those are likely because you're running tests for a different Lua version than Lua 5.1.
The tests can be a bit finicky when you run them locally. You need Lua 5.4 to compile lux-cli, but you also need Lua 5.1 on the PATH (before Lua 5.4) to run the impure tests (because plenary.nvim doesn't exist for other Lua versions.
I've updated the Nix devShells in #1153.

One more thing:

We'll probably also want to add a callback to the `fetch_options` in `lux-lib/src/operations/fetch.rs`:

https://github.com/lumen-oss/lux/blob/5436e583d0fc3c50f18e48a8aaf4597c8cf7707f/lux-lib/src/operations/fetch.rs?plain=1#L189-L196

ALameLlama:
I've swapped to using `nix develop .#lua51` and now I have all tests passing, 

I've also added 2 new test cases

 test_latest_semver_tag_ssh_user and test_latest_commit_sha_ssh_user and they both seem to pass for me without `invalid or unknown remote ssh hostkey`
 
 ```
        PASS [   1.387s] lux-lib git::utils::tests::test_latest_commit_sha_http
        PASS [   1.556s] lux-lib git::utils::tests::test_latest_semver_tag_http
        PASS [   1.450s] lux-lib::build builtin_build
        PASS [   2.499s] lux-lib::build test_build_multiple_treesitter_parsers
        PASS [   1.854s] lux-lib::build treesitter_parser_build
        PASS [   4.166s] lux-lib git::utils::tests::test_latest_semver_tag_ssh_user
        PASS [   4.269s] lux-lib git::utils::tests::test_latest_commit_sha_ssh_user
```

can you double check the output of `ssh-add -l`


Also let me know if you need me to change the env var to something like LUX_SKIP_SSH_TESTS since I'm not sure if it'll break CI runs

**Comment by ALameLlama** [comment]
I also added the same panic and got this when running the test
```
        FAIL [   1.150s] lux-lib git::utils::tests::test_latest_semver_tag_ssh_user
  stdout ───

    running 1 test
    test git::utils::tests::test_latest_semver_tag_ssh_user ... FAILED

    failures:

    failures:
        git::utils::tests::test_latest_semver_tag_ssh_user

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 136 filtered out; finished in 1.13s

  stderr ───

    thread 'git::utils::tests::test_latest_semver_tag_ssh_user' panicked at lux-lib/src/git/utils.rs:53:9:
    Some("git")
    note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

        FAIL [   1.164s] lux-lib git::utils::tests::test_latest_commit_sha_ssh_user
  stdout ───

    running 1 test
    test git::utils::tests::test_latest_commit_sha_ssh_user ... FAILED

    failures:

    failures:
        git::utils::tests::test_latest_commit_sha_ssh_user

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 136 filtered out; finished in 1.15s

  stderr ───

    thread 'git::utils::tests::test_latest_commit_sha_ssh_user' panicked at lux-lib/src/git/utils.rs:92:9:
    Some("git")
    note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
    ```

ALameLlama:
I can also add these tests but currently they get the malformed hostname, my upstream fix is merged but no new release has been tagged.
```
    #[tokio::test]
    async fn test_latest_semver_tag_ssh_schema() {
        if std::env::var("LUX_SKIP_IMPURE_TESTS").unwrap_or("0".into()) == "1" {
            println!("Skipping impure test");
            return;
        }
        println!("Skipping test until upstream git_url_parse is updated");
        return;
        let url = "ssh://github.com/lumen-oss/lux.git".parse().unwrap();
        assert!(latest_semver_tag(&url).unwrap().is_some());
    }
    
    #[tokio::test]
    async fn test_latest_commit_sha_ssh_schema() {
        if std::env::var("LUX_SKIP_IMPURE_TESTS").unwrap_or("0".into()) == "1" {
            println!("Skipping impure test");
            return;
        }
        println!("Skipping test until upstream git_url_parse is updated");
        return;
        let url = "ssh://github.com/lumen-oss/lux.git".parse().unwrap();
        assert!(latest_commit_sha(&url).unwrap().is_some());
    }
```

mrcjkb:
> can you double check the output of `ssh-add -l`

It outputs the SSH keys I have exposed via my KeePassXC SSH agent.

> Also let me know if you need me to change the env var to something like `LUX_SKIP_SSH_TESTS` since I'm not sure if it'll break CI runs

I think we should probably gate this behind a `ssh_tests` feature flag (disabled by default), since we can't be sure an agent is running.
I suppose it might be possible to spawn a custom SSH agent with the `ssh_agent_lib` crate, but I think that would be overkill for now, and it might conflict if an agent is running already.

ALameLlama:
> It outputs the SSH keys I have exposed via my KeePassXC SSH agent.

I'll be pretty busy the next few days but I can try and setup KeePassXC and see if I get the same issue. I am using 1Password and it seems to be working.

> I think we should probably gate this behind a ssh_tests feature flag (disabled by default), since we can't be sure an agent is running.

Do we have something like this in the code already so I can have a look at how I'd implement it? thanks

mrcjkb:
> I'll be pretty busy the next few days but I can try and setup KeePassXC and see if I get the same issue. I am using 1Password and it seems to be working.

I wouldn't bother. We already know that the tests cannot be reproduced reliably.

> Do we have something like this in the code already so I can have a look at how I'd implement it? thanks

Not yet. 
It should be as simple as adding

```toml
[features]
ssh-tests = []
```

to `lux-lib/Cargo.toml`, and then annotating the test with

```rust
#[tokio::test]
#[cfg(feature = "ssh-tests")]
async fn test_latest_semver_tag_ssh_schema() {
```

You may also need to add

```toml
[features]
ssh-tests = ["lux-lib/ssh-tests"]
```

to the `Cargo.toml` files in `lux-lua` and `lux-cli`.

ALameLlama:
I've added the feature and tested it with 'cargo tt --features ssh-tests' and without '--features' and it working as excepted. let me know if you want any other changes

lux-lib/src/git/utils.rs
```diff
@@ -122,15 +122,21 @@ mod tests {
...
-            return;
-        }
         let url = "git@github.com:lumen-oss/lux.git".parse().unwrap();
         assert!(latest_semver_tag(&url).unwrap().is_some());
     }
 
+    #[tokio::test]
+    #[cfg(feature = "ssh-tests")]
+    async fn test_latest_semver_tag_ssh_schema() {
+        println!("Skipping test until upstream git_url_parse is updated from 0.6.0");
```
mrcjkb: todo: let's remove this. 

We can wait for https://github.com/tjtelan/git-url-parse-rs/pull/82 before merging.

mrcjkb:
Merging, as the fix itself doesn't depend on the git_url_parse upgrade and the test doesn't run in CI

PR #1189, opened 2025-11-07
feat: Add `lx debug dependencies` subcommand

### LINKED ISSUE #1188: [feature] `lx debug` subcommand for available toolchains
Check for:

- C compiler toolchain
- make
- cmake
- cargo
- pkg-config 
- Lua probe result
- ???

For executables, we can use the `which` crate (which we use already).

Should be implemented in lux-lib, returning a `DependencyReport` that we can expose via lux-lua and print with human readable + json output in lux-cli

cnaples79:
## Summary
Implements a new `lx debug dependencies` subcommand that checks for required build tools and dependencies.

## Motivation
As requested in #1188, this provides users with a quick way to verify whether they have all the necessary dependencies installed for building Lua rocks.

### 1. New Module: `lux-lib/src/dependencies.rs`
- Added `DependencyReport` struct to hold dependency check results
- Added `DependencyStatus` struct for individual dependency information
- Implemented `generate()` method that checks for:
  - C compiler toolchain (tries cc, gcc, clang in order)
  - make
  - cmake
  - cargo
  - pkg-config
- Uses the existing `which` crate to locate executables

### 2. CLI Integration: `lux-cli/src/debug.rs`
- Added `Dependencies` variant to the `Debug` enum
- Added `OutputFormat` enum supporting human-readable and JSON output
- Added command-line argument for format selection

### 3. Handler Implementation: `lux-cli/src/dependencies.rs`
- Created `check_dependencies()` function to handle the command
- Implemented `print_human_readable()` for user-friendly output with:
  - Status indicators (✓/✗)
  - Dependency names
  - File paths when found
  - Summary showing found vs missing count
- Implemented `print_json()` for machine-readable output

### 4. Binary Integration: `lux-cli/src/bin/lx.rs`
- Added handler in the `Debug` match arm
- Added import for dependencies module

### Human-readable output (default):
```bash
$ lx debug dependencies
Dependency Check Report
=======================

✓ C compiler (cc): Found (/usr/bin/cc)
✓ make: Found (/usr/bin/make)
✓ cmake: Found (/usr/bin/cmake)
✓ cargo: Found (/home/user/.cargo/bin/cargo)
✗ pkg-config: Not found

Summary:
  Found: 4/5
  Missing: 1/5
```

### JSON output:
```bash
$ lx debug dependencies --format json
{
  "c_compiler": {
    "name": "C compiler (cc)",
    "found": true,
    "path": "/usr/bin/cc",
    "version": null
  },
  "make": {
    "name": "make",
    "found": true,
    "path": "/usr/bin/make",
    "version": null
  },
  ...
}
```

## Design Decisions

- **Used existing `which` crate**: Already in workspace dependencies, no new deps needed
- **C compiler detection**: Tries multiple common compilers (cc, gcc, clang) for better compatibility
- **Dual output format**: Human-readable for interactive use, JSON for scripting
- **Extensible structure**: Easy to add more dependencies or version detection in the future
- **Non-blocking**: All checks are independent, failure of one doesn't affect others

## Testing Notes
- Code structure follows existing patterns in the codebase
- Implementation tested locally

Fixes #1188

### DIFF
--- a/lux-cli/src/debug.rs
+++ b/lux-cli/src/debug.rs
@@ -2,7 +2,20 @@ use crate::{
     project::DebugProject,
     unpack::{Unpack, UnpackRemote},
 };
-use clap::Subcommand;
+use clap::{Args, Subcommand};
+
+#[derive(Args)]
+pub struct Dependencies {
+    /// Output format: human-readable or json
+    #[arg(long, default_value = "human")]
+    pub format: OutputFormat,
+}
+
+#[derive(Clone, clap::ValueEnum)]
+pub enum OutputFormat {
+    Human,
+    Json,
+}
 
 #[derive(Subcommand)]
 pub enum Debug {
@@ -14,4 +27,6 @@ pub enum Debug {
     UnpackRemote(UnpackRemote),
     /// View information about the current project.
     Project(DebugProject),
+    /// Check for required dependencies (C compiler, make, cmake, cargo, pkg-config).
+    Dependencies(Dependencies),
 }
--- a/lux-lib/src/dependencies.rs
+++ b/lux-lib/src/dependencies.rs
@@ -1,0 +1,72 @@
+use serde::{Deserialize, Serialize};
+use std::path::PathBuf;
+use which::which;
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct DependencyStatus {
+    pub name: String,
+    pub found: bool,
+    pub path: Option<PathBuf>,
+    pub version: Option<String>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct DependencyReport {
+    pub c_compiler: DependencyStatus,
+    pub make: DependencyStatus,
+    pub cmake: DependencyStatus,
+    pub cargo: DependencyStatus,
+    pub pkg_config: DependencyStatus,
+}
+
+impl DependencyReport {
+    pub fn generate() -> Self {
+        Self {
+            c_compiler: check_c_compiler(),
+            make: check_executable("make"),
+            cmake: check_executable("cmake"),
+            cargo: check_executable("cargo"),
+            pkg_config: check_executable("pkg-config"),
+        }
+    }
+}
+
+fn check_executable(name: &str) -> DependencyStatus {
+    match which(name) {
+        Ok(path) => DependencyStatus {
+            name: name.to_string(),
+            found: true,
+            path: Some(path),
+            version: None,
+        },
+        Err(_) => DependencyStatus {
+            name: name.to_string(),
+            found: false,
+            path: None,
+            version: None,
+        },
+    }
+}
+
+fn check_c_compiler() -> DependencyStatus {
+    // Try common C compilers in order of preference
+    let compilers = vec!["cc", "gcc", "clang"];
+
+    for compiler in compilers {
+        if let Ok(path) = which(compiler) {
+            return DependencyStatus {
+                name: format!("C compiler ({})", compiler),
+                found: true,
+                path: Some(path),
+                version: None,
+            };
+        }
+    }
+
+    DependencyStatus {
+        name: "C compiler".to_string(),
+        found: false,
+        path: None,
+        version: None,
+    }
+}
--- a/lux-cli/src/bin/lx.rs
+++ b/lux-cli/src/bin/lx.rs
@@ -5,7 +5,7 @@ use eyre::Result;
 use lux_cli::{
     add, build, check, completion, config,
     debug::Debug,
-    doc, download, exec, fetch, format, generate_rockspec, info, install, install_lua,
+    dependencies, doc, download, exec, fetch, format, generate_rockspec, info, install, install_lua,
     install_rockspec, lint, list, outdated, pack, path, pin, project, purge, remove, run, run_lua,
     search, shell, test, uninstall, unpack, update,
     upload::{self},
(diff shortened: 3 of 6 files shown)


lux-lib/src/dependencies.rs
```diff
@@ -0,0 +1,72 @@
...
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct DependencyReport {
+    pub c_compiler: DependencyStatus,
+    pub make: DependencyStatus,
+    pub cmake: DependencyStatus,
+    pub cargo: DependencyStatus,
+    pub pkg_config: DependencyStatus,
+}
```
mrcjkb, line 20: suggestion: 

1. Instead of a `found` flag, we can make invalid states irrepresentable by using `Option`.
2. Naming: 
  - `Tool` instead of `DependencyStatus`.
  - `ToolchainReport` instead of `DependencyReport`
3. As these types are part of the lux-lib API, we should avoid exposing the fields and instead expose getters.

```rust
pub struct Tool {
    name: String,
    version: Option<String>,
    path: PathBuf,
}

impl Tool {
    name(&self) -> &String {
        &self.name
    }
    // ...
}

pub struct ToolchainReport {
    c_compiler: Option<Tool>,
    make: Option<Tool>,
    cmake: Option<Tool>,
    cargo: Option<Tool>,
    pkg_config: Option<Tool>,
}

impl ToolchainReport {
    c_compiler(&self) -> Option<&Tool> {
        self.c_compiler.as_ref()
    }
    // ...
}
```

lux-lib/src/dependencies.rs
```diff
@@ -0,0 +1,72 @@
...
+            found: false,
+            path: None,
+            version: None,
+        },
+    }
+}
+
+fn check_c_compiler() -> DependencyStatus {
+    // Try common C compilers in order of preference
+    let compilers = vec!["cc", "gcc", "clang"];
```
mrcjkb, line 53: issue: We should use `cc` to detect the compiler toolchain:

```rust
cc::Build::new()
    .try_get_compiler()
    .ok()
    .map(|compiler| {
        let mut cmd = compiler.to_command();
        let path = PathBuf::from(cmd.get_program());
        let version = get_version(cmd);
        // ...
    })
```

lux-lib/src/dependencies.rs
```diff
@@ -0,0 +1,72 @@
...
+    }
+}
+
+fn check_executable(name: &str) -> DependencyStatus {
+    match which(name) {
+        Ok(path) => DependencyStatus {
+            name: name.to_string(),
+            found: true,
+            path: Some(path),
+            version: None,
```
mrcjkb, line 40: todo: The version is always `None` in this PR.

We can use `std::process::Command` to get the version:

```rust
fn try_get_version(mut cmd: Command) -> Option<String> {
    cmd.arg("--version")
        .output()
        .ok()
        .map(|output| String::from_utf8_lossy(&output.stdout).to_string())
        .and_then(|stdout| try_parse_version(&stdout))
}

fn try_parse_version(&stdout) -> Option<String> {
    // Maybe parsing the version is overkill and we can just
    // pass stdout to our `Tool` struct.
}
```

lux-cli/src/debug.rs
```diff
@@ -2,7 +2,20 @@ use crate::{
...
+    /// Output format: human-readable or json
+    #[arg(long, default_value = "human")]
+    pub format: OutputFormat,
+}
+
+#[derive(Clone, clap::ValueEnum)]
+pub enum OutputFormat {
+    Human,
+    Json,
+}
```
mrcjkb, line 18: todo: We already have a `OutputFormat` enum defined in `lux-cli/src/check.rs`.
Let's move that to a `lux-cli/src/args.rs` and reuse it.

lux-cli/src/debug.rs
```diff
@@ -14,4 +27,6 @@ pub enum Debug {
     UnpackRemote(UnpackRemote),
     /// View information about the current project.
     Project(DebugProject),
+    /// Check for required dependencies (C compiler, make, cmake, cargo, pkg-config).
```
mrcjkb, line 30:
```suggestion
    /// Check for available toolchains.
```

lux-cli/src/debug.rs
```diff
@@ -2,7 +2,20 @@ use crate::{
...
 };
-use clap::Subcommand;
+use clap::{Args, Subcommand};
+
+#[derive(Args)]
+pub struct Dependencies {
+    /// Output format: human-readable or json
+    #[arg(long, default_value = "human")]
+    pub format: OutputFormat,
+}
```
mrcjkb, line 12: todo: I initially used the word "dependencies" in the issue title, but I think `lx debug toolchains` is a better command.

Reason: The toolchains aren't strictly "dependencies" of Lux. For example, you can build most packages without Cargo. You only need Cargo if you're building a package that uses the `rust-mlua` build backend.

lux-cli/src/debug.rs
```diff
@@ -2,7 +2,20 @@ use crate::{
     project::DebugProject,
     unpack::{Unpack, UnpackRemote},
 };
-use clap::Subcommand;
+use clap::{Args, Subcommand};
+
+#[derive(Args)]
+pub struct Dependencies {
+    /// Output format: human-readable or json
```
mrcjkb, line 9: issue: No need to specify "human-readable or json" here.
Clap will generate help for the enum variants.

```suggestion
    /// The output format.
```

results in the following CLI `--help` message:

```
--output-format <OUTPUT_FORMAT>  The output format [default: text] [possible values: json, text]
help
```

mrcjkb (changes requested):
Thanks for the PR 🙏

mrcjkb:
Closing due to inactivity.

Issue #953, opened 2025-08-05
[Featue] Install LLS_Addons if using libraries that require them for checks


mrcjkb:
#849 only implements `lx check` for the `src` and `lua` directories.
It would be nice if it would also check `spec`, `test` and `tests` directories.
Some binary rocks (e.g. libuv) have types defined in [LLS_Addons](https://github.com/LuaLS/LLS-Addons).

`lx check` could check the project dependencies and test dependencies, and if there exists an LLS add-on for it, install the LLS-Addons to a dedicated location in the `.lux` tree.

We should probably package LLS-Addons for luarocks. Someone seems to have had the same idea: https://github.com/LuaLS/luarocks-build-addon

Perhaps we should add a `check_dependencies` field to the `lux.toml` format.

keithharvey:
We merged Lux into BAR and it's working well for lint and test. One gap: no type-checking coverage for test code -- busted/luassert globals are invisible to the language server without the LuaCATS addons.

We tried vendoring the [LuaCATS/busted](https://github.com/LuaCATS/busted) and [LuaCATS/luassert](https://github.com/LuaCATS/luassert) repos into the project but that wasn't well received. Any movement on the luarocks-packaging approach you mentioned, or a direction I could help with?

The other gap is our unconventional folder structure and lack of a src/ folder thanks to RecoilEngine constraints, so being able to customize the paths for some of these commands like `lx fmt` would be nifty until we get a chance to subdirectory our game directories over in Recoil to conform with lua package norms.

mrcjkb:
@keithharvey you can follow the progress of packaging LuaLS add-ons for luarocks in the [luarocks-build-addon](https://github.com/LuaLS/luarocks-build-addon) repository and by looking at their [manifest on luarocks.org](https://luarocks.org/m/lls-addons).
It looks like they haven't packaged the busted or luassert add-ons yet.

I've opened some issues for the ability to specify directories for various commands:

- #1407
- #1408
- #1409

I probably won't be able to work on them immediately, but PRs are welcome 😄.

Regarding my initial idea of adding a `check_dependencies` field to `lux.toml`: That probably won't be necessary, as it should be possible to use the existing `test_dependencies` field.

PR #1190, opened 2025-11-07
feat: luarc dependency overrides

keithharvey:
### PR Summary

This change works as intended for me, but please confirm whether this approach feels necessary.
I’m happy to add specs or update the docs if we decide to merge it.

---

### Context

```toml
[test_dependencies.busted_intellisense]
git = "github:LuaCATS/busted"
version = "5ed85d0e016a5eb5eca097aa52905eedf1b180f1"
```

Currently produces a `luarc.json` entry hard-coded to **src**:

```json
".lux/5.1/test_dependencies/5.1/3b7...5-luassert_intellisense@d352...9-1/src"
```

However, this directory is empty for LuaCATS repositories, which keep their type definitions in `/library`. In our luarocks stub structure, that path is categorized as **etc** (`...9-1/etc`).

---

### Solution

This adds optional support for specifying the `.luarc` subdirectory in `lux.toml`:

```toml
[test_dependencies.busted_intellisense]
git = "github:LuaCATS/busted"
version = "5ed85d0e016a5eb5eca097aa52905eedf1b180f1"
luarc = "etc"
```

Now the generated `luarc.json` entry correctly points to **etc**:

```json
".lux/5.1/test_dependencies/5.1/3b7...5-luassert_intellisense@d352...9-1/etc"
```

---

### Notes

This works across dependency types. The `luarc` field name can be a little confusing for git dependencies, which are probably the main use case. It asks for a luarock-style subpath in a plain git source, so it can look odd (“what’s `/etc` in my repo?”). Since this mainly targets our Git pseudo luarock stubs, the tradeoff feels reasonable as long as we add a short note under “Git Dependencies” in the docs.

---

### LLM Disclosure

I worked through the problem myself and used GPT-5 to draft the Rust implementation. We stumbled a couple of times before landing on something that made sense to me. IANARD, but I really like its type system.

### DIFF
--- a/lux-lib/src/operations/gen_luarc.rs
+++ b/lux-lib/src/operations/gen_luarc.rs
@@ -76,14 +76,23 @@ async fn do_generate_luarc(args: GenLuaRc<'_>) -> Result<(), GenLuaRcError> {
         .await
         .unwrap_or_else(|_| "{}".into());
 
+    // Read any optional overrides from lux.toml to allow non-src library dirs for LuaLS
+    let luarc_overrides = read_luarc_dependency_overrides(project);
+
     let dependency_tree = project.tree(config)?;
     let dependency_dirs = lockfile
         .local_pkg_lock(&LocalPackageLockType::Regular)
         .rocks()
         .values()
-        .map(|dependency| dependency_tree.installed_rock_layout(dependency))
-        .filter_map(Result::ok)
-        .map(|rock_layout| rock_layout.src)
+        .flat_map(|dependency| {
+            let name = dependency.name().to_string();
+            let override_dirs = luarc_overrides.get(&name).cloned();
+            dependency_tree
+                .installed_rock_layout(dependency)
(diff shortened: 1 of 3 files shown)


mrcjkb:
Thanks for the PR :pray: 

I don't think this approach is how we will want to solve it.
It would probably be better just to create a RockSpec for the package so that the meta sources are installed into the correct source directory.

See also #953.

keithharvey:
Yeah, agreed. I wouldn’t want to formally support this long-term either; once something like this lands, it’s hard to ever remove. The only reason I reached for it is that even core libs like `busted` and `luassert` don’t yet follow the convention, and there’s a decent chance we’ll keep seeing that in older or niche repos. So it felt like a practical bridge until the rockspecs exist.

Please close this if you are sure. I'll attempt the LLS solution and see how feasible it is.

keithharvey:
Closing this, I'm pretty sure

PR #1267, opened 2025-12-30
refactor!: prevent dependency cycles

### LINKED ISSUE #1245: [Refactor/CI] investigate tooling to prevent cyclic dependencies
For example, [cargo-ferris-wheel](https://docs.rs/cargo-ferris-wheel/latest/cargo_ferris_wheel/).

mrcjkb:
Closes #1245.


mrcjkb:
Hmm, it seems to fail not on cyclic module dependencies, but also on functions that recurse on themselves, etc. Adding `--no-fns --no-sysroot --no-traits --no-types --no-uses` flags doesn't seem to help.

Issue #1266, opened 2025-12-29
[Feature] vendored sources


mrcjkb:
> We'll have to implement something equivalent to [cargo-vendor](https://docs.rs/crate/cargo-vendor/latest) for Lux.
> 
> See also: https://github.com/NixOS/nixpkgs/blob/1884f539e5e331e121d261b37c79c108d5d3e91b/pkgs/build-support/rust/import-cargo-lock.nix#L292
> 
> I'll see if I can find some time for it soon 😄 

See #1265

PR #1271, opened 2026-01-04
feat: `vendor` command

mrcjkb:
Prerequisite for #1266


mrcjkb:
The work on this PR is done, but it can't be merged yet, as it is useless without a follow-up PR that adds the ability to use the vendored sources.

mrcjkb:
todo:

- [x] vendor rockspecs
- [x] vendor sources
- [x] vendor packed rocks

Issue #1298, opened 2026-01-19
[BUG] luafilesystem built with Lux cannot be used on macOS or Windows


eamonburns:
`lx test --help` has a notice for `busted-nlua`:

> currently broken on macOS and Windows

There is also a comment for a test:

https://github.com/lumen-oss/lux/blob/91dd7a6baf9ca7a2f62f3f1c70ddb7a6c4b8aabe/lux-lib/tests/test.rs#L100-L105

Is there any workaround using just the `busted` backend and manually specifying the `nlua` interpreter?

mrcjkb:
Not to my knowledge.
Iirc, I was able to reproduce the segfault in Neovim without busted or Lux (which would make this a Neovim issue).

mrcjkb:
Just ran a test on macos. The segfault only occurs if luafilesystem is built using Lux, not LuaRocks.
So this is a Lux issue.

I also run into this on macos - which is likely related:

```
lx --lua-version 5.1 install luafilesystem
lx --lua-version 5.1 lua
> require('lfs')

error loading module 'lfs' from file '/Users/mrcjkb/Library/Application Support/org.lumenlabs.lux/tree/5.1/611b19eca413e8557258c177563402d31094cde79f82d340f5648d2ac8f3b8a1-luafilesystem@1.9.0-1/lib/lfs.so':
        dynamic libraries not enabled; check your Lua installation
stack traceback:
        [C]: ?
        [C]: in function 'require'
        stdin:1: in main chunk
        [C]: ?
```

mrcjkb:
#1304 fixes the busted-nlua test runner on macOS.
We will have to wait for https://github.com/lunarmodules/luasystem/pull/81 to be released before I can continue work on troubleshooting/fixing it for Windows.

mrcjkb:
I just bumped nixpkgs in #1258 and the busted-nlua tests failed because Neovim (downloaded by actions-setup-vim) was built with an older glibc version than the one Lux used to build the `luasystem` and `luafilesystem` packages. Perhaps this is what is happening on Windows, too.

mrcjkb:
Potential references for building Lua for MSVC: 

- https://github.com/luarocks/gh-actions-lua/blob/master/main.js
- https://github.com/vtudorache/lua-msvc

PR #1304, opened 2026-01-21
fix(test/busted-nlua): unbreak on macOS

mrcjkb:
Partial fix for #1298.
Follow-up to #1300.

### DIFF
--- a/lux-lib/src/lua_installation/mod.rs
+++ b/lux-lib/src/lua_installation/mod.rs
@@ -258,7 +258,12 @@ impl LuaInstallation {
 
     /// NOTE: In luarocks, these are behind a link_lua_explicity config option
     pub(crate) fn lib_link_args(&self, compiler: &cc::Tool) -> Vec<String> {
-        self.dependency_info.lib_link_args(compiler)
+        if cfg!(target_os = "macos") {
+            // On macos, linking Lua can lead to duplicate symbol errors
+            Vec::new()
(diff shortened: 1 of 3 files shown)


mrcjkb:
Not marking this as draft because I need the macos and windows tests to run in CI.

mrcjkb:
After #1300, lua 5.1 no longer errors on macos when requiring `lfs` (built with Lua 5.1 headers), but luajit (and neovim built with luajit) segfaults.

When requiring `lfs` built with LuaJIT headers, neither nvim or luajit segfault.

lux-lib/src/build/utils.rs
```diff
@@ -455,6 +455,7 @@ async fn link_c_artifacts(
         cmd.arg("/link")
             .arg(format!("/DEF:{}", def_file.display()))
             .arg(format!("/OUT:{}", output_path.display()))
+            .arg("/FORCE:MULTIPLE") // HACK: Workaround for https://github.com/lunarmodules/luasystem/pull/81
```
mrcjkb: Workaround for https://github.com/lunarmodules/luasystem/pull/81.
Doesn't work, because the duplicate symbol (I believe) leads to this error, which is thrown by luasystem via lua-term:

```
E5113: Error while calling lua chunk: 'for' initial value must be a number
    stack traceback:
    	[C]: in function 'isatty'
```

PR #1371, opened 2026-03-08
refactor!: add true sandboxing support to Lux, simplify packaging process

vhyrro:
(this PR should be merged with rebase to preserve commits)

Recently we've started relying on `luau` and `mlua` for parsing of rockspecs, manifests and other Lua scripts. This PR rewires the codebase to use `piccolo` instead -- a sandboxed, stackless Lua VM written in Rust. For our use cases, it boasts very high compatibility (98.7%) with all known rockspecs hosted on luarocks.org.

This massively simplifies our internal architecture, since we now solely rely on `serde` for deserialization instead of having duplicate logic for Lua. It also greatly simplifies packaging the project, since we no longer have to synchronize the features passed between `lux-lua` and `lux-lib`. In fact, `lux-lua` is now self-contained and no longer needs a complicated build process.

With piccolo we can now also protect against DoS attacks by limiting the amount of computations a script can perform before it is timed out :)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1993,28 +1993,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "gc-arena"
-version = "0.5.3"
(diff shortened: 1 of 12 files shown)


vhyrro:
Really weird that the integration tests are failing... I'll have to take a look.

PR #1389, opened 2026-03-19
fix: add missing DisplayAsLuaKV implementation for `cargo_extra_args`

vhyrro:
in a crazy turn of events this caused `cargo_extra_args` to not be included in the published  `lux-lua` rockspecs, thereby making them unbuildable

### DIFF
--- a/lux-lib/src/lua_rockspec/build/mod.rs
+++ b/lux-lib/src/lua_rockspec/build/mod.rs
@@ -887,6 +887,17 @@ impl DisplayAsLuaKV for BuildSpecInternal {
                 ),
             });
         }
+        if let Some(cargo_extra_args) = &self.cargo_extra_args {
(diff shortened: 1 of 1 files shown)


vhyrro:
It's quite true that we should probably create a comprehensive test suite for this part of the code, or even better make it a derive macro on structs so that way we never miss a field. For now, since this is a trivial fix, I'll merge it as-is once tests pass to make sure we publish a fixed release ASAP.

mrcjkb:
> It's quite true that we should probably create a comprehensive test suite for this part of the code, or even better make it a derive macro on structs so that way we never miss a field. For now, since this is a trivial fix, I'll merge it as-is once tests pass to make sure we publish a fixed release ASAP.

We should also add some roundtrip property tests for types that can be parsed from Lua and have a DisplayAsLuaKV instance.

PR #1399, opened 2026-03-22
feat: add `lx sync` command

vhyrro:
Sometimes the local package installs can get out of sync with the `lux.toml`. Having a `sync` command to get it all back on track is a nice convenience feature.

Note: there seems to be an issue where if you remove `.lux/`, `lx sync` will still claim that everything's alright. our `Sync` builder should probably also verify that the files exist on disk and reinstall if not.


vhyrro:
in principle, yes. But `lx build --only-deps` is not concerned with syncing test dependencies and it's less discoverable for the average user. This is for those cases where you're working on some Lua code with LSP and you want to ensure that all the source files have been pulled down properly for type checking. Wdyt?

PR #1400, opened 2026-03-22
feat(sync): if package files are deleted then redownload the package

vhyrro:


vhyrro:
I'll probably have a look at the failing tests tomorrow.

lux-lib/src/operations/sync.rs
```diff
@@ -35,6 +38,8 @@ pub struct Sync<'a> {
     progress: Option<Arc<Progress<MultiProgress>>>,
     /// Whether to validate the integrity of installed packages.
     validate_integrity: Option<bool>,
+    /// When `true`, skip filesystem existence checks and rely on the lockfile alone.
```
mrcjkb: suggestion: Clarify that we're using the install tree's lockfile, not the project lockfile.

```suggestion
    /// When `true`, skip filesystem existence checks and rely on the install tree's lockfile alone.
```

mrcjkb (changes requested):

PR #1402, opened 2026-03-23
refactor: move `LuaVersion` out of `config` module

vhyrro:
this might just have been the most boring task of my life :joy:


vhyrro:
moved `LuaVersion` out of `config` into a new `lua_version` module. This is a precursor to the `AvailableLuaVersions` struct which will give us more granularity over which Lua version to use for a given codebase.

PR #1414, opened 2026-03-25
feat(lux-lua/operations): more lux-lua API surface

vhyrro:
For lux.nvim to be finalized we need to expose more of the operations. Keeping this as draft because I want to add `add` and `remove` for per-project operations.

This PR might get refactored a lot so I recommend doing a review after I undraft it :)

### DIFF
--- a/lux-lua/src/lua_impls.rs
+++ b/lux-lua/src/lua_impls.rs
@@ -1654,7 +1654,7 @@ impl Into<BuildBehaviour> for BuildBehaviourLua {
 /// ```lua
 /// "say >= 1.3"
 ///
-/// { package = "say >= 1.3", entry_type = "entrypoint", pin = false, opt = false, build_behaviour = false }
+/// { package = "say >= 1.3", entry_type = "entrypoint", pin = false, opt = false, build_behaviour = "no_force" }
 /// ```
 #[derive(Deserialize)]
 #[serde(untagged)]


lux-lua/src/lua_impls.rs
```diff
@@ -1623,3 +1626,110 @@ fn map_dependency_type_names(deps: DependencyType<PackageNameLua>) -> Dependency
...
+    #[default]
+    Entrypoint,
+    DependencyOnly,
+}
+
+/// Intermediate struct for deserialization. Takes on two variants:
+/// ```lua
+/// "say >= 1.3"
+///
+/// { package = "say >= 1.3", entry_type = "entrypoint", pin = false, opt = false, build_behaviour = false }
```
mrcjkb: issue: `build_behaviour: bool` is not intuitive (We should also get rid of the `From<bool> for BuildBehaviour` instance).

How about we name this field `force_rebuild: bool`?
Or, we could make it `---@field build_behaviour 'force' | 'noforce'`.

lux-lua/src/lua_impls.rs
```diff
@@ -1623,3 +1626,110 @@ fn map_dependency_type_names(deps: DependencyType<PackageNameLua>) -> Dependency
...
+                pin,
+                opt,
+                build_behaviour,
+            } => {
+                let et = match entry_type.to_string().as_str() {
+                    "entrypoint" => EntryType::Entrypoint,
+                    "dependency_only" => EntryType::DependencyOnly,
+                    s => {
+                        return Err(LuaError::RuntimeError(format!(
+                            "unknown entry_type '{}': expected 'entrypoint' or 'dependency_only'",
```
mrcjkb: suggestion: Use `EnumIter` to display the the expected values (in case we add more one day).

lux-lua/src/lua_impls.rs
vhyrro: https://github.com/lumen-oss/lux/issues/1419 for the `From<bool>` refactor. I'll probably mirror entry_type here and make it a string `'force' | 'no_force'` (matching snake case)

lux-lua/src/lua_impls.rs
```diff
@@ -1623,3 +1615,130 @@ fn map_dependency_type_names(deps: DependencyType<PackageNameLua>) -> Dependency
...
+            Self::Force => BuildBehaviour::Force,
+        }
+    }
+}
+
+/// Intermediate struct for deserialization. Takes on two variants:
+/// ```lua
+/// "say >= 1.3"
+///
+/// { package = "say >= 1.3", entry_type = "entrypoint", pin = false, opt = false, build_behaviour = false }
```
mrcjkb: issue: documentation is outdated 


```suggestion
/// { package = "say >= 1.3", entry_type = "entrypoint", pin = false, opt = false, build_behaviour = "no_force" }
```

PR #1425, opened 2026-03-31
feat!: remove From<bool> for BuildBehaviour

### LINKED ISSUE #1419: [Refactor] Remove `From<bool>` for `BuildBehaviour`
Makes little sense semantically.
Taken from: https://github.com/lumen-oss/lux/pull/1414#discussion_r2996149583

niklasmarderx:
Remove the `From<bool>` impl for `BuildBehaviour` as suggested in #1419 — the conversion has no clear semantic meaning (`true` → `Force` isn't obvious without reading the impl).

The single call site in `install.rs` now uses an explicit `if`/`else`, and the `Deserialize` impl inlines the same logic directly.

Closes #1419

### DIFF
--- a/lux-lib/src/build/mod.rs
+++ b/lux-lib/src/build/mod.rs
@@ -37,7 +37,6 @@ use make::MakeError;
 
 use patch::{Patch, PatchError};
 use rust_mlua::RustError;
-use serde::{Deserialize, Deserializer};
 use source::SourceBuildError;
 use ssri::Integrity;
 use thiserror::Error;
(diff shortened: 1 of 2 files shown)


lux-lib/src/build/mod.rs
```diff
@@ -174,16 +174,10 @@ impl<'de> Deserialize<'de> for BuildBehaviour {
...
-        if value {
-            Self::Force
+        if bool::deserialize(deserializer)? {
+            Ok(Self::Force)
         } else {
-            Self::NoForce
+            Ok(Self::NoForce)
         }
     }
 }
```
mrcjkb: todo: I'm pretty sure there's no reason for this `Deserialize` instance for `BuildBehaviour` to exist. Let's remove it - it makes no sense.

mrcjkb:
Thanks for the PR 🙏

niklasmarderx:
Removed the entire `Deserialize` impl and the now-unused `serde::{Deserialize, Deserializer}` import. Compiles clean.

mrcjkb:
Renamed PR title prefix because this is a breaking change for lux-lib.

mrcjkb:
Windows failure is unrelated - [community.chocolatey.org](https://community.chocolatey.org/api/v2/) is down.

PR #1450, opened 2026-04-15
ci: give lux-lua its own separate release version

vhyrro:
it's tedious to have to synchronize to the lux-cli version at all times (we can't ever provide incremental updates to lux-lua). this adds github-only releases to lux-lua and also does the proper artifact building for those releases.

### DIFF
--- a/lux-lua/Cargo.toml
+++ b/lux-lua/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "lux-lua"
-version.workspace = true
+version = "0.28.1"
 edition = "2021"
 publish = false
 
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -21,7 +21,7 @@ jobs:
   dist-package:
     name: ${{ matrix.target }}
     if: github.repository_owner == 'lumen-oss' && (
-      ( github.event_name == 'release' && startsWith(github.event.release.name, 'v')) ||
+      ( github.event_name == 'release' && (startsWith(github.event.release.name, 'v') || startsWith(github.event.release.name, 'lux-lua-v'))) ||
         github.event_name == 'workflow_dispatch' ||
         github.event_name == 'pull_request'
       )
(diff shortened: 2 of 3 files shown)


lux-lua/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "lux-lua"
-version.workspace = true
+version = "0.28.1"
```
mrcjkb, line 3: issue(needs-discussion): We keep the lux-cli and lux-lua versions in sync for a few reasons:

- Our lux-cli release artifacts are bundled with lux-lua (e.g. so commands like `lx run` can use the `lux.loader`).
- The same is done in package repos ([nixpkgs](https://github.com/NixOS/nixpkgs/blob/7e495b747b51f95ae15e74377c5ce1fe69c1765f/pkgs/development/lua-modules/lux-lua.nix#L29), [arch linux](https://gitlab.archlinux.org/archlinux/packaging/packages/lux/-/blob/main/PKGBUILD?ref_type=heads), [termux](https://github.com/termux/termux-packages/blob/cf2186b5bfc57221030816bcf0f3c407fdf576c8/packages/lux-cli/build.sh#L4)).
  
> [!CAUTION]
>
> This change **would be breaking**:
> - For those package repos, as they use the same source source pin for both lux-cli and lux-lua.
If the lux-lua version can diverge from lux-cli, package repos will have to use different packaging instructions for lux-cli and lux-lua, with different source pins. Otherwise, lux-cli might end up being bundled with a build of lux-lua that hasn't been released.
It's doable, but this will require communication with the various package maintainers.
> - For our own CI, as `cargo xtask dist-package` bundles the currently checked out version of lux-lua.

For iterating on lux.nvim development, it might be easier to build lux-lua locally, rather than pulling GitHub release artifacts.

vhyrro: That's sensible, it's a tedious system but if it really does help then it's best we keep it :)

PR #1473, opened 2026-04-26
fix(fmt): allow specifying relative paths

benniekiss:
This PR allows passing relative paths of files and folders to `lx fmt`. Relative paths are now resolved against the project root folder. It is still possible to pass absolute paths since `Path.join` will replace the path if so. 

I also removed what I believe to be an unnecessary call to `to_path_buf()`.

Previously, the `startswith` check would only pass if absolute paths were provided since the `file.path()` is absolute.

### DIFF
--- a/lux-cli/src/format.rs
+++ b/lux-cli/src/format.rs
@@ -36,7 +36,6 @@ pub fn format(args: Fmt) -> Result<()> {
 
     let workspace_or_file = args
         .workspace_or_file
-        .as_ref()
         .map(|path| std::path::absolute(path).ok())
         .flatten();
 
--- a/lux-cli/src/format.rs
+++ b/lux-cli/src/format.rs
@@ -37,7 +37,8 @@ pub fn format(args: Fmt) -> Result<()> {
     let workspace_or_file = args
         .workspace_or_file
         .as_ref()
-        .map(|path| project.root().join(path));
+        .map(|path| std::path::absolute(path).ok())
+        .flatten();
 
     WalkDir::new(project.root().join("src"))
         .into_iter()
--- a/lux-cli/src/format.rs
+++ b/lux-cli/src/format.rs
@@ -34,6 +34,11 @@ pub fn format(args: Fmt) -> Result<()> {
         .map(|config: String| toml::from_str(&config).unwrap_or_default())
         .unwrap_or_default();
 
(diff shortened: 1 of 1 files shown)


mrcjkb (changes requested):
Thanks 🙏 

In its current implementation, this is a breaking change. 

suggestion: Perhaps we should just use `std::path::absolute`

benniekiss:
`path::absolute` definitely seems like the better choice! I'll push the change in a little

lux-cli/src/format.rs
```diff
@@ -34,6 +34,11 @@ pub fn format(args: Fmt) -> Result<()> {
         .map(|config: String| toml::from_str(&config).unwrap_or_default())
         .unwrap_or_default();
 
+    let workspace_or_file = args
+        .workspace_or_file
+        .map(|path| std::path::absolute(path).ok())
```
mrcjkb: issue: `.ok()).flatten()` fails silently. Use `?` instead, to fail if `absolute` fails.
You will have to convert this from a `map` to a pattern match to be able to propagate the error.

mrcjkb (changes requested):

lux-cli/src/format.rs
benniekiss: [`transpose`](https://doc.rust-lang.org/std/option/enum.Option.html#method.transpose) actually seems purpose built for this!

mrcjkb (approved):

lux-cli/src/format.rs
mrcjkb: Oh yeah, forgot about that.

lux-cli/src/format.rs
```diff
@@ -34,6 +34,11 @@ pub fn format(args: Fmt) -> Result<()> {
         .map(|config: String| toml::from_str(&config).unwrap_or_default())
         .unwrap_or_default();
 
+    let workspace_or_file = args
+        .workspace_or_file
+        .map(|path| std::path::absolute(path))
```
mrcjkb: clippy is failing:


```suggestion
        .map(std::path::absolute)
```

mrcjkb (changes requested):

mrcjkb (approved):

PR #1480, opened 2026-04-29
feat(http): set User-Agent header on outbound HTTP requests

### LINKED ISSUE #1479: Set a user agent when communicating with luarocks.org
Would be nice to have a user agent when talking to luarocks.org so we can see adoption, versions etc.

The luarocks cli currently sends something like `LuaRocks/3.11.0 linux-x86_64 via LuaSocket`

Minimally `lux/VERSION` would be helpful. I didn't see anything in this repo already setting the user agent

Thanks

tmchow:
Closes #1479

Adds a `USER_AGENT` constant and an `http_client()` helper in `lux-lib/src/lib.rs` so every outbound HTTP request to a luarocks server identifies itself as `lux/<CARGO_PKG_VERSION>`. As you said in the issue, the luarocks CLI sends `LuaRocks/3.11.0 ...`; this is the same idea so the server-side maintainers can see adoption + version distribution.

## What this PR does NOT touch

- Test files and fixtures continue to use `reqwest::Client::new()`; they hit local mocks where a UA isn't useful and dropping in `http_client()` would require plumbing through `lux-lib`.
- `TOOL_VERSION` (the API-compat string) is unchanged; this PR only adds `USER_AGENT`.
- No reqwest feature flags or version changes.

## Verification

- `cargo build -p lux-lib` and `cargo build -p lux-lib --no-default-features` both clean.
  - The full workspace build needs `gpg-error` system lib (gpgme feature), which I don't have on this machine, so I built lux-lib directly.
- `codex review --base upstream/main` returned: "No actionable correctness issues were found in the diff. The changes consistently apply a configured reqwest User-Agent while preserving the existing upload HTTPS-only client configuration."
- `git diff --check` clean.

Once merged, luarocks.org will see traffic like `User-Agent: lux/0.36.4` from every lux instance.

### DIFF
--- a/lux-lib/src/lib.rs
+++ b/lux-lib/src/lib.rs
@@ -28,5 +28,18 @@ pub(crate) mod variables;
 /// two versions match (meaning we can safely communicate with the server).
 pub const TOOL_VERSION: &str = "1.0.0";
 
+/// User-Agent string sent on all outbound HTTP traffic to luarocks-compatible servers.
+pub const USER_AGENT: &str = concat!("lux/", env!("CARGO_PKG_VERSION"));
+
+/// Returns a reqwest::Client preconfigured with the lux User-Agent header.
+/// Use this for every outbound HTTP request to a luarocks server so the
+/// upstream operator can see adoption / version distribution.
+pub fn http_client() -> reqwest::Client {
+    reqwest::Client::builder()
+        .user_agent(USER_AGENT)
+        .build()
+        .expect("failed to build default reqwest client")
+}
+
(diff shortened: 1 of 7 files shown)


mrcjkb:
[Strict No LLM/AI Policy - CONTRIBUTING.md](https://github.com/lumen-oss/lux/blob/main/CONTRIBUTING.md#strict-no-llm--no-ai-policy)

tmchow:
Sorry!  Skill was used to generate the PR desc.

PR #1481, opened 2026-04-30
feat!: set user agent header for web requests

### LINKED ISSUE #1479: Set a user agent when communicating with luarocks.org

mrcjkb:
- Wraps `reqwest::Client` constructor/builder calls with a `new_https_client` wrapper.
- Sets the user agent from the config, with the following defaults:
  - lux-cli: `lux/<version>`
  - lux-lib: `lux-lib/<version>`
  - lux-lua: `lux-lua/<version>`
  (Note that `<version>` may be different for each of the crates).
- **BREAKING:** ~All~ Most web requests are now HTTPS-only (previously, only `upload` requests were).
  I suggest YAGNI for HTTP support - We can enable it again if people need it.

### DIFF
--- a/lux-cli/src/bin/lx.rs
+++ b/lux-cli/src/bin/lx.rs
@@ -17,6 +17,8 @@ use lux_lib::{
     lua_version::LuaVersion,
 };
 
+const DEFAULT_USER_AGENT: &str = concat!("lux/", env!("CARGO_PKG_VERSION"));
+
 #[tokio::main(flavor = "multi_thread")]
 async fn main() -> Result<()> {
     let cli = Cli::parse();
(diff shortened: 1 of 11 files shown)


mrcjkb:
:thinking: using a HTTPS-only client causes a source integrity check failure.

mrcjkb:
> 🤔 using a HTTPS-only client causes a source integrity check failure.

Disabling HTTPS-only mode for fetching sources, as `source.url` may be HTTP URLs and I don't want to unnecessarily break existing lockfiles.

mrcjkb:
Merging. We can make changes in a follow-up PR if necessary.

teto:
good thing you reverted the https -only.

PR #1502, opened 2026-05-10
refactor!: prepare to support workspaces with multiple projects

mrcjkb:
This is a breaking change for `lux-lib` and `lux-lua`, as the entry point into a single-project workspace is `Workspace::current` and not `Project::current`.


lux-cli/src/add.rs
```diff
@@ -38,11 +39,19 @@ pub struct Add {
...
+    /// Project to modify.
+    #[arg(short, long, visible_short_alias = 'p')]
+    package: Option<PackageName>,
 }
 
 pub async fn add(data: Add, config: Config) -> Result<()> {
-    let mut project = Project::current()?.ok_or_eyre("No project found")?;
-
+    let mut workspace = Workspace::current_or_err()?;
+    if data.package.is_none() && workspace.members().len() > 1 {
```
vhyrro: suggestion: perhaps enforce this condition at the type level via `Workspace::current_or_err()?.single_or_select(data.package)?`, which returns a single `Project` instead of having to then iterate afterwards?
I feel like it could be easy for us to forget to include these guards in other internal code in the future, and the current behaviour is a bit hacky (later on we are looping over an array of a single package).

lux-cli/src/add.rs
```diff
@@ -56,13 +65,20 @@ pub async fn add(data: Add, config: Config) -> Result<()> {
...
 
     if !data.package_req.is_empty() {
-        project
-            .add(lua_dependency::DependencyType::Regular(dependencies), &db)
-            .await?;
-        project
-            .add_git(lua_dependency::LuaDependencyType::Regular(git_dependencies))
-            .await?;
-        sync_dependencies_if_locked(&project, progress.clone(), &config).await?;
+        for project in workspace.try_members_mut(&data.package)? {
```
vhyrro: suggestion: remove the need for iterating here (see suggestion above)

lux-cli/src/pin.rs
```diff
@@ -29,53 +29,81 @@ pub struct ChangePin {
...
+    #[arg(short, long, visible_short_alias = 'p')]
+    package: Option<PackageName>,
 }
 
 pub async fn set_pinned_state(data: ChangePin, config: Config, pin: PinnedState) -> Result<()> {
-    match Project::current()? {
-        Some(mut project) => {
+    match Workspace::current()? {
+        Some(mut workspace) => {
+            if data.package.is_none() && workspace.members().len() > 1 {
```
vhyrro: suggestion: also remove the implicit if check here

lux-cli/src/remove.rs
```diff
@@ -23,33 +24,55 @@ pub struct Remove {
...
-    let mut project = Project::current()?.ok_or_eyre("No project found")?;
+    let mut workspace = Workspace::current()?.ok_or_eyre("No project found")?;
     let progress = MultiProgress::new_arc(&config);
 
-    if !data.package.is_empty() {
-        project
-            .remove(lua_dependency::DependencyType::Regular(data.package))
-            .await?;
-        sync_dependencies_if_locked(&project, progress.clone(), &config).await?;
+    if data.package.is_none() && workspace.members().len() > 1 {
```
vhyrro: ditto here

lux-cli/src/update.rs
```diff
@@ -32,14 +32,24 @@ pub struct Update {
...
 
 pub async fn update(args: Update, config: Config) -> Result<()> {
     let progress = MultiProgress::new_arc(&config);
     progress.map(|p| p.add(ProgressBar::from("🔎 Looking for updates...".to_string())));
 
     if args.toml {
-        let mut project = Project::current()?.ok_or_eyre("No project found")?;
+        let mut workspace = Workspace::current_or_err()?;
+
+        if args.package.is_none() && workspace.members().len() > 1 {
```
vhyrro: ditto here

lux-lib/src/workspace/mod.rs
```diff
@@ -0,0 +1,340 @@
...
+    }
+
+    /// Mutable reference to the members of this workspace.
+    pub fn members_mut(&mut self) -> &mut NonEmpty<Project> {
+        &mut self.members
+    }
+
+    /// Try to iterate over the projects in this workspace, optionally filtered by package name.
+    /// Fails if a package name is specified, but not found.
+    pub fn try_members(&self, name: &Option<PackageName>) -> Result<Vec<&Project>, WorkspaceError>
```
vhyrro: suggestion: fundamentally this function seems like it has a wrong signature. we have an iteration function whose optional parameter restricts it always to a list of 1 items. I can see that this shortens the code in many places (because we don't have to check for an individually selected package but can just use a loop always), but it requires implicit if checks which can be forgotten. We'd benefit from a more explicit function which natively does all checks for us:
```rs
workspace.select_or_for_each(data.package, |project| project.do_work())?;
// or whatever naming would work best
```
With this we should be able to remove `try_members` and `try_members_mut` altogether? Then rename `try_member` to `select` or `select_package()`.

vhyrro (dismissed):
rest of the code looks great, excellent work 🎉

lux-cli/src/update.rs
mrcjkb: Actually, there's no reason to restrict the number of projects to operate on for the `update` command. We ignore dependencies that aren't found in the project toml anyway.

lux-lib/src/workspace/mod.rs
mrcjkb: I don't think making it take a lambda instead of returning an iterator makes the signature any more correct. 
Personally, I'm not a fan of `for_each` functions. Unlike `map`, `and_then`, `filter`, etc., which offer compositional benefits, `for_each` is purely for side-effects (it doesn't buy us anything apart from just iterating).
But compared to simple for loops, it makes propagating errors with `?` and dealing with async more awkward.

> it requires implicit if checks which can be forgotten

We don't always need those checks - just for operations that should only be performed on a single package.

vhyrro: >  Personally, I'm not a fan of for_each functions. Unlike map, and_then, filter, etc., which offer compositional benefits, for_each is purely for side-effects (it doesn't buy us anything apart from just iterating).

That's fine, I'm just wondering if it's possible to create a nicer signature instead of a function like this to skip the need for implicit if checks on member length. I'd even prefer more code and more verbosity over the implicit single-loop we have.

> We don't always need those checks - just for operations that should only be performed on a single package.

Yep, the point is that these checks are not encoded in the type system in any way and aren't enforceable. It makes the flow of the code hard to follow (despite the fact that there's a for loop, in all cases apart from `update` we never allow operating on more than one project at a single time, so the loop is misleading).

Ultimately this is getting quite bikesheddy though so you're free to resolve the conversation and we can merge as is. It's just a small code style disagreement :)

mrcjkb: I've replaced `try_members_*` with `.select_member(package: &PackageName)` and `.members()`, combined with an `if let Some(package) ...` check at the call sites.
It adds quite a bit of boilerplate, but it's not too bad as long as the stuff in each branch is extracted into a function.

vhyrro (approved):

Issue #1148, opened 2025-10-27
[Feature] local dependencies


mrcjkb:
### Discussed in https://github.com/lumen-oss/lux/discussions/1144

<div type='discussions-op-text'>

<sup>Originally posted by **ALameLlama** October 26, 2025</sup>
I was wonder if it's currently possible to install local packages e.g
lib-1
lib-2

lib-2 needs to install lib-1.

I've tried a few different things by setting the source.url and generating a rockspec and then trying to install the rockspec but I can't get this working.

Is there a better way to handle this? I want to be able to install packages from a local source so I can test changes before pushing anything.

Thanks
</div>

PR #1503, opened 2026-05-10
feat: support workspaces with multiple projects

### LINKED ISSUE #1325: [Feature] multi-project support
These days, it's not uncommon to have multiple projects in a monorepo.
Package managers like cabal and cargo support this - why not Lux?

mrcjkb:
Stacked on #1502.
Closes #1325.

This does not yet add support for local dependencies (#1148, e.g. packages in a workspace depending on each other). That will follow in a later PR:

- #1508

### DIFF
--- a/lux-lib/src/workspace/mod.rs
+++ b/lux-lib/src/workspace/mod.rs
@@ -16,7 +16,10 @@ use crate::{
     package::PackageName,
     project::{Project, ProjectError, PROJECT_TOML},
     tree::{Tree, TreeError},
+    workspace::workspace_toml::WorkspaceToml,
 };
+
+pub mod workspace_toml;
 
 pub const WORKSPACE_TOML: &str = PROJECT_TOML;
 pub(crate) const LUX_DIR_NAME: &str = ".lux";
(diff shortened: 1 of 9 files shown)


lux-lib/src/workspace/mod.rs
```diff
@@ -250,9 +257,8 @@ where {
                 WorkspaceError::ReadLuxTOML(toml_path.to_string_lossy().to_string(), err)
             })?;
             let root = start.as_ref();
-
             if toml_content.contains("[workspace]") {
-                unimplemented!("multi-project workspaces")
+                Ok(Some(Self::from_toml(&toml_content, root)?))
```
vhyrro: issue: this can be easily defeated if someone has `summary = "[workspace] file for <project_name>"`. perhaps do a fallback instead of trying to rely on substring checking? `Self::from_toml(...).or(...)`, with some way of distinguishing "there is no [workspace]" vs "there was an error".

vhyrro (changes requested):

lux-lib/src/workspace/mod.rs
mrcjkb: I don't want to try to parse the workspace toml with a fallback, because I want to fail if it's invalid.
But we can parse a `toml::Table` object and check if it has a `"workspace"` key.

vhyrro (dismissed):

Issue #1108, opened 2025-10-07
[Feature] cache builds


mrcjkb:
Currently, `lx build` rebuilds even if the source has not changed. 
It might be useful to compute the hash of all sources, to avoid rebuilding if the package has already been built.

PR #1508, opened 2026-05-14
feat: support local dependencies

### LINKED ISSUE #1148: [Feature] local dependencies

mrcjkb:
Stacked on #1503.
Closes #1148.

For now, this builds local dependencies twice (once as a dependency, then as the main package).
In most cases, this should not be an issue, as we by default only force a rebuild for entrypoints. If it does turn out to be awkward, we can address it in a follow-up PR or  #1108.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,8 +100,9 @@ The following table provides a brief comparison:
 | Static type checking                                                                    | :white_check_mark:           | :x:                |
 | Generate a `.luarc` file with dependencies                                              | :white_check_mark:           | :x:                |
 | Git dependencies in local projects                                                      | :white_check_mark:           | :x:                |
-| Vendor sources for offline use                                                          | :white_check_mark:           | :x:                |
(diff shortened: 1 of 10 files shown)


mrcjkb:
Todo:

- [x] resolve local dependencies for build
- [x] test case for multi-project workspace with local dependencies

vhyrro (dismissed):

PR #1514, opened 2026-05-18
feat(lua): restore platform build targets and add freebsd support

benniekiss:
This PR restores the correct platform build targets when building lua distributions.

The targets were actually removed in a previous commit, but this introduced a bug where, at least on macOS, `lx lua` could not load dynamic modules because it was compiling with the `generic` target. I went ahead and added a build target for `freebsd` as well for completeness, but I don't have a way to test/verify the target.

The previous commit and current behavior also make a distinction between lua version 5.4 and whether to use `linux-readline` or the `linux` build target. I kept this branch in, but I don't know how necessary it is.

I also removed some redundant references on strings because of a clippy warning.

I think this should also fix issues when running tests on non-5.4 lua versions, which I believed was caused by the incorrect version handling in the build path.

- [X] Fits [CONTRIBUTING.md].
- [X] Fits [No LLM / No AI Policy].

### DIFF
--- a/lux-lib/src/operations/build_lua.rs
+++ b/lux-lib/src/operations/build_lua.rs
@@ -440,12 +440,19 @@ async fn do_build_lua_unix(
     let progress = args.progress;
     let install_dir = args.install_dir;
 
-    progress.map(|p| p.set_message(format!("🛠️ Building Lua {}", &pkg_version)));
+    progress.map(|p| p.set_message(format!("🛠️ Building Lua {}", pkg_version)));
 
-    let build_target = if cfg!(target_os = "linux") && matches!(&lua_version, LuaVersion::Lua54) {
-        "linux"
+    let build_target = if cfg!(target_os = "linux") {
+        if matches!(&lua_version, LuaVersion::Lua54) {
+            "linux"
+        } else {
+            "linux-readline"
+        }
+    } else if cfg!(target_os = "macos") {
+        "macosx"
+    } else if cfg!(target_os = "freebsd") {
+        "freebsd"
     } else {
-        // For other lua versions, the "linux" target uses readline.
         "generic"
     };
 
--- a/lux-lib/src/operations/build_lua.rs
+++ b/lux-lib/src/operations/build_lua.rs
@@ -477,7 +477,7 @@ async fn do_build_lua_unix(
         Ok(output) if build_target != "generic" => {
             progress.map(|p| {
                 p.println(format!(
-                    r#"⚠️ WARNING: Could not build for platform '{build_target}'. Attempting a 'generic' platform build.
+                    r#"⚠️ WARNING: Could not build Lua target '{build_target}'. Attempting a 'generic' Lua target build.
 Some functionality may be limited, including dynamic module loading and REPL history.
     
 stderr:
--- a/lux-lib/src/operations/build_lua.rs
+++ b/lux-lib/src/operations/build_lua.rs
@@ -440,8 +440,6 @@ async fn do_build_lua_unix(
     let progress = args.progress;
     let install_dir = args.install_dir;
 
-    progress.map(|p| p.set_message(format!("🛠️ Building Lua {}", pkg_version)));
-
     let build_target = if cfg!(target_os = "linux") {
         // only lua 5.4 has a specific `linux-readline` target
         if matches!(&lua_version, LuaVersion::Lua54) {
@@ -457,6 +455,13 @@ async fn do_build_lua_unix(
         "generic"
     };
 
+    progress.map(|p| {
+        p.set_message(format!(
+            "🛠️ Building Lua {} ({})",
+            pkg_version, build_target
+        ))
+    });
+
     match Command::new(config.make_cmd())
         .current_dir(build_dir)
         .stdout(Stdio::piped())
@@ -466,6 +471,43 @@ async fn do_build_lua_unix(
         .await
     {
         Ok(output) if output.status.success() => utils::log_command_output(&output, config),
+        // The lua build may fail for some platform specific reason.
+        // Typically, the `readline` headers cannot be found. In these
+        // cases, try the build again with the `generic` platform target.
+        Ok(output) if build_target != "generic" => {
+            utils::log_command_output(&output, config);
+            progress.map(|p| {
+                p.println(format!(
+                    "Could not build for platform {build_target}. Attempting a generic platform build.\n{}",
+                    "Some functionality may be limited, including dynamic module loading and REPL history.",
+                ))
+            });
+            match Command::new(config.make_cmd())
+                .current_dir(build_dir)
+                .stdout(Stdio::piped())
+                .stderr(Stdio::piped())
+                .arg("generic")
+                .output()
+                .await
+            {
+                Ok(output) if output.status.success() => utils::log_command_output(&output, config),
+                Ok(output) => {
+                    return Err(BuildLuaError::CommandFailure {
+                        name: "build".into(),
+                        status: output.status,
+                        stdout: String::from_utf8_lossy(&output.stdout).into(),
+                        stderr: String::from_utf8_lossy(&output.stderr).into(),
+                    });
+                }
+                Err(err) => {
+                    return Err(BuildLuaError::Io(io::Error::other(format!(
+                        "Failed to run `{} build`:\n{}",
+                        config.make_cmd(),
+                        err,
+                    ))))
+                }
+            }
+        }
         Ok(output) => {
             return Err(BuildLuaError::CommandFailure {
                 name: "build".into(),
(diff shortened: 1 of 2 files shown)


benniekiss:
ill investigate why the build test is failing, since it was passing locally

mrcjkb:
> ill investigate why the build test is failing, since it was passing locally

Linux:

```
make: *** No rule to make target 'linux-readline'.  Stop.
```

macOS:

```
luaconf.h:275:10: fatal error: 'readline/readline.h' file not found
```

IIRC, we had issues with the `linux-readline` target, so we decided not to build it: #976
I don't remember all the details.

benniekiss:
> macOS:
> 

failing on readline for macos is what's interesting to me, since it shouldn't be looking for it. But it seems that the `make` invocation is setting `-DLUA_USE_LINUX`, even for the macOS target.

EDIT: well after looking at the lua makefiles, the macosx target does set the `-DLUA_USE_LINUX` flag for some versions, and it should be using `readline` as well. However, for `lua 5.4`, the makefile sets `-DLUA_USE_MACOSX`, ~~but invoking the tests with nix on my machine, and in the CI, still shows it using the `LUA_USE_LINUX` flag~~ it was just failing on the `lua 5.1` build, so it is just a readline issue. pardon my noise!

mrcjkb:
since readline might not be available, we should probably either:

- check its availability before selecting the target.
- fall back to building the generic target it building the readline target fails (and emit a warning message to stderr).

benniekiss:
> * check its availability before selecting the target.

This is what the previous behavior did. Were there issues with it, or do you think it's ok to restore that code?

The issue with the generic target is that it does not export the dynamic symbols -- it's not built with the `-Wl,-E` flags, and the lua makefile requires `LUA_USE_DLOPEN` to also be set I believe. These options are set when using the platform specific targets. So if there is fallback to the generic build, these flags also have to be set so that modules can be loaded. Implementing it is a little tedious, since each lua version and target has slightly different config requirements.

 It makes sense to me to include that as well, but it adds some maintenance overhead.

mrcjkb:
> This is what the previous behavior did. Were there issues with it, or do you think it's ok to restore that code?

Yes. Even if pkg-config or Lux can find readline, the Lua build can fail to find it. So if we restore that behaviour, we need a way to tell the build target how to find it.

benniekiss:
> we need a way to tell the build target how to find it.

Can that be done with passing `-I{readline_src}` to the lua build? I'm not certain on how to test the flaky lua search 

EDIT: using `-I` is how the lua makefiles set alternate locations, such as using `edit` on the BSDs.

mrcjkb:
> > we need a way to tell the build target how to find it.
> 
> Can that be done with passing `-I{readline_src}` to the lua build? I'm not certain on how to test the flaky lua search
> 
> EDIT: using `-I` is how the lua makefiles set alternate locations, such as using `edit` on the BSDs.

That sounds like the best way to do it.

benniekiss:
So I was trying to get the old header search code working, but on my macOS 26 system, it actually was unable to find the header files, while the regular lua build could find them.

The most recent commits implement a fallback build path. I still think it may be worthwhile to implement the custom header searcher, but I couldn't get my head around why it was unable to find the header files on my system. I also do not know in what circumstances the searcher would find the header files, but lua would not.

For now, this reimplements the platform specific build with a fallback to a generic build if any build error occurs.

lux-lib/src/operations/build_lua.rs
```diff
@@ -458,6 +471,43 @@ async fn do_build_lua_unix(
...
     {
         Ok(output) if output.status.success() => utils::log_command_output(&output, config),
+        // The lua build may fail for some platform specific reason.
+        // Typically, the `readline` headers cannot be found. In these
+        // cases, try the build again with the `generic` platform target.
+        Ok(output) if build_target != "generic" => {
+            utils::log_command_output(&output, config);
+            progress.map(|p| {
+                p.println(format!(
+                    "Could not build for platform {build_target}. Attempting a generic platform build.\n{}",
```
mrcjkb:
```suggestion
                    ""⚠️ WARNING: Could not build Lua build target {build_target}. Attempting a 'generic' platform build.\n{}",
```

It should also include output's stderr.

lux-lib/src/operations/build_lua.rs
```diff
@@ -458,6 +471,43 @@ async fn do_build_lua_unix(
         .await
     {
         Ok(output) if output.status.success() => utils::log_command_output(&output, config),
+        // The lua build may fail for some platform specific reason.
+        // Typically, the `readline` headers cannot be found. In these
+        // cases, try the build again with the `generic` platform target.
+        Ok(output) if build_target != "generic" => {
+            utils::log_command_output(&output, config);
+            progress.map(|p| {
+                p.println(format!(
```
mrcjkb, line 479: issue: This shouldn't be a progress message, as it will disappear when the fallback build is done.
Use `eprintln!` instead.

lux-lib/src/operations/build_lua.rs
mrcjkb: quibble: `format!"\n{}", ""` is a bit awkward.
You can use a [raw string literal](https://doc.rust-lang.org/reference/tokens.html#grammar-RAW_STRING_LITERAL) instead:

```rust
r#"First line
Second line
"#
```

mrcjkb (changes requested):

lux-lib/src/operations/build_lua.rs
benniekiss: should I remove the `log_command_output` above it?

benniekiss: This was to address indentation formatting, since it would be otherwise written as:

```rust
            ....
            r#"⚠️ WARNING: Could not build for platform {build_target}. Attempting a 'generic' platform build.
Some functionality may be limited, including dynamic module loading and REPL history"#
```

But that's fine with me!

benniekiss: Also, the emoji does not print in my terminal when using `eprintln`

lux-lib/src/operations/build_lua.rs
benniekiss: the `progress.println` actually does remain on screen once the build is finished

benniekiss: for reference:

```shell
⚠️ WARNING: Could not build for platform macosx. Attempting a 'generic' platform build.
Some functionality may be limited, including dynamic module loading and REPL history.

STDERR:
lundump.c:244:33: warning: adding 'int' to a string does not append to the string [-Wstring-plus-int]
  244 |   checkliteral(S, LUA_SIGNATURE + 1, "not a");  /* 1st char already checked */
      |                   ~~~~~~~~~~~~~~^~~
lundump.c:244:33: note: use array indexing to silence this warning
  244 |   checkliteral(S, LUA_SIGNATURE + 1, "not a");  /* 1st char already checked */
      |                                 ^  
      |                   &             [  ]
1 warning generated.

  🌔 Installed Lua (5.3) to /Users/benniemt/Library/Application Support/org.lumenlabs.lux/tree/5.3/.lua                                                              

```

lux-lib/src/operations/build_lua.rs
mrcjkb: > should I remove the log_command_output above it?

Yes, that probably makes sense.

mrcjkb: Can you adjust the indentation of the second line in the raw string literal?

benniekiss: PTAL at the recent commit for the formatting, if that's alright

lux-lib/src/operations/build_lua.rs
```diff
@@ -458,6 +471,47 @@ async fn do_build_lua_unix(
         .await
     {
         Ok(output) if output.status.success() => utils::log_command_output(&output, config),
+        // The lua build may fail for some platform specific reason.
+        // Typically, the `readline` headers cannot be found. In these
+        // cases, try the build again with the `generic` platform target.
+        Ok(output) if build_target != "generic" => {
+            progress.map(|p| {
+                p.println(format!(
+                    r#"⚠️ WARNING: Could not build for platform '{build_target}'. Attempting a 'generic' platform build.
```
mrcjkb: suggestion: Not quite what I had suggested previously.

- We should include _what_ we could not build in the warning, otherwise it might be confusing.
- "target" instead of "for platform" (linux-readline is a target, not a platform).

```suggestion
                    r#"⚠️ WARNING: Could not build Lua target: '{build_target}'. Attempting a 'generic' platform build.
```

mrcjkb (changes requested):

lux-lib/src/operations/build_lua.rs
benniekiss: That makes sense to me -- should the `'generic' platform` be changed as well? Like `'generic' Lua target`

mrcjkb: Sure, why not.

mrcjkb (approved):

mrcjkb:
> This may close #1298

The issue was resolved for macos in #1304.
You're probably right about the issue being caused by the `'generic'` target, but this PR doesn't resolve the issue for Windows either.

benniekiss:
Good to know! I was experiencing what I thought was that issue when running tests on macOS, but I can't reproduce it now.

PR #1532, opened 2026-05-27
fix(gen-rockspec): multi-line strings and quotes

### LINKED ISSUE #1531: newlines, backslashes and quotes in lux.toml generate invalid rockspecs
This lux.toml:

```toml
package = "my-lua-project"
lua = ">=5.2"

[source]
url = 'https://example.com'

[description]
summary = "A sample project."
detailed = '''
first line
second line
'''
maintainer = "myself"
labels = [ "learning" ]
license = "MPL-2.0"

[build]
type = "builtin"
```

Generates this rockspec:

```lua
rockspec_format = "3.0"
package = "my-lua-project"
version = "1.3.7-1"

description = {
summary = "A sample project.",
detailed = "first line
second line
",
license = "MPL-2.0",
maintainer = "myself",
labels = {
"learning",
},
}

dependencies = {
"lua>=5.2",
}

source = {
url = "https://example.com",
}

build = {
type = "builtin",
}
```

Which is invalid. Lua short literal strings are not allowed to contain unescaped line breaks. I've noticed the same thing happens with other characters, so a toml `"foo\\bar"` becomes a Lua `"foo\bar"`, which escapes to a backspace character, and a toml `'foo"bar'` becomes a Lua `"foo"bar"`, which is also invalid.

This will need either to escape special characters or use long literal strings (with a leading newline, otherwise leading newlines in the toml data will end up being swallowed), scanning the string contents ahead of time to determine how many equal signs the long string delimiters need.

mrcjkb:
Fixes #1531.

### DIFF
--- a/lux-lib/src/lua_rockspec/serde_util.rs
+++ b/lux-lib/src/lua_rockspec/serde_util.rs
@@ -189,6 +189,8 @@ impl Display for DisplayLuaValue {
             DisplayLuaValue::String(s) => {
                 if s.contains('\n') {
                     write!(buf, "[[\n{s}\n]]")?;
+                } else if s.contains("\"") {
+                    write!(buf, "'{s}'")?;
                 } else {
                     write!(buf, "\"{s}\"")?;
                 }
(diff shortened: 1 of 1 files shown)


Taywee:
This implementation will fail if the string contains the character sequence `]]`. The branches also fail to account for it containing multiple violating characters. For instance, a string with `]]`, `'`, and `"`. This needs proper escaping, or for long strings, to scan the string to see how many equal signs it needs. A string might contain `]]`, `]=]`, and `]==]`, for instance, in which case the string will need to begin with `[===[` and end with `]===]`.

My PR for this is #1534

mrcjkb:
Superseded by #1534.

PR #1545, opened 2026-05-31
feat(workspace): support globs in `members` paths

### LINKED ISSUE #1543: [Feature] support globs in workplace member paths
```toml
[workspace]
members = ["glob:projects/*"]
```

should resolve to all subdirectories in `projects/` that contain a `lux.toml`.

estelledc:
Closes #1543.

## Behavior

Glob patterns in `[workspace] members` now expand against the workspace root, keeping only entries that contain a project file. Plain paths keep the existing exact-lookup behavior, so a typo still errors with `ProjectNotFound`.

```toml
[workspace]
members = ["projects/*"]
```

resolves to every subdirectory of `projects/` that has a `lux.toml`.

## Why silently skip non-project matches?

Real workspaces will end up matching co-located build outputs (`target/`, `.lux/`, ad-hoc caches) when users use a glob. Failing the whole workspace load on those would force users to either name every project explicitly (defeating the feature) or maintain ignore patterns. Failing on a non-existent **literal** path is still useful — that's a typo, not a build artifact — so we keep that branch unchanged.

## Tests

Two unit tests reusing the existing `multi-project` fixture:

- `workspace_members_glob_expands_directories` — `projects/*` resolves to both `foo` and `bar`.
- `workspace_members_glob_skips_non_project_directories` — adding a bare `projects/build-cache/` directory is silently dropped; the workspace still loads with the two real projects.

`cargo clippy -p lux-lib --all-targets -- -D warnings` and `cargo fmt --check` clean.

## Dependency

Pulls in `glob = "0.3"` (the standard Rust glob crate, ~50 KB).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3342,6 +3342,7 @@ dependencies = [
  "futures",
  "git-url-parse",
  "git2",
+ "glob",
  "gpgme",
  "hex",
  "html-escape",
--- a/lux-lib/Cargo.toml
+++ b/lux-lib/Cargo.toml
@@ -74,6 +74,7 @@ zip = "8.6"
 
 git-url-parse = { workspace = true }
 git2 = { workspace = true }
+glob = "0.3"
 ignore = { workspace = true }
 indicatif = { workspace = true }
 itertools = { workspace = true }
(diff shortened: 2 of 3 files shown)


mrcjkb:
Thanks for your PR 🙏 

We have [a strict no LLM/no AI policy](https://github.com/lumen-oss/lux/blob/main/CONTRIBUTING.md#strict-no-llm--no-ai-policy).

PR #1554, opened 2026-06-01
feat(pack): support packing individual workspace members

### LINKED ISSUE #1552: [Feature] Support `--package` flag for `lx pack` in workspaces
Currently, running `lx pack` in a workspace packs all packages in the workspace. Adding a `--package` flag would allow packing specific workspace packages.

Another approach could be to prioritize workspace packages if the name is provided on the commandline, such as `lx pack {package}`.

benniekiss:
This PR adds support for packing individual workspace members.

When invoked with a package name argument, `lx pack` will first attempt to find and pack a workspace member of the same name. If there is no matching workspace member, it falls through to searching for installed rocks, then remote rocks, to pack.

To avoid duplicating the workspace packing logic, a function was added to handle packing workspace members. The function packs a specific member if one is provided, otherwise, all workspace members are packed.

The CLI docstring was updated to elaborate on the workspace packing behavior.

Closes #1552

- [X] Fits [CONTRIBUTING.md].
- [X] Fits [No LLM / No AI Policy].

### DIFF
--- a/lux-cli/src/pack.rs
+++ b/lux-cli/src/pack.rs
@@ -1,4 +1,7 @@
-use std::{path::PathBuf, str::FromStr};
+use std::{
+    path::{Path, PathBuf},
+    str::FromStr,
+};
 
 use crate::build;
 use clap::Args;
@@ -11,7 +14,7 @@ use lux_lib::{
     lua_rockspec::RemoteLuaRockspec,
     lua_version::LuaVersion,
     operations::{self, Install, PackageInstallSpec},
-    package::PackageReq,
+    package::{PackageName, PackageReq},
     progress::MultiProgress,
     rockspec::Rockspec as _,
     tree,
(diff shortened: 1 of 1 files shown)


lux-cli/src/pack.rs
```diff
@@ -49,8 +52,9 @@ impl FromStr for PackageOrRockspec {
 #[derive(Args)]
 pub struct Pack {
     /// Path to a RockSpec or a package query for a package to pack.{n}
-    /// Prioritises installed rocks and will install a rock to a temporary{n}
-    /// directory if none is found.{n}
+    /// Prioritises workspace members then installed rocks.{n}
```
mrcjkb: suggestion: 
```suggestion
    /// Prioritises local projects if in a workspace, then installed rocks.{n}
```

lux-cli/src/pack.rs
```diff
@@ -59,18 +63,76 @@ pub struct Pack {
...
     let progress = MultiProgress::new_arc(&config);
-    let rock_paths: Result<Vec<PathBuf>> = match args.package_or_rockspec {
+    let rock_paths: Vec<PathBuf> = match args.package_or_rockspec {
+        // Prioritize packing workspace members if the provided package name matches a member name
+        Some(PackageOrRockspec::Package(package_req))
+            if let Ok(rock_path) =
+                pack_workspace(Some(package_req.name()), &dest_dir, &config).await =>
+        {
+            Ok(rock_path)
+        }
```
mrcjkb: issue: This will fail silently and fall back to packing a local package, if the conditions for packing a workspace are met, but the pack operation itself fails.

This branch should check the conditions for trying to pack a workspace member (e.g. we're in a workspace and there exists a member that matches the requested package), and then call `pack_workspace(...)?`.

mrcjkb (changes requested):
Thanks 🙏

lux-cli/src/pack.rs
benniekiss: Good point! I also realize that packages passed with a specific version should also be checked.

I was thinking of having it pack the workspace member if the version from the CLI matches the current project version, and falling through to installed rocks. But I could also see utility in always searching installed/remote rocks if a version is provided

mrcjkb: I think falling through to installed rocks on version mismatch would be simpler to implement, plus it provides more flexibility.

benniekiss: ptal at the most recent commit. It's a long `is_ok_and` chain to get to the workspace member version, but it adds checks for an existing project and matching version.

lux-cli/src/pack.rs
```diff
@@ -59,18 +63,83 @@ pub struct Pack {
...
 pub async fn pack(args: Pack, config: Config) -> Result<()> {
     let lua_version = LuaVersion::from(&config)?.clone();
     let dest_dir = std::env::current_dir()?;
     let progress = MultiProgress::new_arc(&config);
-    let rock_paths: Result<Vec<PathBuf>> = match args.package_or_rockspec {
+    let rock_paths: Vec<PathBuf> = match args.package_or_rockspec {
+        // Prioritize packing workspace members if the provided
+        // package name and version matches a workspace member
+        Some(PackageOrRockspec::Package(package_req))
+            if Workspace::current_or_err().is_ok_and(|ws| {
```
mrcjkb: issue: `current_or_err().is_ok_and(...` will fail silently if we find a workspace but fail to parse it. I'd use this instead:

```suggestion
            if Workspace::current()?.is_some_and(|ws| {
```

lux-cli/src/pack.rs
```diff
@@ -59,18 +63,83 @@ pub struct Pack {
...
+                ws.select_member(package_req.name()).is_ok_and(|project| {
+                    project
+                        .toml()
+                        .version()
+                        .is_ok_and(|version| package_req.version_req().matches(&version))
+                })
+            }) =>
```
mrcjkb: suggestion: extract this into a `has_matching_workspace_member` helper function (then we don't need the comment above).

mrcjkb (changes requested):

lux-cli/src/pack.rs
benniekiss: done! makes it much cleaner as well

Issue #1568, opened 2026-06-04
[Feature] `lx dist flat-archive` for install tree archives


mrcjkb:
`lx dist --package=<PackageReq> archive --format=<tar|zip|...>`

- `--package` should be optional if in a workspace.
- `--format` should be optional (default is to copy the tree into a path).
- Because flat install trees can't support conflicting dependencies, the command should fail if there are any, perhaps with an optional `--force` flag to bypass it.

(Whatever formats are supported by the zip crate)

See also:

- https://github.com/lumen-oss/lux/discussions/1566#discussion-10194134

PR #1585, opened 2026-06-08
feat: `InstallProject` operation

mrcjkb:
Preparation for #1568.


lux-lib/src/operations/deps.rs
mrcjkb, line 1: todo: 

- [x] committed accidentally. Delete this.

PR #1587, opened 2026-06-10
feat!: `lx dist flat-archive`

### LINKED ISSUE #1568: [Feature] `lx dist flat-archive` for install tree archives

mrcjkb:
Stacked on #1586.

Closes #1568.

```console
Distribute an archive of a flat install tree which includes all dependencies.
 The resulting archive does not include the `etc` directory or build dependencies
 Unlike a Lux tree, dependency conflicts are not supported/handled.

Usage: lx dist flat-archive [OPTIONS] [PACKAGE_OR_ROCKSPEC]

Arguments:
  [PACKAGE_OR_ROCKSPEC]
          Path to a RockSpec or a package query for a package to distribute.
           Prioritises local projects if in a workspace, then installed rocks.
           If there is no matching workspace member or installed rock,
           a rock will be downloaded and installed to a temporary directory.
           In case of multiple matches, the latest version will be distributed.

           Examples:
           - "pkg"
           - "pkg@1.0.0"
           - "pkg>=1.0.0"
           - "/path/to/foo-1.0.0-1.rockspec"

           If not set, lux will attempt to distribute the current project.
           Must be set in multi-project workspaces

Options:
  -d, --destination <DESTINATION>
          The destination path. Defaults to '<cwd>/<package>-<version>.zip'.

          [aliases: -d]

  -c, --compression-method <COMPRESSION_METHOD>
          Possible values:
          - stored:   Store the install tree as is
          - deflated: Compress the install tree using Deflate
          - bzip2:    Compress the install tree using BZIP2
          - xz:       Compress the install tree using XZ
          - zstd:     Compress the install tree using `ZStandard`
          - lzma:     Compress the install tree using LZMA

          [default: stored]
          [aliases: -c]

      --porcelain
          Output a JSON path

  -h, --help
          Print help (see a summary with '-h')
```


mrcjkb:
- [x] Hmm now that I think of it, this should probably be `lx dist flat-archive`.

mrcjkb:
@TaylorRichberger could you please confirm if this matches what you are looking for? :pray:

vhyrro (approved):
LGTM

TaylorRichberger:
Yes, that looks like exactly what I want; that's great!

PR #1609, opened 2026-06-22
feat(fmt): add support for specifying directories and files

### LINKED ISSUE #1408: [Feature] Support for specifying directories for `lx fmt`

vovacha:
Closes #1408.
Adds support for specifying directories and files for `lx fmt`.

Changes:
- Added `PathTarget` and `classify_path` to distinguish between different paths (workspace, loose dir/file). Those might be reused for #1407 and #1409
- Now there are `format_project` and `format_loose`. To prevent duplication, the shared formatting code was extracted into `FmtConfig`, which also handles configs resolution
- Added unit tests and some helpers: `ensure_no_package`, `is_lua_file`

> If formatting requires a Lua version, then I would suggest to fail. Lux tries to auto-detect the Lua version (if it's not set) when building the config.

At the moment it takes the version from `Config::lua_version` as it was suggested. And same as before, it uses luafmt's own default if version is None - `unwrap_or(.syntax.level)`.

- [x] Fits [CONTRIBUTING.md].


lux-cli/src/format.rs
```diff
@@ -36,155 +36,197 @@ enum FmtBackend {
...
+                for project in workspace.members() {
+                    format_project(&args, &workspace, project, &config)?;
+                }
+            }
+        }
+        PathTarget::File(file) => {
+            ensure_no_package(&args)?;
+            if !is_lua_file(&file) {
+                bail!("not a Lua (.lua) file: {}", file.display());
+            }
```
mrcjkb: issue: Not all Lua files have a `.lua` extension. `.rockspec` is an example, and there might be other Lua file extensions we're not aware of.
___

suggestion: I think it's fine to skip the `is_lua_file` check. It's unlikely that any of the formatting backends will successfully format a non-Lua file.
For example, stylua will emit a bunch of `unexpected token` errors if you try to format a markdown file.

lux-cli/src/format.rs
```diff
@@ -36,155 +36,197 @@ enum FmtBackend {
...
+            format_loose(std::iter::once(file), &root, &args.backend, &config)?;
+        }
+        PathTarget::Directory(dir) => {
+            ensure_no_package(&args)?;
+            let files = WalkDir::new(&dir)
+                .into_iter()
+                .filter_map(Result::ok)
+                .filter(|entry| entry.file_type().is_file())
+                .map(|entry| entry.into_path())
+                .filter(|path| is_lua_file(path));
```
mrcjkb: note: We should probably keep this filter though.

lux-cli/src/format.rs
```diff
@@ -36,155 +36,197 @@ enum FmtBackend {
...
+
+    format_files(
+        lua_files.chain(rockspec.exists().then_some(rockspec)),
+        &configs,
+        &args.backend,
+    )
+}
+
+fn is_lua_file(path: &Path) -> bool {
+    path.extension().is_some_and(|ext| ext == "lua")
```
mrcjkb: suggestion:


```suggestion
    path.extension().is_some_and(|ext| ext == "lua" || ext == "rockspec")
```

lux-cli/src/format.rs
```diff
@@ -36,155 +36,197 @@ enum FmtBackend {
...
-    if rockspec.exists() {
-        let unformatted_code = std::fs::read_to_string(&rockspec)?;
-        let formatted_code = match args.backend {
+        Self {
+            stylua,
+            luafmt,
+            luafmt_syntax_level,
+            editorconfig: root.join(".editorconfig"),
+        }
+    }
```
mrcjkb, line 129: issue: If I want to format a subdirectory in my workspace, this will not pick up my workspace's configs.

___

suggestion: Check if `root` is part of a workspace (using `Workspace::from`).
If it is, resolve the configs from the workspace root. Otherwise, fall back to `root`.

mrcjkb (changes requested):
Thanks 🙏

vovacha:
@mrcjkb Thanks for the review. I addressed all the issues you mentioned. Additionally I removed `path_filter` since this logic is now handled by `format_loose`. Can you check if everything looks good now?

mrcjkb (approved):

PR #1626, opened 2026-06-28
feat(config)!: add `luarc_file_name` option

### LINKED ISSUE #1233: [Feature] config option for default generated lua RC file name
Auto add project dependencies added using `lx add <thing>` to the list of libraries in `.emmyrc.json` as is currently done with `.luarc.json`

[Emmylua](https://github.com/EmmyLuaLs/emmylua-analyzer-rust) is a promising, newish (rewritten in rust), lsp for lua. It's configuration file, `.emmyrc.json`, supports the same `workspace.library: []` format that luals's `.luarc.json` does: (https://github.com/EmmyLuaLs/emmylua-analyzer-rust/blob/main/docs/config/emmyrc_json_EN.md#:~:text=rules-,library).

Isvane:
# Summary

Closes #1233

Adds an `ls_config_file` option so users can choose their preferred LSP config filename (like `.emmyrc.json`). Right now, Lux defaults to creating a `.luarc.json` on fresh projects even if the user prefers EmmyLua.

Please let me know if I missed anything, messed up the logic or existing feature, or if there's a better way to approach this!

# Changes
- Added `ls_config_file: Option<String>` to `Config` and `ConfigBuilder`.
- Updated `luarc_path(&self, config: &Config)` to immediately return the custom path if defined, bypassing filesystem checks.
- Fed `&config` / `config` into various `workspace.luarc_path()` calls across the codebase.
- Added unit tests.

- [x] Fits [CONTRIBUTING.md].

### DIFF
--- a/lux-lib/src/config/mod.rs
+++ b/lux-lib/src/config/mod.rs
@@ -53,7 +53,7 @@ pub struct Config {
     user_agent: String,
 
     generate_luarc: bool,
-    ls_config_file: Option<String>,
+    luarc_file_name: String,
     wrap_bin_scripts: bool,
 }
 
@@ -221,9 +221,9 @@ impl Config {
         self.generate_luarc
     }
 
-    /// The custom language server configuration file name, if specified
-    pub fn ls_config_file(&self) -> Option<&String> {
-        self.ls_config_file.as_ref()
+    // Lua runtime configuration file name
+    pub fn luarc_file_name(&self) -> &str {
+        &self.luarc_file_name
     }
 
     /// Whether to wrap installed Lua bin scripts to be executed with
@@ -294,7 +294,7 @@ pub struct ConfigBuilder {
     entrypoint_layout: RockLayoutConfig,
     user_agent: Option<String>,
     generate_luarc: Option<bool>,
-    ls_config_file: Option<String>,
+    luarc_file_name: Option<String>,
     wrap_bin_scripts: Option<bool>,
 }
 
@@ -487,9 +487,11 @@ impl ConfigBuilder {
         }
     }
 
-    pub fn ls_config_file(self, file: Option<String>) -> Self {
+    /// Lua runtime configuration file name
+    /// Default: `.luarc.json`
+    pub fn luarc_file_name(self, file: Option<String>) -> Self {
         Self {
-            ls_config_file: file.or(self.ls_config_file),
+            luarc_file_name: file.or(self.luarc_file_name),
             ..self
         }
     }
@@ -543,7 +545,9 @@ impl ConfigBuilder {
             vendor_dir: self.vendor_dir,
             user_agent: self.user_agent.unwrap_or(DEFAULT_USER_AGENT.into()),
             generate_luarc: self.generate_luarc.unwrap_or(true),
-            ls_config_file: self.ls_config_file,
+            luarc_file_name: self
+                .luarc_file_name
+                .unwrap_or_else(|| ".luarc.json".to_string()),
             wrap_bin_scripts: self.wrap_bin_scripts.unwrap_or(true),
         })
     }
@@ -577,7 +581,7 @@ impl From<Config> for ConfigBuilder {
             entrypoint_layout: value.entrypoint_layout,
             user_agent: Some(value.user_agent),
             generate_luarc: Some(value.generate_luarc),
-            ls_config_file: value.ls_config_file,
+            luarc_file_name: Some(value.luarc_file_name),
             wrap_bin_scripts: Some(value.wrap_bin_scripts),
         }
     }
--- a/lux-lib/src/config/mod.rs
+++ b/lux-lib/src/config/mod.rs
@@ -53,6 +53,7 @@ pub struct Config {
     user_agent: String,
 
     generate_luarc: bool,
+    ls_config_file: Option<String>,
     wrap_bin_scripts: bool,
 }
 
@@ -220,6 +221,11 @@ impl Config {
         self.generate_luarc
     }
 
+    /// The custom language server configuration file name, if specified
+    pub fn ls_config_file(&self) -> Option<&String> {
+        self.ls_config_file.as_ref()
+    }
+
     /// Whether to wrap installed Lua bin scripts to be executed with
     /// the detected or configured Lua installation.
     /// If `true`, individual rocks can still disable wrapping of their own bin scripts.
@@ -288,6 +294,7 @@ pub struct ConfigBuilder {
     entrypoint_layout: RockLayoutConfig,
     user_agent: Option<String>,
     generate_luarc: Option<bool>,
+    ls_config_file: Option<String>,
     wrap_bin_scripts: Option<bool>,
 }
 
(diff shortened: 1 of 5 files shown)


lux-lib/src/config/mod.rs
```diff
@@ -220,6 +221,11 @@ impl Config {
         self.generate_luarc
     }
 
+    /// The custom language server configuration file name, if specified
```
mrcjkb: suggestion:

- It's not only used by language servers. 
- This should not return an `Option` — see my other comment.

```suggestion
    /// Lua runtime configuration file name
```

lux-lib/src/config/mod.rs
```diff
@@ -53,6 +53,7 @@ pub struct Config {
     user_agent: String,
 
     generate_luarc: bool,
+    ls_config_file: Option<String>,
```
mrcjkb: suggestion: Call this `luarc_file_name` instead of `ls_config_file`.

- It's not only used by language servers (`lx check` uses it, for example).
- `_filename` instead of `_file` to clarify that it's not a full file path.

mrcjkb: issue: This shouldn't be an `Option`. Instead, delete the `LUARC` constant in the `workapace` module and in `ConfigBuilder::build`, set this field to `.luarc.json` if the `ConfigBuilder` field is `None`.
That will simplify the logic in the `workspace` module.

lux-lib/src/config/mod.rs
```diff
@@ -480,6 +487,13 @@ impl ConfigBuilder {
         }
     }
 
+    pub fn ls_config_file(self, file: Option<String>) -> Self {
```
mrcjkb: todo: Doc comment: 


```suggestion
    /// Lua runtime configuration file name
    /// Default: `.luarc.json`
    pub fn ls_config_file(self, file: Option<String>) -> Self {
```

lux-lib/src/workspace/mod.rs
```diff
@@ -245,7 +245,11 @@ impl Workspace {
     }
 
     /// Get the `.luarc.json` or `.emmyrc.json` path.
-    pub fn luarc_path(&self) -> PathBuf {
+    pub fn luarc_path(&self, config: &Config) -> PathBuf {
```
mrcjkb, line 247: issue: this is a breaking API change, so the conventional commit prefix should be `feat(config)!:` (with a bang).

lux-lib/src/workspace/mod.rs
```diff
@@ -450,4 +454,54 @@ members = [ "glob:projects/*" ]
...
+
+    #[tokio::test]
+    async fn test_luarc_path_custom_config() {
+        let sample_project: PathBuf = "resources/test/sample-projects/init/".into();
+        let project_root = assert_fs::TempDir::new().unwrap();
+        project_root.copy_from(&sample_project, &["**"]).unwrap();
+
+        let workspace = Workspace::from(&project_root).unwrap().unwrap();
+
+        let mut builder = crate::config::ConfigBuilder::default();
```
mrcjkb: quibble: there's no need to use a `mut ConfigBuilder` to build the config. See our other tests to see how we do this.

lux-lua/src/lua_impls.rs
```diff
@@ -2638,7 +2638,7 @@ impl TypedUserData for WorkspaceLua {
         });
 
         methods.add_method("luarc_path", |_, this, ()| {
-            Ok(this.0.luarc_path().to_slash_lossy().into_owned())
+            Ok(this.0.luarc_path(config).to_slash_lossy().into_owned())
```
mrcjkb: issue: this won't compile. See the other methods that take a `Config`.

mrcjkb (changes requested):
Thanks 🙏

Isvane:
Thank you for the review!

I have updated the PR. Let me know if there's anything else need to be tweaked or if i miss something.

lux-lua/src/lua_impls.rs
```diff
@@ -2638,7 +2638,7 @@ impl TypedUserData for WorkspaceLua {
         });
 
         methods.add_method("luarc_path", |_, this, ()| {
-            Ok(this.0.luarc_path().to_slash_lossy().into_owned())
+            Ok(this.0.luarc_path(&config.0).to_slash_lossy().into_owned())
```
mrcjkb, line 2641: issue: this still won't compile because the method takes `()` as a parameter. It needs to take `config: ConfigLua` (see the other methods).

lux-lib/src/workspace/mod.rs
```diff
@@ -450,4 +451,55 @@ members = [ "glob:projects/*" ]
...
+
+    #[tokio::test]
+    async fn test_luarc_path_fallback_luarc() {
+        use assert_fs::prelude::*;
+
+        let sample_project: PathBuf = "resources/test/sample-projects/init/".into();
+        let project_root = assert_fs::TempDir::new().unwrap();
+        project_root.copy_from(&sample_project, &["**"]).unwrap();
+
+        project_root.child(".luarc.json").touch().unwrap();
```
mrcjkb: suggestion:


```suggestion
        let luarc_file = project_root.child(".luarc.json");
        luarc_file.touch().unwrap();
```

...then change the assertion below to use `luarc_file.to_path_buf()` (I'm on my mobile so I don't remember the exact method for converting a `TempFile` to a `PathBuf`)

lux-lib/src/workspace/mod.rs
```diff
@@ -450,4 +451,55 @@ members = [ "glob:projects/*" ]
...
+
+    #[tokio::test]
+    async fn test_luarc_path_fallback_emmyrc() {
+        use assert_fs::prelude::*;
+
+        let sample_project: PathBuf = "resources/test/sample-projects/init/".into();
+        let project_root = assert_fs::TempDir::new().unwrap();
+        project_root.copy_from(&sample_project, &["**"]).unwrap();
+
+        project_root.child(".emmyrc.json").touch().unwrap();
```
mrcjkb: ditto

lux-lib/src/workspace/mod.rs
```diff
@@ -450,4 +451,55 @@ members = [ "glob:projects/*" ]
...
+
+    #[tokio::test]
+    async fn test_luarc_path_custom_config() {
+        let sample_project: PathBuf = "resources/test/sample-projects/init/".into();
+        let project_root = assert_fs::TempDir::new().unwrap();
+        project_root.copy_from(&sample_project, &["**"]).unwrap();
+
+        let workspace = Workspace::from(&project_root).unwrap().unwrap();
+
+        let config = crate::config::ConfigBuilder::default()
```
mrcjkb: quibble: Add `ConfigBuilder` to the imports of this test module and unqualify this call.

mrcjkb (changes requested):

Isvane:
Ah... yep. Wow. I don't know how I managed to stare right past the `()` signature and completely miss that twice. Sorry about that!

I've finally updated the `luarc_path` method to actually accept `config: ConfigLua`. `ConfigBuilder` is also imported, `assert_fs::prelude::*` too, and the assertions are cleanly using `luarc_file.path().to_path_buf()`.

Sorry for the extra notification noise :sweat_smile:

mrcjkb (approved):

PR #1644, opened 2026-07-01
fix(upload): generate `-1` specrev, not `-2` if package doesn't exist

mrcjkb:


lux.toml
```diff
@@ -1,4 +1,4 @@
-package = "lux-lua"
+package = "lux-lua-test"
```
mrcjkb:
```suggestion
package = "lux-lua"
```

PR #1645, opened 2026-07-02
feat(debug): add `debug toolchains` subcommand

### LINKED ISSUE #1188: [feature] `lx debug` subcommand for available toolchains

Isvane:
# Description

Close #1188 

This try to completes the work started by @cnaples79 in #1189 to address #1188. It introduces a diagnostic subcommand to inspect the host system's build environments, incorporating feedback from the previous review.

Let me know if there's anything I should change or implement.

# Changes

- Eliminated the loose `found: bool` flag. Wrapped found-only properties (`path`, `version`) inside a nested `ToolInfo` struct bound to an `Option<ToolInfo>`.
- Hid all internal fields on `Tool`, `ToolInfo`, and `ToolchainReport` and use public getters instead.
- Use `cc::Build` to resolves the C compiler.
- Moved `OutputFormat` out of `check.rs` and into `args.rs` to be reusable.

- [x] Fits [CONTRIBUTING.md].


lux-cli/src/debug.rs
```diff
@@ -1,8 +1,15 @@
...
     project::DebugProject,
     unpack::{Unpack, UnpackRemote},
 };
-use clap::Subcommand;
+use clap::{Args, Subcommand};
+
+#[derive(Args)]
+pub struct Toolchains {
+    #[arg(long, default_value = "text")]
+    pub format: OutputFormat,
```
mrcjkb: issue: 

- This is inconsistent with the `Check` command.
- The field shouldn't be public (and the struct should probably go in the `debug::toolchains` module - see my other comment).
- Add doc comment for the CLI.


```suggestion
    /// The output format.
    #[arg(long, default_value = "text", value_enum, ignore_case = true)]
    output_format: OutputFormat,
```

lux-cli/src/debug/toolchains.rs
mrcjkb, line 1: todo: This module should probably be in `debug::toolchains`.

mrcjkb (changes requested):
Thanks 🙏 

LGTM, just some small change requests.

Isvane:
Suggestion applied :saluting_face:

vhyrro (approved):
LGTM

mrcjkb (approved):

PR #1662, opened 2026-07-08
feat: generate man pages for all subcommands

mrcjkb:


mrcjkb:
nix build verified locally.

PR #1665, opened 2026-07-12
feat!: improved error messages

vhyrro:
Draft, wouldn't bother reviewing yet.

This PR hotswaps the codebase with `miette`, a library for beautified error messages. Lux's biggest weakness right now is error messages that can be hard to understand or conceptualize. The goal of this PR is twofold:
- [x] Mark every `thiserror` struct with `Diagnostic` for miette support
- [ ] Improve error messages throughout the entire project (this one will take a bit)

Uncertain TODOs:
- Write tests for error cases, to ensure rendering looks as expected?

### DIFF
--- a/lux-cli/src/bin/lx.rs
+++ b/lux-cli/src/bin/lx.rs
@@ -1,7 +1,6 @@
 use std::time::Duration;
 
 use clap::Parser;
-use eyre::Result;
 use lux_cli::{
     add, build, check, completion, config,
     debug::{self, Debug},
@@ -17,13 +16,35 @@ use lux_lib::{
     lockfile::PinnedState::{Pinned, Unpinned},
     lua_version::LuaVersion,
 };
+use miette::{MietteHandlerOpts, Result};
+
+use lux_cli::utils::error::clap_to_miette;
 
 const DEFAULT_USER_AGENT: &str = concat!("lux/", env!("CARGO_PKG_VERSION"));
-
 #[tokio::main(flavor = "multi_thread")]
 async fn main() -> Result<()> {
-    color_eyre::install()?;
-    let cli = Cli::parse();
+    miette::set_hook(Box::new(|_| {
+        Box::new(
+            MietteHandlerOpts::new()
+                .terminal_links(true)
+                .unicode(true)
+                .context_lines(3)
+                .tab_width(4)
+                .break_words(true)
+                .with_cause_chain()
+                .build(),
+        )
+    }))?;
+    let cli = match Cli::try_parse() {
+        Ok(cli) => cli,
+        Err(err) => match clap_to_miette(err) {
+            Ok(report) => return Err(report),
+            Err(text) => {
+                print!("{text}");
+                return Err(miette::miette!("not enough arguments supplied"));
+            }
+        },
+    };
 
     let lua_version = cli.lua_version.or({
         if cli.nvim {
--- a/lux-cli/src/utils/error.rs
+++ b/lux-cli/src/utils/error.rs
@@ -1,0 +1,624 @@
+//! Utility module for converting clap parsing errors into good-looking [`miette`] errors
+use std::{error::Error, ffi::OsStr, path::Path};
+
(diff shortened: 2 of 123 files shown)


vhyrro:
Improving all of the errors will take wayyy too long for a single PR, so let's chunk it up into smaller PRs. This one acts as a baseline:

- [x] All CLI errors are prettified
- [x] TOML parsing errors are vastly improved
- [x] All errors now move through miette's Diagnostic struct with full thiserror compatibility.

I'll undraft after I fix the conflicts.

lux-cli/src/bin/lx.rs
```diff
@@ -17,13 +16,36 @@ use lux_lib::{
...
+                .break_words(true)
+                .with_cause_chain()
+                .build(),
+        )
+    }))?;
+    let cli = match Cli::try_parse() {
+        Ok(cli) => cli,
+        Err(err) => match clap_to_miette(err) {
+            Ok(report) => return Err(report),
+            Err(text) => {
```
vhyrro, line 42: note(non-blocking): I've considered making a custom enum so that we don't have to forward regular text through Err, but I think it's okay as it stands. Thoughts welcome.

lux-cli/src/utils/error.rs
```diff
@@ -0,0 +1,641 @@
...
+    ArgumentConflict {
+        arg: Option<String>,
+        prior: Vec<String>,
+    },
+    InvalidUtf8,
+    Other(String),
+}
+
+impl ErrorContext {
+    fn extract(error: &clap::Error) -> Self {
```
vhyrro, line 64: note: this function has to be a bit ugly because of the format clap gives us data (why an iterator, clap??). I hope that at some point this will get refactored into less of a mess 😂

lux-cli/src/utils/error.rs
```diff
@@ -0,0 +1,641 @@
...
+        .into_iter()
+        .chain(args)
+        .map(|a| shell_escape(&a))
+        .collect::<Vec<_>>()
+        .join(" ")
+}
+
+fn shell_escape(arg: &OsStr) -> String {
+    shlex::try_quote(&arg.to_string_lossy())
+        .unwrap_or_else(|_| unreachable!())
```
vhyrro: note: per shlex's own documentation, the only error try_quote can return is `Nul` for when the arg is completely empty. This isn't a case we'll reach (famous last words).

lux-cli/src/utils/error.rs
```diff
@@ -0,0 +1,616 @@
...
+            }
+            Self::TooManyValues {
+                actual, expected, ..
+            }
+            | Self::TooFewValues {
+                actual, expected, ..
+            }
+            | Self::WrongNumberOfValues {
+                actual, expected, ..
+            } => expected.and_then(|e| actual.map(|a| format!("expected {e} value(s), got {a}"))),
```
mrcjkb: suggestion: format as `value` or `values` depending on whether or not `e == 1`.

lux-cli/src/utils/error.rs
mrcjkb: suggestion: add a comment explaining this, in case it changes in the future:


```suggestion
        .unwrap_or_else(|err| unreachable!("arg cannot be empty. Unexpected error: {:?}", err))
```

lux-cli/src/bin/lx.rs
```diff
@@ -17,13 +16,36 @@ use lux_lib::{
...
+                print!("{text}");
+                return Ok(());
+            }
+        },
```
mrcjkb, line 46: issue: Returning `Ok(())` on `Err` looks suspicious (and it'll result in the wrong exit code).
suggestion: Introduce a simple miette error type for the case in which `clap_to_miette` cannot convert a clap error to a miette error.

Or, return `Err(miette!("{}", text))` (no idea if that compiles)

mrcjkb (changes requested):

mrcjkb (dismissed):
Approved pending successful CI run

PR #1672, opened 2026-07-12
fix(lux-lua): manual `UserData` impls

### LINKED ISSUE #1667: bug(lux-lua): Lockfile lua API is broken since migrating to `mlua-extras`
Since `lux-lua@0.33.4`, I seem to no longer be able to use the lockfile API in lua. Previously, all functions worked, but now, I am receiving errors related to no such attribute on the userdata.

I haven't taken the time to test other parts of the API, but this may extend beyond the lockfile.

Here's a small MRE. It works on `v0.33.4`, but fails on any version after:

```lua
local lux = require "lux"

local config = lux.config.builder()
    :generate_luarc(false)
    :user_tree(user_tree)
    :build()

local tree = config:user_tree("5.5")
local lockfile = tree:lockfile()

print(lockfile:version())
```

Output:

```shell
❯ lx lua test.lua
Welcome to the lux Lua repl for test.
Run `lx lua --help` for options.
To exit type 'exit()' or <C-d>.

/opt/homebrew/bin/lua5.5: runtime error: type does not contain field 'version'
stack traceback:
        [C]: in metamethod 'index'
        test.lua:13: in main chunk
        [C]: in ?
```

Here's the code section:

https://github.com/lumen-oss/lux/blob/bccb4d90c6e6080b5e216ba42eab9f274cc44c64/lux-lua/src/lua_impls.rs#L2459-L2503

mrcjkb:
The derived ones are broken.
Fixes #1667.

### DIFF
--- a/lux-lua/src/config.rs
+++ b/lux-lua/src/config.rs
@@ -141,4 +141,25 @@ mod tests {
         .exec()
         .unwrap();
     }
+
+    #[tokio::test]
+    async fn lua_api_test_tree_lockfile_api() {
+        let temp = assert_fs::TempDir::new().unwrap();
+        let lua = Lua::new();
+        lua.globals().set("user_tree", temp.path()).unwrap();
+        lua.globals().set("lux", crate::lux(&lua).unwrap()).unwrap();
+        lua.load(
+            r#"
+        local config = lux.config
+        local full_config = config.builder()
+            :user_tree(user_tree)
+            :build()
+        local tree = full_config:user_tree("5.5")
+        local lockfile = tree:lockfile()
+        print(lockfile:version())
+    "#,
+        )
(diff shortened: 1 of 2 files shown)


lux-lua/src/config.rs
```diff
@@ -141,4 +141,25 @@ mod tests {
         .exec()
         .unwrap();
     }
+
+    #[tokio::test]
+    async fn lua_api_test_tree_lockfile_api() {
+        let temp = assert_fs::TempDir::new().unwrap();
```
mrcjkb, line 147: todo: We'll probably want more tests like these so that we don't run into regressions (for another time)

mrcjkb: #1673

PR #1675, opened 2026-07-12
test(lux-lua): add Lua API tests for `UserData` types

### LINKED ISSUE #1673: [Tests] Test the Lua API of each lux-lua `UserData` instance.
When we switched to mlua-extras, our derived` mlua_extras::UserData` instances broke (#1667).
As a hotfix, I have added manual `mlua::UserData` instances and added a single test to confirm it works in #1672.
To prevent something like this from going unnoticed, we should add a small Lua API test for each type that we have a `UserData` instance for.

mrcjkb:
Closes #1673

### DIFF
--- a/lux-lua/src/lua_impls.rs
+++ b/lux-lua/src/lua_impls.rs
@@ -3330,3 +3330,893 @@ enum PackageInstallSpecInput {
         build_behaviour: BuildBehaviourLua,
     },
 }
+
+#[cfg(test)]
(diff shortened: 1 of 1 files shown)


mrcjkb (commented):
lua-tests need a `writableTmpDirAsHomeHook` in the `nativeCheckInputs`.
Or, the tests should be rewritten not too need `$HOME` access.

PR #1693, opened 2026-07-15
feat: allow dispatching logging to lua

vhyrro:
In `lux.nvim` we need the ability to silence indicatif spinners and to dispatch the progress messages elsewhere (specifically to a sink on the Lua side which captures the messages and lets us drain them to display them).

This means the progress messages will arrive in chunks, but will arrive fairly frequently due to the high traffic of `await` calls.

### DIFF
--- a/lux-lib/src/operations/exec.rs
+++ b/lux-lib/src/operations/exec.rs
@@ -126,11 +126,9 @@ async fn exec(run: Exec<'_>) -> Result<(), ExecError> {
     let lua_init = if run.disable_loader.unwrap_or(false) {
         None
     } else if user_tree.version().lux_lib_dir().is_none() {
-        eprintln!(
-            "⚠️ WARNING: lux-lua library not found.
-    Cannot use the `lux.loader`.
-    To suppress this warning, set the `--no-loader` option.
-                    "
+        crate::logging::warn(
+            "⚠️ WARNING: lux-lua library not found.\n    Cannot use the `lux.loader`.\n    To suppress this warning, set the `--no-loader` option.".to_string(),
+            Some("exec".to_string()),
         );
         None
     } else {
--- a/lux-lib/src/build/utils.rs
+++ b/lux-lib/src/build/utils.rs
@@ -509,10 +509,16 @@ async fn link_c_artifacts(
     };
     if config.verbose() {
         if !&output.stdout.is_empty() {
-            println!("{}", String::from_utf8_lossy(&output.stdout));
+            crate::logging::info(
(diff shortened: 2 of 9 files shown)


mrcjkb:
question: wouldn't this be a good opportunity to introduce [tracing](https://crates.io/crates/tracing)?

note: the indicatif stuff should probably be a separate "progress reporting" API.

vhyrro:
> question: wouldn't this be a good opportunity to introduce [tracing](https://crates.io/crates/tracing)?
> 
> note: the indicatif stuff should probably be a separate "progress reporting" API.

absolutely, but I think we should delegate that for a separate PR. adding proper logging throughout lux will take just as much as implementing beautified error messages. adding tracing just to replace ~5 `println!` and `eprintln!` statements in this PR doesn't make sense imo

Also: the faster I can push this PR through, the faster we can get `lux.nvim`. Side quests will slow us down :sweat_smile:

vhyrro:
This PR implements a static sink for pushing error messages which can be read off by an API consumer (in our case lux-lua). This is the simplest method:
1. Realtime message passing requires enabling the mlua `send` feature which wreaks havoc
2. Wrapping this in a struct would mean passing that struct around for the entire program's lifetime
3. By having a drain feature wrapped in a queue it gives us a nice interface for using this in lua (set up a timer which drains the message log and displays it in fidget.nvim)

lux-lib/src/operations/exec.rs
```diff
@@ -134,11 +126,9 @@ async fn exec(run: Exec<'_>) -> Result<(), ExecError> {
     let lua_init = if run.disable_loader.unwrap_or(false) {
         None
     } else if user_tree.version().lux_lib_dir().is_none() {
-        eprintln!(
-            "⚠️ WARNING: lux-lua library not found.
-    Cannot use the `lux.loader`.
-    To suppress this warning, set the `--no-loader` option.
-                    "
+        crate::logging::warn(
+            "⚠️ WARNING: lux-lua library not found.\n    Cannot use the `lux.loader`.\n    To suppress this warning, set the `--no-loader` option.".to_string(),
```
mrcjkb, line 130: suggestion: Perhaps the `⚠️ WARNING: ` prefix should be added by the `logging` module implementation?

mrcjkb (approved):

lux-lib/src/operations/exec.rs
vhyrro: I wanted to implement this but let's leave it as it is. There are multiple parts of the codebase that use `bar.println("WARNING: ")` as well and that has no notion of what a "warning" is. I've made a tracking issue: #1697

PR #1694, opened 2026-07-15
feat(error-reporting)!: more diagnostics

mrcjkb:
- build module
- manifest
- project toml

### DIFF
--- a/lux-lib/src/build/cmake.rs
+++ b/lux-lib/src/build/cmake.rs
@@ -39,7 +39,7 @@ pub enum CMakeError {
     #[error("failed to run `cmake` step: {0}")]
     Io(io::Error),
     #[error("failed to write CMakeLists.txt: {0}")]
-    WriteCmakeListsError(io::Error),
+    WriteCmakeLists(io::Error),
     #[error("failed to run `cmake` step: '{0}' command not found!")]
     #[diagnostic(help("run `lx debug toolchains` to check available build tools."))]
     CommandNotFound(String),
@@ -84,7 +84,7 @@ impl BuildBackend for CMakeBuildSpec {
         let mut args = Vec::new();
         if let Some(content) = self.cmake_lists_content {
             let cmakelists = build_dir.join("CMakeLists.txt");
-            std::fs::write(&cmakelists, content).map_err(CMakeError::WriteCmakeListsError)?;
+            std::fs::write(&cmakelists, content).map_err(CMakeError::WriteCmakeLists)?;
             args.push(format!("-G\"{}\"", cmakelists.display()));
         } else if cfg!(all(target_os = "windows", target_arch = "x86_64")) {
             // With msvc and x64, CMake does not select it by default so we need to be explicit.
--- a/lux-lib/src/build/cmake.rs
+++ b/lux-lib/src/build/cmake.rs
@@ -41,10 +41,11 @@ pub enum CMakeError {
     #[error("failed to write CMakeLists.txt: {0}")]
     WriteCmakeListsError(io::Error),
     #[error("failed to run `cmake` step: '{0}' command not found!")]
+    #[diagnostic(help("run `lx debug toolchains` to check available build tools."))]
     CommandNotFound(String),
     #[error(transparent)]
     #[diagnostic(transparent)]
-    VariableSubstitutionError(#[from] VariableSubstitutionError),
+    VariableSubstitution(#[from] VariableSubstitutionError),
 }
 
 struct CMakeVariables;
--- a/lux-lib/src/build/command.rs
+++ b/lux-lib/src/build/command.rs
@@ -31,8 +31,12 @@ pub enum CommandError {
     #[diagnostic(transparent)]
     Paths(#[from] PathsError),
     #[error("'build_command' and 'install_command' cannot be empty.")]
+    #[diagnostic(help(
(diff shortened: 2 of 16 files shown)


lux-lib/src/build/cmake.rs
```diff
@@ -41,10 +41,11 @@ pub enum CMakeError {
     #[error("failed to write CMakeLists.txt: {0}")]
     WriteCmakeListsError(io::Error),
```
vhyrro: suggestion: also remove the `Error` suffix here so that the entire enum is consistent?
```suggestion
    WriteCmakeLists(io::Error),
```

lux-lib/src/build/command.rs
```diff
@@ -31,8 +31,12 @@ pub enum CommandError {
     #[diagnostic(transparent)]
     Paths(#[from] PathsError),
     #[error("'build_command' and 'install_command' cannot be empty.")]
+    #[diagnostic(help(
+        "set a command in the `build.build_command` or `build.install_command` field."
+    ))]
```
vhyrro: suggestion: miette allows setting a `url()` field. Perhaps we could point this url to:
- The luarocks documentation for these fields
or
- A `docs.rs` link for lux-lib which explains this?

lux-lib/src/build/command.rs
```diff
@@ -31,8 +31,12 @@ pub enum CommandError {
...
     EmptyCommand,
     #[error("error parsing command:\n{command}\n\nerror: {err}")]
+    #[diagnostic(help("quote arguments containing spaces in the build/install command."))]
```
vhyrro: issue: I don't think this is easy enough to understand. Perhaps `put arguments containing spaces in quotes in the build/install command.`? I had to reread that twice as a native speaker to get it 😅

lux-lib/src/build/mod.rs
```diff
@@ -159,8 +159,18 @@ pub enum BuildError {
...
-    #[error("source integrity mismatch.\nExpected: {expected},\nbut got: {actual}")]
+    #[error(
+        r#"source integrity mismatch.
+- source: {src}
+- expected: {expected}
+- got: {actual}"#
+    )]
+    #[diagnostic(help(
+        r#"the source may have been modified or a tag may have been moved.
+check the source, then use `--no-lock` to update the hash."#
```
vhyrro: suggestion: maybe we can be a bit more informative here? Instead of just saying "use --no-lock", maybe we can give a full command? `lx --no-lock update --toml` perhaps?

lux-lib/src/lockfile/mod.rs
```diff
@@ -802,7 +802,7 @@ pub enum LockfileIntegrityError {
     )]
     #[diagnostic(help(
         r#"The maintainer may have force-uploaded a different rockspec.
```
vhyrro: same suggestion as above

lux-lib/src/lockfile/mod.rs
```diff
@@ -818,7 +818,7 @@ Check the rockspec source, then use `--no-lock` to update the hash.
     )]
     #[diagnostic(help(
         r#"The maintainer may have force-uploaded a different source or moved a tag.
-Check the source URL, then use `--no-lock` to update the hash.
+check the source URL, then use `--no-lock` to update the hash.
```
vhyrro: ditto

lux-lib/src/manifest/metadata_from_server.rs
```diff
@@ -20,8 +20,13 @@ pub enum ManifestFromServerError {
     #[error(transparent)]
     Io(#[from] io::Error),
     #[error("failed to pull manifest:\n{0}")]
+    #[diagnostic(help(
+        r#"check your network connection and server configuration.
+if the issue persists, the server may be temporarily unavailable."#
+    ))]
     Request(#[from] reqwest::Error),
     #[error("failed to parse manifest:\n{0}")]
+    #[diagnostic(help("the server returned a manifest that is not valid UTF-8."))]
```
vhyrro: suggestion: maybe mention that this may not be the user's fault but that they should check their internet connection?

lux-lib/src/operations/build_lua.rs
```diff
@@ -64,8 +64,18 @@ pub enum BuildLuaError {
...
-    #[error("source integrity mismatch.\nExpected: {expected},\nbut got: {actual}")]
+    #[error(
+        r#"source integrity mismatch.
+- source: {src}
+- expected: {expected}
+- got: {actual}"#
+    )]
+    #[diagnostic(help(
+        r#"the source may have been modified or a tag may have been moved.
+check the source, then use `--no-lock` to update the hash."#
```
vhyrro: same suggestion as above to give more useful info than just `--no-lock`

vhyrro (changes requested):

lux-lib/src/build/mod.rs
mrcjkb: The problem is that we don't know which subcommand is being run.

vhyrro: how hard would it be to trace the subcommand being issued? if it's very hard, then maybe reword the error to `check the source, then rerun the command with --no-lock to update the hash`?

vhyrro (approved):
🎉

PR #1699, opened 2026-07-16
refactor!: unify logging

max397574:
Ref: https://github.com/lumen-oss/lux/issues/1697

For now I just did put a log level parameter into the progress.println function. Instead of introducing a new `Severity` enum, I used the existing `crate::logging::LogLevel` enum. Therefore just 2./3. is addressed.

Is there a particular reason why we'd want to go through `logging::warn` etc instead of just do as I did?

By reusing the existing enum I think the following is no longer relevant
> Refactor logging:warn(), logging::error() etc to either use the enum Severity or to unify with the progress_bar.println() function.

TODO (separate PR?):
- [ ] Change CLI to no longer use `[e]println!`, tracing instead if I understood correctly, haven't dug into this yet

- [x] Fits [CONTRIBUTING.md].

### DIFF
--- a/lux-lib/src/operations/build_lua.rs
+++ b/lux-lib/src/operations/build_lua.rs
@@ -489,14 +489,14 @@ async fn do_build_lua_unix(
         Ok(output) if !output.status.success() && build_target != "generic" => {
             progress.map(|p| {
                 p.println(format!(
-                    r#"⚠️ WARNING: Could not build Lua target '{build_target}'. Attempting a 'generic' Lua target build.
+                    r#"Could not build Lua target '{build_target}'. Attempting a 'generic' Lua target build.
 Some functionality may be limited, including dynamic module loading and REPL history.
 
 stderr:
 {}
 "#,
                     String::from_utf8_lossy(&output.stderr)
-                ))
+                ),crate::logging::LogLevel::Warn)
             });
             let fallback_output = Command::new(config.make_cmd())
                 .current_dir(build_dir)
--- a/lux-lib/src/progress.rs
+++ b/lux-lib/src/progress.rs
@@ -1,6 +1,9 @@
 use std::{borrow::Cow, sync::Arc, time::Duration};
 
-use crate::{config::Config, logging::LoggingState};
+use crate::{
+    config::Config,
+    logging::{LogLevel, LoggingState},
+};
 
 mod private {
     pub trait HasProgress {}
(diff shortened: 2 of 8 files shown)


vhyrro:
changes look good to me, yes let's reuse the enum :+1: 

we go through `warn` and the other functions because not every warning or error comes from a progress bar - that's for those cases. perhaps drop `tracing` for this PR - I think we could break this up into two on further thought :)

max397574:
So would there be anything left to do here?

Otherwise this would be ready for review ig.

lux-lib/src/operations/build_lua.rs
```diff
@@ -478,14 +478,14 @@ async fn do_build_lua_unix(
         Ok(output) if !output.status.success() && build_target != "generic" => {
             progress.map(|p| {
                 p.println(format!(
-                    r#"⚠️ WARNING: Could not build Lua target '{build_target}'. Attempting a 'generic' Lua target build.
```
max397574, line 492: I removed this because it is represented with the level now.

The way I understood it, we just collect logs in like a queue which then is read out, I assumed whoever then does something with these log entries should prepend such a text if they want it.

mrcjkb: issue: If we remove this here, we probably want to add it as a prefix in the `crate::logging::LoggingState::Disabled` case in the `progress` module.
We could also add a prefix for error messages.

lux-lib/Cargo.toml
```diff
@@ -92,6 +92,8 @@ tokio = { workspace = true }
 walkdir = { workspace = true }
 which = { workspace = true }
 zip = { workspace = true }
+tracing = { workspace = true }
```
mrcjkb: issue: looks like this isn't needed yet

mrcjkb (changes requested):

lux-lib/Cargo.toml
max397574: removed

lux-lib/src/operations/build_lua.rs
max397574: added prefix (also for error)
For info I didn't add a prefix, let me know if I should also add one there

max397574:
I realized that the `finish_with_message` function just below is really similar, it also hardcodes the level

Should I make the same changes there?

lux-lib/src/progress.rs
```diff
@@ -128,20 +128,28 @@ impl ProgressBar {
         self.0.position()
     }
 
-    pub fn println<M>(&self, message: M)
+    pub fn println<M>(&self, message: M, level: crate::logging::LogLevel)
```
mrcjkb: note: this is a breaking change as it changes a public function signature (added a `!` to the PR title prefix).

mrcjkb:
> I realized that the `finish_with_message` function just below is really similar, it also hardcodes the level
> 
> Should I make the same changes there?

Sure, why not?

mrcjkb (dismissed):

lux-lib/src/progress.rs
max397574: will change the commit message accordingly as well

mrcjkb (approved):

PR #1702, opened 2026-07-16
chore(ci): fix release-plz failing for lux-macros

vhyrro:


vhyrro:
trivial, so merging right away

PR #1703, opened 2026-07-16
fix(logging): also route MultiProgress output through the sink

vhyrro:
quick fix for lux-lua


vhyrro:
another case where I'll merge because it's a trivial fix.

PR #1706, opened 2026-07-17
feat(error-reporting)!: config + rockspec diagnostics

mrcjkb:
This PR adds some more error diagnostics, focused on config and lua rockspecs.

Lua rockspecs used to render horribly, because ottavino wraps Lux errors in a `ExternError` within `lua.try_enter`. That resulted in 3 copies of the entire rockspec content being printed.
With miette's `source_code`, we render a snippet of the rockspec content only once and add labels pointing to the package name and the field that could not be parsed:

image

### DIFF
--- a/lux-lib/src/config/mod.rs
+++ b/lux-lib/src/config/mod.rs
@@ -111,9 +111,22 @@ impl Config {
     pub fn enabled_dev_servers(&self) -> Result<Vec<Url>, ConfigError> {
         let mut enabled_dev_servers = Vec::new();
         if self.enable_development_packages {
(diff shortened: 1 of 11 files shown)


lux-lib/resources/test/luv-1.48.0-2.rockspec
```diff
@@ -10,7 +10,7 @@ dependencies = {
 }
 
 build = {
-  type = 'cmake',
+  type = 5,
```
mrcjkb: issue: accidentally committed this

PR #1709, opened 2026-07-17
refactor!: replace progress bars with `tracing`

vhyrro:


mrcjkb:
rebased on `main` to resolve conflicts (and renamed the second commit because my config disallows pushing commits with a `wip:` prefix).

lux-lib/src/build/utils.rs
```diff
@@ -537,16 +537,10 @@ async fn link_c_artifacts(
...
-            );
+            tracing::info!("{}", String::from_utf8_lossy(&output.stdout));
         }
         if !&output.stderr.is_empty() {
-            crate::logging::warn(
-                String::from_utf8_lossy(&output.stderr).to_string(),
-                Some("build".to_string()),
-            );
+            tracing::warn!("{}", String::from_utf8_lossy(&output.stderr));
         }
```
mrcjkb: suggestion: We no longer need to check `if config.verbose()` if we're using tracing. lux-cli can use `config.verbose()` to se the tracing level to output (I've added that in 37eae42079155ef34bc8a56b427769dcd3a6ef70)

lux-lib/src/build/utils.rs
```diff
@@ -642,16 +636,10 @@ pub(crate) async fn install_binary(
...
+            tracing::info!("{}", String::from_utf8_lossy(&output.stderr));
         }
         if !output.stdout.is_empty() {
-            crate::logging::info(
-                String::from_utf8_lossy(&output.stdout).to_string(),
-                Some("build".to_string()),
-            );
+            tracing::info!("{}", String::from_utf8_lossy(&output.stdout));
         }
     }
```
mrcjkb: ditto

lux-lib/src/build/mod.rs
```diff
@@ -232,46 +226,22 @@ async fn install<R: Rockspec + HasIntegrity, T: InstallTree>(
...
+    tracing::info!(message = "📋 Copying Lua modules...");
     for (target, source) in &install_spec.lua {
         let absolute_source = build_dir.join(source);
         utils::copy_lua_to_module_path(&absolute_source, target, &output_paths.src)?;
-        progress.map(|p| p.set_position(p.position() + 1));
-    }
-    if lib_len > 0 {
-        progress.map(|p| p.set_message("📋 Compiling C libraries..."));
     }
+    tracing::info!(message = "📋 Compiling C libraries...");
```
mrcjkb: issue: `INFO` is the baseline log level. I don't think libraries should log at this level, as it can pollute the output of packages that use this lib.

suggestion: Let's make these child spans.

lux-lib/src/build/rust_mlua.rs
```diff
@@ -86,7 +84,6 @@ impl BuildBackend for RustMluaBuildSpec {
         build_args.push("--features");
         build_args.push(&features);
         build_args.extend(self.cargo_extra_args.iter().map(|arg| arg.as_str()));
-        progress.map(|bar| bar.set_message("🛠️ Building (🦀 rust-mlua)..."));
```
mrcjkb, line 89: suggestion: I think it's a good idea to have some child spans here, so we know which build backend is active.

mrcjkb (changes requested):

mrcjkb (dismissed):

mrcjkb:
I think this is good to merge.
If we do, let's rebase so all the changes don't get lost in the changelog.

vhyrro:
has the issue with the duplicate spinners been resolved properly?

lux-cli/src/install_lua.rs
```diff
@@ -1,41 +1,18 @@
...
     // TODO: Detect when path already exists by checking `Lua::path()` and prompt the user
     // whether they'd like to forcefully reinstall.
-    let lua = LuaInstallation::install(version_stringified, &config, &bar).await?;
-    let lua_root = lua
+    let lua = LuaInstallation::install(version_stringified, &config).await?;
+    let _lua_root = lua
         .includes()
         .first()
         .and_then(|dir| dir.parent())
         .ok_or_else(|| miette!("error getting lua include parent directory"))?;
```
vhyrro: question: will we want to readd this back? if not, we can remove `lua_root` entirely

lux-cli/src/purge.rs
```diff
@@ -19,15 +15,8 @@ pub async fn purge(config: Config) -> Result<()> {
             .prompt()
             .into_diagnostic()?
     {
-        let root_dir = tree.root();
+        let _root_dir = tree.root();
```
vhyrro: ditto

lux-cli/src/install_lua.rs
mrcjkb: Yes, but as an info trace.

lux-cli/src/purge.rs
mrcjkb: Yes, but as a span

mrcjkb:
> has the issue with the duplicate spinners been resolved properly?

Pretty much. `luasystem` and `fzy` no longer produce duplicate spinners on my machine. `busted` sometimes does and sometimes doesn't (when it does, it's only 2-3 instead of dozens).
It's much better than what we had before though.

PR #1718, opened 2026-07-20
feat(tracing): more trace spans

### LINKED ISSUE #920: make --verbose more verbose
follow up of https://github.com/nvim-neorocks/lux/pull/876 / https://github.com/nvim-neorocks/lux/issues/854

I've not been able to install anything with lux yet because I hit: 

```
./result/bin/lx --lua-version 5.1 --verbose install alogger
Error: failed to build alogger: error building Lua from source:
Cannot read '/nix/store/eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee-vendor-cargo-deps/c19b7c6f923b580ac259164a89f2577984ad5ab09ee9d583b888f934adbbe8d0/lua-src-548.1.1/lua-5.1.5': No such file or directory (os error 2)
```

I am in a nix-shell with pkg-config and a lua5.1 interpreter. I have been a bit short on time to look at the source yet but if lux could output the reasoning and the commands it used to reach that conclusion it would help me find a workaround. There should be no need to build a lua interpreter since I have one so there might be some bad interaction with pkg-config and my setup.
Maybe --verbose is meant only for user-messages ? in which case I would be looking for a --verbose=DEBUG / --log-level=DEBUG ?

mrcjkb:
Add a bunch of trace spans to lux-lib, mostly for IO actions.

Closes #920.

To get trace output, run:

```
RUST_LOG=trace lx ...
```

`lx --verbose ...` runs with debug logs enabled.

### DIFF
--- a/lux-cli/src/lib.rs
+++ b/lux-cli/src/lib.rs
@@ -134,7 +134,8 @@ pub struct Cli {
     #[arg(long, value_name = "variable", visible_short_alias = 'v', value_parser = parse_key_val::<String, String>)]
     pub variables: Option<Vec<(String, String)>>,
 
-    /// Display verbose output of commands executed.
+    /// Display verbose output of commands executed, enabling DEBUG logs.{n}
+    /// To enable TRACE logs, set RUST_LOG=trace.
     #[arg(long)]
     pub verbose: bool,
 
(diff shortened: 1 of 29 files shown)


mrcjkb:
Merging, as this is a trivial change that doesn't add any business logic.

PR #1731, opened 2026-07-22
feat(ui): slow down moon spinner animation

### LINKED ISSUE #1729: [Feature] Reduce speed of CLI moon emoji spinners
I really like the new moon emoji spinners! But currently, they spin too fast, and it actually makes me a little queasy to look at them in the terminal for too long

The spinners can be set to a fixed rate using [`with_tick_settings`](https://docs.rs/tracing-indicatif/latest/tracing_indicatif/struct.IndicatifLayer.html#method.with_tick_settings), and I would suggest doubling the default tick rate to 200ms (default is 100ms)

```rust
.with_tick_settings(tracing_indicatif::TickSettings {
    term_draw_hz: 20,
    default_tick_interval: Some(Duration::from_millis(200)),
    footer_tick_interval: Some(Duration::from_millis(200)),
    ..Default::default()
});
```

mrcjkb:
Closes #1729.

I tried a tick interval of 200 ms, but it felt a bit too choppy. 165 ms seems to be the sweet spot.

### DIFF
--- a/lux-cli/src/bin/lx.rs
+++ b/lux-cli/src/bin/lx.rs
@@ -117,13 +117,20 @@ async fn main() -> Result<()> {
     if config.no_progress() || !std::io::stderr().is_terminal() {
         tracing_subscriber::registry().with(fmt_layer).init();
     } else {
-        let indicatif_layer = progress::IndicatifLayer::new().with_progress_style(
-            indicatif::ProgressStyle::with_template(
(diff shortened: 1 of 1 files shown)


mrcjkb:
@benniekiss could you please test this? :pray:

benniekiss:
This speed looks great! Thank you

PR #1737, opened 2026-07-26
feat: default `--lua-version` to project's exact Lua constraint

### LINKED ISSUE #1170: [Feature] `--lua-version` should default to project Lua version if specified with `==` constraint
Given:

- Lua 5.2 on PATH
- `lua = "== 5.1"` in lux.toml

Expected behaviour:

- `--lua-version` should default to `5.1`

Actual behaviour:

- `--lua-version` defaults to `5.2` because of detected Lua 5.2
- This results in an error: "Lua 5.2

dhmztr:
- [x] Fits [CONTRIBUTING.md].

### DIFF
--- a/lux-cli/src/bin/lx.rs
+++ b/lux-cli/src/bin/lx.rs
@@ -1,7 +1,6 @@
 use std::io::IsTerminal;
 
 use std::time::Duration;
-use strum::IntoEnumIterator;
 
 use clap::Parser;
 use lux_cli::{
@@ -18,8 +17,6 @@ use lux_lib::{
     config::{tree::RockLayoutConfig, ConfigBuilder},
     lockfile::PinnedState::{Pinned, Unpinned},
     lua_version::LuaVersion,
-    project::project_toml::PartialProjectToml,
-    workspace::Workspace,
 };
 
 use miette::{IntoDiagnostic, MietteHandlerOpts, Result};
@@ -71,12 +68,7 @@ async fn main() -> Result<()> {
                 None
             }
         })
-        .or_else(|| {
-            let subcommand = cli.command.project_package()?;
-            let current_workspace = Workspace::current().ok().flatten()?;
-            let item = current_workspace.single_member_or_select(subcommand).ok()?;
-            exact_lua_version(item.toml())
-        });
+        .or_else(|| cli.command.detect_lua_version());
 
     let mut config_builder = ConfigBuilder::new()?
         .dev(Some(cli.dev))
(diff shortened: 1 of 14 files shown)


mrcjkb (changes requested):
Thanks for your PR 🙏

And sorry for the GitHub copilot noise - I had disabled that but apparently GitHub has re-enabled it and I can't find a setting to disable it anymore 😒

We have an open issue for this: #1170.

The problem with the approach in this PR is that not all commands are project-specific. For instance, `lx install` should never pick up the Lua version from a project in the `cwd`.
An option would be to add something like this to the [`Commands` enum](https://github.com/dhmztr/lux/blob/fba84277411fb863a7b269a2132deabd8d70ba7e/lux-cli/src/lib.rs#L183):

```rust
impl Commands {
  pub fn detect_lua_version(&self) -> Result<Option<LuaVersion>> {
    match self {
      // split on project/non-project commands
    }
  }
}
```

Regarding the handling of multiple projects in a workspace: Most project-specific commands have an optional `--package` flag, which can be used to determine the project to get the Lua version from. 

I wouldn't add a public `exact_lua_version` method to the `PartialProjectToml` impl. It's only used in one place, so it should be implemented as an internal helper function in `lux-cli/lib.rs`.

lux-cli/src/bin/lx.rs
```diff
@@ -57,13 +62,21 @@ async fn main() -> Result<()> {
...
+        .or({
+            if cli.nvim {
+                Some(LuaVersion::Lua51)
+            } else {
+                None
+            }
+        })
+        .or_else(|| {
+            let subcommand = cli.command.project_package()?;
+            let current_workspace = Workspace::current().ok().flatten()?;
```
mrcjkb: suggestion: Extract this logic into a `Command::lua_version` function.

issue: `Command::lua_version` can't rely solely on the presence of a `--package` flag. Some commands (e.g. `sync`, `vendor`, ...) operate on projects without one. `lx dist flat_archive` has a `PackageOrRockspec`, which can specify a project package if in a workspace, or a remote package otherwise. 

note: this is non-blocking. I'm okay with merging this as is and making the changes myself, if you'd prefer.

lux-cli/src/lib.rs
```diff
@@ -350,6 +350,27 @@ pub enum Commands {
...
+            Self::Pin(p) | Self::Unpin(p) => Some(&p.package),
+            Self::Remove(r) => Some(&r.package),
+            Self::Test(t) => Some(&t.package),
+            Self::Update(u) => Some(&u.package),
+            Self::Run(r) => Some(&r.build.package),
+            Self::Dist(b) => match b {
+                Dist::Bin(b) => Some(&b.package),
+                Dist::FlatArchive(_) => None,
+            },
+            _ => None,
```
mrcjkb: suggestion: it's best to do an explicit pattern match on each subcommand without a catch-all, so that we're forced to update this if we add a new subcommand in the future.

mrcjkb (changes requested):
todo: clippy checks are failing

mrcjkb (approved):

Issue #23, opened 2024-06-22
[Feature] `rocks --lsp` command for `rocks.toml`


mrcjkb:
- We already have all the logic needed to query luarocks,org (except for custom servers like rocks-binaries) and parse versions, etc.
- This could be used by rocks.nvim to provide
  - Diagnostics about outdated or out-of-sync[^1] rocks
  - Code actions for editing `rocks.toml`
  - Completions for rocks and versions

[^1]: Would need the ability to query installed rocks - which has a different tree structure than rocks when installed with luarocks.

- Not entirely sure if `rocks` is the right tool for this, but I think it would have better performance being in a separate process than within Neovim.
- Would we want a separate `rocks-lsp-lib` crate, to keep it separate from `rocks-lib`?

vhyrro:
If the LSP is "lightweight enough" (which I can imagine it being, since it's just reading a TOML file and providing basic diagnostics by using the library), we could happily make a `rocks lsp` subcommand that would launch the LSP.

We could gate this behind a feature flag for convenience, if someone really doesn't want the subcommand nor its dependencies :)

PR #1746, opened 2026-07-28
feat: lux-lsp

vhyrro:
This PR adds a new `lux-lsp` binary to the workspace tree.

We need it for the following reasons:
1. `lux.nvim` needs a nice way of displaying diagnostics
2. But to do so it'd need to create its own `tracing` subscriber, create a unified lua table structure for these error messages and then convert these messages to `fidget.nvim`-compatible messages in `lux.nvim`

Solution:
1. Create an LSP module - we've been meaning to do this anyway because we wanted to provide diagnostics for lux.toml #23 
2. The LSP attaches itself to a workspace
3. On the `lux-lua` side, we call `require("lux").progress.set_connection(workspace)`
4. All `tracing` calls get forwarded to the LSP
5. The LSP then emits these `tracing` calls as LSP-compatible progress messages

This is exceptionally useful as `fidget.nvim` natively intercepts LSP progress messages and displays them, meaning we get working diagnostics almost for free.


lux-lib/src/progress/client.rs
```diff
@@ -0,0 +1,63 @@
...
+            let _ = serde_json::to_writer(&mut *w, msg);
+            let _ = writeln!(w);
+            let _ = w.flush();
+        }
+    }
+}
+
+#[derive(Debug, Diagnostic, Error)]
+pub(crate) enum ConnectError {
+    #[error("failed to read LSP port file: {0}")]
```
mrcjkb: suggestion: `: {0}` is only necessary if the source error isn't set with `source` or `from`.


```suggestion
    #[error("failed to read LSP port file")]
```

suggestion: It would be useful to include the port file path

(ditto for `ParsePort`)

lux-lib/src/progress/mod.rs
```diff
@@ -0,0 +1,25 @@
...
+
+pub fn set_connection(workspace: &Workspace) {
+    match LspClient::connect(workspace) {
+        Ok(client) => {
+            if let Ok(mut guard) = CLIENT.write() {
+                *guard = Some(Arc::new(client));
+            }
+        }
+        Err(_) => {
+            tracing::trace!("no lx-lsp server found; LSP progress disabled");
```
mrcjkb: suggestion: include the error in the trace.

suggestion: this can probably be debug level

lux-lsp/src/main.rs
```diff
@@ -0,0 +1,209 @@
+use std::borrow::Cow;
+use std::io::Write;
+use std::sync::Mutex;
+
+use eyre::{Context, Result};
```
mrcjkb, line 5: suggestion: Why not miette?

vhyrro: I found it makes little sense to use it. The `tower-lsp` produces its own colourful error messages in stderr and users don't directly interact with the LSP for the most part anyway. A waste of development time for almost no gain at all :)

mrcjkb (dismissed):

PR #1755, opened 2026-07-30
fix: Lua installations on non-nixos distros

vhyrro:
There was a bug with `pkg-probe` trimming `-L /usr/include` causing the lua installation detection to fail. This PR fixes that as well as a secondary bug where custom build backends would not be downloaded properly causing a "package not found" error. This is because we invoke `luarocks make` with `--deps-mode none` when running `lx sync`.

### DIFF
--- a/lux-lib/src/operations/resolve.rs
+++ b/lux-lib/src/operations/resolve.rs
@@ -15,8 +15,9 @@ use crate::{
     lockfile::{
         LocalPackageId, LocalPackageSpec, Lockfile, LockfilePermissions, OptState, PinnedState,
     },
+    lua_rockspec::BuildBackendSpec,
     operations::{FetchVendored, FetchVendoredError},
-    package::{PackageName, PackageReq},
+    package::{PackageName, PackageReq, PackageVersionReqError},
     remote_package_db::RemotePackageDB,
     rockspec::Rockspec,
     tree,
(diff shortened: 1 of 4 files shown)


lux-lib/src/operations/resolve.rs
```diff
@@ -217,6 +218,28 @@ where
                                     })
                                     .collect_vec();
 
+                                if let Some(BuildBackendSpec::LuaRock(backend)) =
+                                    &rockspec.build().current_platform().build_backend
+                                {
+                                    let backend_req = unsafe {
+                                        PackageReq::new_unchecked(
```
mrcjkb: issue: We shouldn't use `new_unchecked` here, because we don't have control over the `backend` string.

mrcjkb (changes requested):

lux-lib/src/operations/resolve.rs
vhyrro: true, crashing here is not something we really want. I'll convert it a bit later

alerque:
As the Arch Linux package maintainer for Lua and Lux, can I ask what you think the problem is on Arch that's different from other distros in detecting the right Lua to use?

vhyrro:
> As the Arch Linux package maintainer for Lua and Lux, can I ask what you think the problem is on Arch that's different from other distros in detecting the right Lua to use?

Hi @alerque , thanks for chiming in. I gave it a deeper look and ran some tests in docker on other systems. It's not just arch but most distros that are non-nix. `pkg-config` natively strips out default library and include paths, which means our internal code never searched for lua headers in the standard locations, that was the sole reason for the errors we were having :)

alerque:
Okay that makes more sense now. At least at first blush this looked like the expected way to do things, not a hack that should be Arch specific. If we're playing nice with the default way `pkg-config` works I'm game with that. If we were doing something wrong/unexpected/sub-par relative to other distros I'd like to know about it. Packaging is never perfect –it's a daily grind of just trying to keep up with releases and gradually make things a little better along the way– but we'd do try hard to keep things we package playing nice together! If there are ever Lua (or Rust) package ecosystem issues with Arch feel free to ping me.

mrcjkb (approved):

PR #1766, opened 2026-08-02
build(nix): don't disable libgit vendoring

mrcjkb:
The git2 crate check for a minimum libgit2 version, which is currently outdated in nixpkgs.
So this PR allows git2 to vendor libgit2.


vhyrro (approved):

mrcjkb:
Will merge the whole stack when it's ready.

PR #1767, opened 2026-08-02
build(nix): debug builds

mrcjkb:


nix/overlay.nix
```diff
@@ -34,42 +34,56 @@
...
+          if release
+          then "--profile release"
+          else "--profile dev"
+        }";
+        cargoBuildCommand = "cargo build ${
+          if release
+          then "--profile release"
+          else "--profile dev"
+        }";
+
```
vhyrro, line 54: question: couldn't this land in `cargoExtraArgs` instead for a major simplification?

mrcjkb: No, we have to override the `cargoBuildCommand` if we want to set the profile.

vhyrro (approved):

PR #1770, opened 2026-08-02
feat!: `lx util` subcommand for manpages & completions

mrcjkb:


vhyrro (commented):
question: wouldn't this be a good moment to create an `lx util` subcommand like `jj` has it where we could place both `completion` and `man` inside?

mrcjkb:
> question: wouldn't this be a good moment to create an `lx util` subcommand like `jj` has it where we could place both `completion` and `man` inside?

Good idea. Let's do it.

vhyrro (approved):

PR #1771, opened 2026-08-02
build(nix): don't rely on xtask for completions & man pages

mrcjkb:


vhyrro:
an excess commit accidentally made it into this PR, is it Github trolling us?

Issue #1218, opened 2025-12-01
[Feature] project-local configs


mrcjkb:
For example in `.lux/config.toml`.

Should be merged with the global user config and take precedence.

mrcjkb:
This feature needs triage, as it can add quite some complexity:

- We construct the config before invoking any commands, passing CLI args to the config builder.
- To override config options with a project-local config, we would have to do so *before* overriding the deserialized `ConfigBuilder` with the CLI args (e.g. with a `ConfigBuilder::override_config(self, other: ConfigBuilder) -> Self` method).
- To do that, we would have to detect the presence of a project and a project-local config *before invoking any commands*.
- **This raises the issue: What if I'm invoking a command that doesn't operate on a project (e.g. `lx install`), but my cwd is a Lux project?**
  In this case, the project-local config would be loaded and would override the user config.
  To prevent this, we would have to implement a `Project::override_config(config: &Config) -> Config` method.
  But that would have to be invoked manually, for each operation that operates on a project.

So far, I haven't come across a use case that requires a project-local config and can't be solved with CLI flags.
I'm inclined to say this is not worth the trade-off of the added complexity at this point (YAGNI).

mrcjkb:
Potential solution: Add a `Commands::config_builder() -> Result<ConfigBuilder>` method, similar to the `Commands::lua_version()` method.
It could return something like `ConfigBuilder::new().merge(workspace.config())` for subcommands that operate on a workspace.

PR #1774, opened 2026-08-02
feat: support detached workspace trees

### LINKED ISSUE #1762: [Feature] ability to detach workspace tree to reside outside of the project root
Currently, the workspace tree `.lux` resides in the workspace root.
We should add a config option and CLI flag that lets the workspace tree reside somewhere else, similar to [Cargo's `build.target-dir`](https://doc.rust-lang.org/cargo/reference/config.html#buildtarget-dir).

mrcjkb:
Closes #1762.

Passing `--tree <tree>` to a workspace subcommand detaches the workspace tree.

### DIFF
--- a/lux-cli/src/bin/lx.rs
+++ b/lux-cli/src/bin/lx.rs
@@ -92,7 +92,8 @@ async fn main() -> Result<()> {
                 .map(|duration| Duration::from_secs(duration as u64)),
         )
         .max_jobs(cli.max_jobs)
-        .user_tree(cli.tree)
+        .user_tree(cli.tree.clone())
+        .workspace_tree(cli.tree)
         .variables(
             cli.variables
                 .map(|variables| variables.into_iter().collect()),
--- a/lux-cli/src/config.rs
+++ b/lux-cli/src/config.rs
@@ -71,6 +71,9 @@ to initialise a config file.
         }
         ConfigCmd::Show => {
             let cfg: ConfigBuilder = config.into();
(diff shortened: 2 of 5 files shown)


vhyrro (changes requested):
issue: the code looks good but in its current form it could cause some problems. If I create a workspace from Lua with a detached tree using my own configuration, a CLI command like `lx sync` will ignore the detached tree entirely because it has no way of knowing that the tree is detached.

My proposed solution: add the ability to specify a workspace-local configuration, either in a dedicated file in the workspace root (`.lux/config` like cargo) or under `[config]` in `lux.toml`. This way you could specify that the tree is detached in the workspace-local configuration and it'd be respected by all `lx` invocations. Wdyt?

mrcjkb:
> issue: the code looks good but in its current form it could cause some problems. If I create a workspace from Lua with a detached tree using my own configuration, a CLI command like `lx sync` will ignore the detached tree entirely because it has no way of knowing that the tree is detached.
> 
> My proposed solution: add the ability to specify a workspace-local configuration, either in a dedicated file in the workspace root (`.lux/config` like cargo) or under `[config]` in `lux.toml`. This way you could specify that the tree is detached in the workspace-local configuration and it'd be respected by all `lx` invocations. Wdyt?

I had considered this, but I wouldn't see it as a blocker.
We have a tracking issue for workspace-local configs:
	
- #1218

vhyrro (approved):

PR #1775, opened 2026-08-02
feat(lsp)!: allow setting port file via `LUX_LSP_PORT_FILE`

mrcjkb:
In Neovim, we can set the `LUX_LSP_PORT_FILE` environment variable to `pathlib.stdpath("run") / lux-lsp`.

### DIFF
--- a/lux-lib/src/progress/mod.rs
+++ b/lux-lib/src/progress/mod.rs
@@ -1,4 +1,5 @@
 use crate::{
+    config::Config,
     progress::client::{LspClient, CLIENT},
     workspace::Workspace,
 };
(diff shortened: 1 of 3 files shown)


lux-lib/src/progress/mod.rs
```diff
@@ -7,8 +8,26 @@ use std::{path::PathBuf, sync::Arc};
...
+
 pub fn lsp_port_path(workspace: &Workspace) -> PathBuf {
-    workspace.root().join(".lux").join("lsp-port")
+    match std::env::var(LUX_LSP_PORT_FILE) {
+        Ok(path) => PathBuf::from(path),
+        Err(_) => {
+            tracing::debug!("{LUX_LSP_PORT_FILE} not set.");
+            match Config::project_dirs()
+                .ok()
+                .and_then(|p| p.runtime_dir().map(|p| p.to_path_buf()))
```
mrcjkb, line 20: `runtime_dir` would be the standard way to do it on Linux & BSD, because it gets cleared after the session (unlike other `XDG_` directories).
The problem is that there's no equivalent for MacOS & Windows. Neovim's `vim.fs.stdpath("run")` falls back to a temp directory on those platforms.

I think the best approach for lux.nvim would be to set the `LUX_LSP_PORT_FILE` environment variable to `pathlib.stdpath("run") / lux-lsp-port`.

vhyrro (approved):
honestly, this works well enough and will give us all the granularity we need to prevent a `.lux/` from being created in user's config directories. I approve of this approach

Issue #1469, opened 2026-04-25
[feat] Allow setting `lx fmt` backend in `lux.toml`


benniekiss:
Unless I'm mistaken, the formatting backend can only be specified on the command-line via the `--backend` flag. Being able to set this in the `lux.toml` would be very helpful, possibly under a `[format.backend]` key. 

This table could also include an `args` key to pass additional args to the backend, and `include` and `exclude` keys with patterns of files to include or exclude.

mrcjkb:
Related: #1218.

This is an interesting case, as formatting is implemented in `lux-cli`, outside of the `lux-lib` crate.
A dedicated key for default `lux-cli` flags could potentially be useful.
On the other hand, I don't know how well clap (the CLI arg parsing library) supports this. It could add a lot of complexity for too little value (see https://github.com/lumen-oss/lux/issues/1218#issuecomment-3708451818). For formatting with pre-configured flags, there exist tools like [treefmt](https://github.com/numtide/treefmt).

@vhyrro do you have any thoughts on this?

benniekiss:
So I was looking over the formatter cli code, and both formatters are actually dispatched via rust -- I was under the impression it was shelling out. With this the case, there aren't really any "args" that could be passed.

I still think having the format table in the project toml would be helpful, rather than an additional config for the cli specifically. Especially since formatting can be so project specific. With the two separate backends, project level config makes it easier for collaboration if a project is using emmylua-codestyle.

I'm only just familiarizing myself with the code, but I've gotten something together minimally adding the `format` table to the project toml. If you'd like to take a look, its a rather small changeset -> https://github.com/lumen-oss/lux/compare/main...benniekiss:lux:fmt-backend. It does require moving some of the functionality into `lux-lib`, and it is rather bespoke for this specific use case, so I understand if this is not the direction you want the project to go.

mrcjkb:
> I still think having the format table in the project toml would be helpful, rather than an additional config for the cli specifically.

We probably don't need a config for every CLI flag, but there will likely be other flags that can be configured per project (for example `lint` or `check` args).
And as you can see in your change, the `[format]` rockspec key has no meaning for `lux-lib`, so we'll probably want a TOML key for things that `lux-lib` can ignore. I'm not too keen on CLI specific types leaking into the lib.

PR #1777, opened 2026-08-03
feat: workspace-local configs

mrcjkb:
Initial implementation for #1218.
This does not add support for subcommands, (e.g. #1469).

### DIFF
--- a/lux-cli/src/lib.rs
+++ b/lux-cli/src/lib.rs
@@ -5,6 +5,7 @@ use crate::{
     project::NewProject,
     util::Util,
 };
+use miette::Result;
 use std::error::Error;
 use std::path::PathBuf;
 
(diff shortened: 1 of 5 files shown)


lux-cli/src/lib.rs
```diff
@@ -441,6 +449,84 @@ impl Commands {
...
+            })
+            | Self::Pack(Pack {
+                package_or_rockspec: None,
+            })
+            | Self::Debug(Debug::Project(_)) => Workspace::current().into_diagnostic(),
+            // non-project commands
+            Self::Debug(Debug::Unpack(_))
+            | Self::Debug(Debug::FetchRemote(_))
+            | Self::Debug(Debug::UnpackRemote(_))
+            | Self::Debug(Debug::Toolchains(_))
```
mrcjkb, line 516: note: We split on each one so that the compiler tells us to add cases if we add new subcommands in the future.

vhyrro (approved):

PR #1803, opened 2026-08-08
feat(build): `release` and `dev` profiles

### LINKED ISSUE #1712: [Feature] dev build profile
`lx` could have a `--profile release|dev` option (defaulting to `dev` for workspace commands except `lx dist`) to speed up build backends for local project builds.

Specifically, the `rust-mlua` build backend could unset the `--release` flag if that environment variable is set.

mrcjkb:
Closes #1712.

- `dev` is the default for local workspace builds.
  It compiles with `-O0` and Cargo's `dev` profile.
- `release` is the default for installs and the `dist`
  commands.
  It compiles with `-O3` and Cargo's `release` profile.

### DIFF
--- a/lux-cli/src/lib.rs
+++ b/lux-cli/src/lib.rs
@@ -149,6 +149,11 @@ pub struct Cli {
     #[arg(long, value_name = "variable", visible_short_alias = 'v', value_parser = parse_key_val::<String, String>)]
     pub variables: Option<Vec<(String, String)>>,
 
+    /// The build profile to use when compiling packages.{n}
+    /// Default: `release`.
+    #[arg(long, value_enum, value_name = "profile")]
+    pub profile: Option<lux_lib::config::build::Profile>,
+
     /// Display verbose output of commands executed, enabling DEBUG logs.{n}
     /// To enable TRACE logs, set RUST_LOG=trace.
     #[arg(long)]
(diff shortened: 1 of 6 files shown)


lux-cli/src/lib.rs
```diff
@@ -454,14 +459,13 @@ impl Commands {
...
-            .flatten()
-        {
-            Ok(config.merge(workspace_config))
+        if let Some(workspace) = self.workspace()? {
+            let config = workspace.config()?.unwrap_or(config);
+            if let Self::Dist(_) = self {
+                Ok(config)
+            } else {
+                Ok(config.default_build_profile(lux_lib::config::build::Profile::Dev))
+            }
```
vhyrro, line 472: question: this deleted the workspace_config merging. Is there a rationale for this that I'm not seeing?

mrcjkb: Good catch. That was an oversight on my end

lux-lib/src/build/utils.rs
```diff
@@ -158,6 +157,11 @@ pub(crate) async fn compile_c_files(
     external_dependencies: &HashMap<String, ExternalDependencyInfo>,
     config: &Config,
 ) -> Result<(), CompileCFilesError> {
+    let span = info_span!(
+        "Compiling C files",
+        profile = config.build_profile().to_string()
+    );
+    let _enter = span.enter();
```
vhyrro, line 164: suggestion: we should not create standalone spans in async contexts. Would you like to fix this now, or will we do a big "fix commit" for each instance this appears in code later?

vhyrro (approved):

lux-lib/src/build/utils.rs
mrcjkb: Let's merge this and I'll rebase #1819 on this.

PR #1811, opened 2026-08-09
fix(dist/bin): resolve native modules via rpath

mrcjkb:


mrcjkb:
Merging, as this is trivial.

PR #1814, opened 2026-08-10
fix: gag stdout during sync to prevent `cargo:` rerun output

vhyrro:


lux-lib/src/operations/sync.rs
```diff
@@ -165,11 +165,14 @@ pub enum SyncError {
...
 #[tracing::instrument(name = "Syncing dependencies", skip_all)]
-
 async fn do_sync(
     args: Sync<'_>,
     lock_type: &LocalPackageLockType,
 ) -> Result<SyncReport, SyncError> {
+    // NOTE(vhyrro): tools like cc and pkg-config leak cargo:rerun-if-env-changed
+    // stdout calls, therefore gag all standard output during sync.
+    let _stdout_gag = gag::Gag::stdout();
+
```
mrcjkb, line 175: question: Could this also suppress trace output? Or are they still emitted because they're declared in a different scope?

suggestion(non-blocking): Add this to the pkg-config probe calls, so that they get gagged if used elsewhere. We never want to see pkgs-config probing output.

mrcjkb (dismissed):

lux-lib/src/operations/sync.rs
vhyrro: tracing output goes to stderr, so it's untouched. Let me add gag to pkg-config :)

PR #1817, opened 2026-08-10
feat!: add install/uninstall reports to lux-lua

vhyrro:
this will allow us to manage the `package.path` in an easier fashion in lux.nvim

### DIFF
--- a/lux-lib/src/operations/uninstall.rs
+++ b/lux-lib/src/operations/uninstall.rs
@@ -62,7 +62,7 @@ where
     State: uninstall_builder::State + uninstall_builder::IsComplete,
 {
     /// Remove the packages.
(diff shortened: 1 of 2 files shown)


lux-lib/src/operations/uninstall.rs
```diff
@@ -62,7 +62,7 @@ where
     State: uninstall_builder::State + uninstall_builder::IsComplete,
 {
     /// Remove the packages.
-    pub async fn remove(self) -> Result<(), RemoveError> {
+    pub async fn remove(self) -> Result<Vec<LocalPackageId>, RemoveError> {
```
mrcjkb, line 65: quibble: pub signature change is a breaking change in lux-lib (commit prefix)

mrcjkb (dismissed):

PR #1819, opened 2026-08-10
fix(tracing): properly instrument async code

mrcjkb:


mrcjkb:
what the hell?

```
error: queries overflow the depth limit!
  |
  = help: consider increasing the recursion limit by adding a `#![recursion_limit = "256"]` attribute to your crate (`exec`)
  = note: query depth increased by 130 when computing layout of `{async block@lux-lib/tests/exec.rs:7:1: 7:32}`
```

PR #1826, opened 2026-08-11
refactor: replace `#[tokio::main]` with `runtime::block_on` per command

mrcjkb:
Reduces stack usage in debug builds by ~1 MB.

### DIFF
--- a/lux-cli/src/bin/lx.rs
+++ b/lux-cli/src/bin/lx.rs
@@ -32,8 +32,7 @@ use tracing_subscriber::{
 use lux_cli::utils::error::clap_to_miette;
 
 const DEFAULT_USER_AGENT: &str = concat!("lux/", env!("CARGO_PKG_VERSION"));
-#[tokio::main(flavor = "multi_thread")]
-async fn main() -> Result<()> {
+fn main() -> Result<()> {
     miette::set_hook(Box::new(|_| {
         Box::new(
             MietteHandlerOpts::new()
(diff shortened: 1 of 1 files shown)


vhyrro (changes requested):
issue: not really a fan of manually having block_on everywhere.

It might be worth looking at https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.thread_stack_size

Instead of doing `block_on` everywhere, why not use this popular pattern instead:
```rs
fn main() {
    tokio::runtime::Builder::new_multi_thread()
        // ...
        .build()
        .unwrap()
        .block_on(async_main())
}

async fn async_main() {}
```

mrcjkb:
> issue: not really a fan of manually having block_on everywhere.
> 
> It might be worth looking at https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.thread_stack_size
> 
> Instead of doing `block_on` everywhere, why not use this popular pattern instead:
> 
> ```rust
> fn main() {
>     tokio::runtime::Builder::new_multi_thread()
>         // ...
>         .build()
>         .unwrap()
>         .block_on(async_main())
> }
> 
> async fn async_main() {}
> ```

The stack size is only a problem for us in unoptimised dev builds on Windows. So increasing it in CI is enough to solve that issue.
I opened this because I think there's no reason to make the whole `main` function async. The setup is sync and `rt.block_on(async_fn())` is not really any more boilerplate than `async_fn().await`.
But if you feel strongly about this, we can just close this PR.

PR #1827, opened 2026-08-12
feat(tracing): include build backend name in luarocks span

mrcjkb:


mrcjkb:
Merging, as this one is trivial.

PR #1836, opened 2026-08-12
fix(install): better error diagnostics

mrcjkb:
- propagate build error diagnostics
- report original error when falling back to `.src.rock` download fails and log the `.src.rock` download error.


mrcjkb:
No idea why GH is showing both commits in this PR.

vhyrro (dismissed):

PR #1839, opened 2026-08-12
feat(fetch): retry downloading rock sources on failure

mrcjkb:
Uses `reqwest-retry` to retry downloading rockspecs, manifests & sources with exponential backoff & jitter.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3586,6 +3586,8 @@ dependencies = [
  "proptest",
  "remove_dir_all",
  "reqwest",
+ "reqwest-middleware",
+ "reqwest-retry",
  "semver",
  "serde",
  "serde-enum-str",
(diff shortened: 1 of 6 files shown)


vhyrro (approved):

mrcjkb:
```
InstallTestDependencies(Install(Build(PackageName("mediator_lua"), FetchSrcError(RequestMiddleware(Middleware(error sending request for url (https://github.com/Olivine-Labs/mediator_lua/archive/v1.1.2-0.tar.gz)

    Caused by:
        0: client error (SendRequest)
        1: http2 error
        2: stream error received: refused stream before processing any application logic))))))
```

We're hitting this quite often. 
mediator_lua doesn't have a .src.rock so the fallback download fails when we hit GitHub rate limits.
Retrying doesn't help here and potentially even makes it worse.

Leaving this open for now, I need to think about this some more.

mrcjkb:
This adds quite a bit of complexity that may be unnecessary since #1843.
Let's close this as YAGNI and revisit another time.

PR #1852, opened 2026-08-16
fix(fetch)!: include ignore files when copying file sources

mrcjkb:
This ensures that we end up with the same hash as a vendored nix source.

### DIFF
--- a/lux-lib/src/fs/tokio.rs
+++ b/lux-lib/src/fs/tokio.rs
@@ -1,8 +1,10 @@
 use super::FsError;
 #[cfg(unix)]
 use std::fs::Permissions;
+use std::io;
 use std::path::Path;
 use tokio::fs;
+use walkdir::WalkDir;
 
 /// Wrapped [`fs::read`].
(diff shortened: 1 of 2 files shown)


lux-lib/src/fs/tokio.rs
```diff
@@ -49,6 +51,34 @@ pub(crate) async fn copy(from: impl AsRef<Path>, to: impl AsRef<Path>) -> Result
...
+/// but skipping `.git` directories.
+pub(crate) async fn copy_dir_contents(
+    src: impl AsRef<Path>,
+    dest: impl AsRef<Path>,
+) -> Result<(), FsError> {
+    let src = src.as_ref();
+    let dest = dest.as_ref();
+    for path in WalkDir::new(src)
+        .into_iter()
+        .filter_entry(|entry| entry.file_name() != ".git")
```
vhyrro: suggestion: maybe we should also filter out other common VCS folders like `.jj/`, `.hg`, `_darcs/`, `.svn/`, `.bzr`?

PR #1857, opened 2026-08-16
feat(config): fall back to temp directory if $HOME not set

mrcjkb:
This allows us to build within a nix sandbox without having to spoof a home directory.

### DIFF
--- a/lux-lib/src/config/mod.rs
+++ b/lux-lib/src/config/mod.rs
@@ -81,19 +81,39 @@ pub struct Config {
 impl Config {
     /// Lux application directories
     pub(crate) fn project_dirs() -> Result<ProjectDirs, NoValidHomeDirectory> {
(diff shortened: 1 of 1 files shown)


vhyrro:
question: is there ever a case where a user might not have their `$HOME` set and be very confused that Lux is using their /tmp instead? `/tmp` isn't particularly large for most regular systems, so I'm wondering if we ever run the risk of an overflow or other error there.

mrcjkb:
> question: is there ever a case where a user might not have their `$HOME` set and be very confused that Lux is using their /tmp instead? `/tmp` isn't particularly large for most regular systems, so I'm wondering if we ever run the risk of an overflow or other error there.

That's why I added a warning log message. tracing-indicatif will display log messages with level >= info.

We build in /tmp, so the overflow risk is present on small systems anyway.

vhyrro (approved):

PR #1862, opened 2026-08-16
feat(nix): lux packaging helpers

### LINKED ISSUE #1265: how to go about a buildLuxPackage ?
hi,

I would like to add a `buildLuxPackage` to nixpkgs. I've vibe-coded a few bits replicating the cargo infrastructure https://github.com/teto/nixpkgs/blob/scratch/pkgs/build-support/fetch-lux-deps.nix , for instance I can run  `nix-build -A rikai-nvim-vendor` to generate a folder with the source of deps.

```
$ tree -L3 $(nix-build -A rikai-nvim-vendor)
./result
├── alogger
│   └── 0.6.0-1
│       ├── alogger-scm-1.rockspec
│       ├── doc
│       ├── examples
│       ├── LICENSE
│       ├── Makefile
│       ├── README.md
│       ├── rockspecs
│       ├── spec
│       └── src
├── busted
│   └── 2.2.0-1
│       ├── action.yml
│       ├── bin
│       ├── busted
│       ├── busted.lua
│       ├── busted-scm-1.rockspec
│       ├── completions
│       ├── CONTRIBUTING.md
│       ├── Dockerfile
│       ├── LICENSE
│       ├── README.md
│       ├── rockspecs
│       ├── spec
│       ├── TODO.md
│       ├── try
│       └── try.bat
├── dkjson
│   └── 2.8-1
│       ├── dkjson-2.8-1.rockspec
│       ├── dkjson.lua
│       ├── jsontest.lua
│       ├── readme.txt
│       ├── speedtest.lua
│       └── versions.txt
├── lsqlite3
│   └── 0.9.6-1
│       ├── doc
│       ├── examples
│       ├── extras
│       ├── HISTORY
│       ├── lsqlite3-0.9.6-1.rockspec
│       ├── lsqlite3.c
│       ├── lsqlite3complete-0.9.6-1.rockspec
│       ├── Makefile
│       ├── manifest
│       ├── manifest.tags
│       ├── manifest.uuid
│       ├── README
│       ├── sqlite3.c
│       ├── sqlite3.h
│       └── test
├── lua_cliargs
│   └── 3.0.2-1
│       ├── bin
│       ├── doc
│       ├── examples
│       ├── LICENSE
│       ├── lua_cliargs-dev-1.rockspec
│       ├── README.md
│       ├── rockspecs
│       ├── spec
│       ├── src
│       └── UPGRADE.md
├── luafilesystem
│   └── 1.8.0-1
│       ├── appveyor.yml
│       ├── config
│       ├── config.win
│       ├── doc
│       ├── LICENSE
│       ├── luafilesystem-scm-1.rockspec
│       ├── Makefile
│       ├── Makefile.win
│       ├── README.md
│       ├── src
│       ├── tests
│       └── vc6
├── luassert
│   └── 1.9.0-1
│       ├── CONTRIBUTING.md
│       ├── LICENSE
│       ├── luassert-scm-1.rockspec
│       ├── README.md
│       ├── rockspecs
│       ├── spec
│       └── src
├── luasystem
│   └── 0.6.3-1
│       ├── CHANGELOG.md
│       ├── config.ld
│       ├── docs
│       ├── doc_topics
│       ├── examples
│       ├── LICENSE.md
│       ├── luasystem-scm-0.rockspec
│       ├── Makefile
│       ├── README.md
│       ├── rockspecs
│       ├── spec
│       ├── src
│       └── system
├── lua-term
│   └── 0.8-1
│       ├── CHANGES
│       ├── COPYING
│       ├── core.c
│       ├── lua-term-0.8-1.rockspec
│       ├── Makefile
│       ├── README.md
│       └── term
├── luv
│   └── 1.51.0-1
│       ├── cmake
│       ├── CMakeLists.txt
│       ├── deps
│       ├── docs.md
│       ├── libluv.pc.in
│       ├── LICENSE.txt
│       ├── README.md
│       └── src
├── mediator_lua
│   └── 1.1.2-0
│       ├── mediator_lua-1.1.2-0.rockspec
│       ├── README.md
│       ├── spec
│       └── src
├── mega.cmdparse
│   └── 1.2.0-1
│       ├── CHANGELOG.md
│       ├── doc
│       ├── LICENSE
│       ├── lua
│       ├── Makefile
│       ├── README.md
│       ├── renovate.json
│       ├── scripts
│       ├── spec
│       └── template.rockspec
├── mega.logging
│   └── 1.1.6-1
│       ├── CHANGELOG.md
│       ├── doc
│       ├── LICENSE
│       ├── lua
│       ├── Makefile
│       ├── README.md
│       ├── renovate.json
│       ├── scripts
│       ├── spec
│       └── template.rockspec
├── nlua
│   └── 0.3.2-1
│       ├── LICENSE
│       ├── nlua
│       ├── nlua.bat
│       ├── nlua-scm-1.rockspec
│       ├── README.md
│       ├── rockspec.template
│       └── spec
├── penlight
│   └── 1.14.0-3
│       ├── appveyor.yml
│       ├── CHANGELOG.md
│       ├── config.ld
│       ├── CONTRIBUTING.md
│       ├── docs
│       ├── docs_topics
│       ├── examples
│       ├── ldoc.ltp
│       ├── LICENSE.md
│       ├── lua
│       ├── penlight-dev-1.rockspec
│       ├── README.md
│       ├── rockspecs
│       ├── run.lua
│       ├── spec
│       └── tests
├── say
│   └── 1.4.1-3
│       ├── CONTRIBUTING.md
│       ├── LICENSE
│       ├── README.md
│       ├── rockspecs
│       ├── say-scm-1.rockspec
│       ├── spec
│       └── src
├── sqlite
│   └── v1.0.0-1
│       ├── CHANGELOG.md
│       ├── doc
│       ├── LICENSE
│       ├── lua
│       ├── Makefile
│       ├── README.md
│       ├── scripts
│       └── test
└── utf8
    └── 1.3-0
        ├── fuzzer
        ├── GraphemeBreakTest.txt
        ├── LICENSE
        ├── lutf8lib.c
        ├── NormalizationTest.txt
        ├── parseucd.lua
        ├── README.md
        ├── rockspecs
        ├── test_compat.lua
        ├── test.lua
        ├── test_pm.lua
        └── unidata.h

102 directories, 111 files
```

I've been playing with the "tracing" crate to try understand how things work in lux at https://github.com/teto/lux/tree/matt/tracing but I thought I would ask for guidelines here rather than working out the good workflow myself as I am not that familiar with lux itself.

The question is how to provide lux with prebuilt dependencies, such that it doesn't access the network while building ? Nix itself is capable of parsing the lux lockfile to retreive the source code as fixed outputs.

mrcjkb:
Closes #1265.

cc @teto @khaneliman this is a staging ground for potential upstreaming to nixpkgs.
Would love to get your feedback...

### DIFF
--- a/.github/workflows/nix-build.yml
+++ b/.github/workflows/nix-build.yml
@@ -76,3 +76,16 @@ jobs:
       shell: bash
       # FIXME: This fails with permission denied on darwin
       if: ${{ matrix.job.target == 'x86_64-linux' }}
+    - name: Lux nix helper tests
+      run: |
+        nix build ".#checks.${{matrix.job.target}}.build-lux-helpers" -Lv --accept-flake-config --log-lines 500 --fallback
+        nix build ".#checks.${{matrix.job.target}}.build-lux-helpers-import-lock" -Lv --accept-flake-config --log-lines 500 --fallback
+        nix build ".#checks.${{matrix.job.target}}.build-lux-helpers-lux-vendor-dir" -Lv --accept-flake-config --log-lines 500 --fallback
+        nix build ".#checks.${{matrix.job.target}}.build-lux-helpers-with-packages" -Lv --accept-flake-config --log-lines 500 --fallback
+        nix build ".#checks.${{matrix.job.target}}.build-lux-helpers-lux-lock" -Lv --accept-flake-config --log-lines 500 --fallback
+        nix build ".#checks.${{matrix.job.target}}.rockspec-cjson" -Lv --accept-flake-config --log-lines 500 --fallback
+        nix build ".#checks.${{matrix.job.target}}.rockspec-luassert" -Lv --accept-flake-config --log-lines 500 --fallback
+        nix build ".#checks.${{matrix.job.target}}.nvim-plugin" -Lv --accept-flake-config --log-lines 500 --fallback
+      shell: bash
+      # FIXME: lux-lua build fails with permission denied on darwin
+      if: ${{ matrix.job.target == 'x86_64-linux' }}
--- a/nix/build-lux-application.nix
+++ b/nix/build-lux-application.nix
@@ -1,0 +1,42 @@
+{
+  lib,
+  buildLuxPackage,
+  makeWrapper,
+  lua,
+}: {
+  runCommand ? null,
+  runArgs ? [],
+  ...
+} @ args: let
+  interpreter =
+    if runCommand == null
+    then "${lua}/bin/lua"
+    else runCommand;
+in
+  buildLuxPackage (
+    removeAttrs args [
+      "runArgs"
+      "runCommand"
+    ]
+    // {
+      nativeBuildInputs = args.nativeBuildInputs or [] ++ [makeWrapper];
+
+      postInstall =
+        (args.postInstall or "")
+        + ''
+          cp -r "$src/." "$out/"
+          mkdir -p "$out/bin"
+
+          LUA_PATH="$(lx --lua-version "${lua.luaversion}" --tree "$out/share/lux" path lua)"
+          LUA_CPATH="$(lx --lua-version "${lua.luaversion}" --tree "$out/share/lux" path c)"
+          LUA_INIT="$(lx --lua-version "${lua.luaversion}" --tree "$out/share/lux" path init)"
+
+          makeWrapper "${interpreter}" "$out/bin/${args.pname}" \
+            --chdir "$out" \
+            --set LUA_INIT "$LUA_INIT" \
+            --set LUA_PATH "$LUA_PATH" \
+            --set LUA_CPATH "$LUA_CPATH" \
+            --add-flags ${lib.escapeShellArg (lib.escapeShellArgs runArgs)}
+        '';
+    }
+  )
--- a/nix/build-lux-package.nix
+++ b/nix/build-lux-package.nix
@@ -1,0 +1,178 @@
+{
+  lib,
+  stdenv,
+  fetchLuxDeps,
+  importLuxLock,
+  lux-cli,
+  luxLoaderSetupHook,
+  pkg-config,
+  lua,
+}:
+lib.extendMkDerivation {
+  constructDrv = stdenv.mkDerivation;
+
+  excludeDrvArgNames = [
+    "luxDeps"
+    "luxHash"
+    "luxLock"
+    "luxRoot"
+    "luxVendorDir"
+  ];
+
+  transformDrv = drv:
+    (lua.pkgs.toLuaModule drv).overrideAttrs (old: {
+      passthru = (old.passthru or {}) // {luxPackage = true;};
+    });
+
+  extendDrvArgs = finalAttrs: {
+    src,
+    luxLock ? null,
+    luxHash ? null,
+    luxDeps ? null,
+    luxVendorDir ? null,
+    luxRoot ? null,
+    buildAndTestSubdir ? null,
+    nvim ? false,
+    nativeBuildInputs ? [],
+    propagatedBuildInputs ? [],
+    postUnpack ? "",
+    ...
+  }: let
+    deps =
+      if luxVendorDir != null
+      then luxVendorDir
+      else if luxDeps != null
+      then luxDeps
+      else if luxLock != null
+      then importLuxLock {lockFile = luxLock;}
+      else if luxHash != null
+      then
+        fetchLuxDeps {
+          inherit src luxRoot;
+          pname = finalAttrs.pname;
+          version = finalAttrs.version;
+          hash = luxHash;
+        }
+      else throw "buildLuxPackage requires either (sorted by precedence) `luxVendorDir`, `luxDeps`, `luxLock` or `luxHash`";
+
+    rootSubdir = lib.optionalString (luxRoot != null) "${luxRoot}/";
+
+    buildSubdir = lib.optionalString (buildAndTestSubdir != null) "${buildAndTestSubdir}/";
+
+    luaVersionFlag =
+      if nvim
+      then "--nvim"
+      else "--lua-version \"${lua.luaversion}\"";
+
+    lockFile =
+      if luxLock != null
+      then luxLock
+      else if !lib.isDerivation src && lib.pathExists "${src}/${rootSubdir}lux.lock"
+      then "${src}/${rootSubdir}lux.lock"
+      else null;
+
+    expectedRocks =
+      if lockFile == null
+      then []
+      else let
+        lock = builtins.fromJSON (lib.readFile lockFile);
+      in
+        lib.concatMap (
+          section:
+            lib.mapAttrsToList (_id: rock: "${rock.name}@${rock.version}") (
+              lock.${section}.rocks or {}
+            )
+        ) [
+          "dependencies"
+          "test_dependencies"
+          "build_dependencies"
+        ];
+  in {
+    name = "lua${lua.luaversion}-${finalAttrs.pname}-${finalAttrs.version}";
+
+    __structuredAttrs = true;
+    strictDeps = true;
+
+    nativeBuildInputs =
+      [
+        lux-cli
+        pkg-config
+        lua
+      ]
+      ++ nativeBuildInputs;
+
+    propagatedBuildInputs =
+      [
+        lua
+        lua.pkgs.lux-lua
+        luxLoaderSetupHook
+      ]
+      ++ propagatedBuildInputs;
+
+    postUnpack =
+      postUnpack
+      + lib.optionalString (lockFile != null) ''
+        for rock in ${lib.escapeShellArgs expectedRocks}; do
+          if [ ! -e "${deps}/$rock" ] && [ ! -e "${deps}/$rock.rock" ]; then
+            echo "ERROR: missing vendored rock: $rock" >&2
+            echo "The vendored dependencies are out of date (stale luxHash, luxDeps, luxLock or luxVendorDir)." >&2
+            echo >&2
+            echo "To fix the issue:" >&2
+            echo '1. Set luxHash to an empty string: `luxHash = "";` (or re-vendor luxDeps/luxLock/luxVendorDir)' >&2
+            echo '2. Build the derivation and wait for it to fail with a hash mismatch' >&2
+            echo '3. Copy the "got: sha256-..." value back into the luxHash field' >&2
+            echo >&2
+            echo 'Note: If you are trying to override luxHash, set `luxDeps = fetchLuxDeps ...` instead.' >&2
+            echo >&2
+            exit 1
+          fi
+        done
+      '';
+
+    buildPhase = ''
+      runHook preBuild
+
+      if (( ''${NIX_DEBUG:-0} >= 1 )); then
+        export RUST_LOG="trace"
+      fi
+
+      ${lib.optionalString (buildAndTestSubdir != null) "pushd ${buildAndTestSubdir}"}
+
+      ${lib.optionalString (luxLock != null) "install -m 644 ${luxLock} lux.lock"}
+
+      lx --profile release \
+         ${luaVersionFlag} \
+         --vendor-dir "${deps}" \
+         build
+
+      ${lib.optionalString (buildAndTestSubdir != null) "popd"}
+
+      runHook postBuild
+    '';
+
+    installPhase = ''
+      runHook preInstall
+
+      mkdir -p "$out/share/lux"
+      cp -rT "${buildSubdir}.lux" "$out/share/lux"
+
+      runHook postInstall
+    '';
+
+    checkPhase = ''
+      runHook preCheck
+
+      ${lib.optionalString (buildAndTestSubdir != null) "pushd ${buildAndTestSubdir}"}
+
+      lx --profile release \
+         ${luaVersionFlag} \
+         --vendor-dir "${deps}" \
+         --tree "$TMPDIR/lux-test" \
+         test
+
+      ${lib.optionalString (buildAndTestSubdir != null) "popd"}
+
+      runHook postCheck
+    '';
+  };
+}
--- a/nix/build-lux-rockspec.nix
+++ b/nix/build-lux-rockspec.nix
@@ -1,0 +1,127 @@
+{
+  lib,
+  stdenv,
+  lua,
+  luaLib,
+  lux-cli,
+  luxLoaderSetupHook,
+  pkg-config,
+  writeTextFile,
+}:
+lib.extendMkDerivation {
+  constructDrv = stdenv.mkDerivation;
+
+  excludeDrvArgNames = [
+    "knownRockspec"
+    "rockspecVersion"
+    "nvim"
+  ];
+
+  transformDrv = drv:
+    (lua.pkgs.toLuaModule drv).overrideAttrs (old: {
+      passthru = (old.passthru or {}) // {luxPackage = true;};
+    });
+
+  extendDrvArgs = finalAttrs: {
+    src,
+    knownRockspec ? null,
+    rockspecVersion ? finalAttrs.version,
+    nvim ? false,
+    nativeBuildInputs ? [],
+    propagatedBuildInputs ? [],
+    ...
+  }: let
+    inherit (finalAttrs) pname version;
+
+    luaVersionDir = lua.luaversion;
+    luaVersionFlag =
+      if nvim
+      then "--nvim"
+      else "--lua-version '${luaVersionDir}'";
+
+    luaDeps = lib.filter (drv: drv ? luaModule) propagatedBuildInputs;
+    luarocksDeps =
+      lib.filter (
+        drv:
+          if drv ? luxPackage
+          then lib.warn "buildLuxRockspec: lux-built dependency ${drv.pname} is not vendored; the build will fail if the rockspec requires it" false
+          else true
+      )
+      luaDeps;
+
+    luarocksConfig = luaLib.generateLuarocksConfig {
+      local_cache = "";
+      requiredLuaRocks = luarocksDeps;
+    };
+
+    luarocksConfigFile = writeTextFile {
+      name = finalAttrs.pname + "-luarocks-config.lua";
+      text = lib.generators.toLua {asBindings = true;} luarocksConfig;
+    };
+
+    rockspecArg =
+      if knownRockspec != null
+      then "cp ${knownRockspec} ./pkg.rockspec"
+      else "cp \"$src/${pname}-${rockspecVersion}.rockspec\" ./pkg.rockspec";
+  in {
+    name = "lua${luaVersionDir}-${pname}-${version}";
+    inherit src;
+
+    __structuredAttrs = true;
+    strictDeps = true;
+    env = {
+      LUAROCKS_CONFIG = luarocksConfigFile;
+    };
+
+    nativeBuildInputs =
+      [
+        lux-cli
+        pkg-config
+        (lua.withPackages (ps: lib.optional (luarocksDeps != []) ps.luarocks))
+      ]
+      ++ nativeBuildInputs;
+
+    propagatedBuildInputs =
+      [lua lua.pkgs.lux-lua luxLoaderSetupHook] ++ propagatedBuildInputs;
+
+    buildPhase = ''
+      runHook preBuild
+
+      ${rockspecArg}
+
+      # Provide the package's own source through a vendor dir, so lux fetches
+      # it locally instead of from the rockspec's remote URL.
+      mkdir -p ./vendor
+
+      ${lib.concatStringsSep "\n" (map (dep: ''
+          luarocks pack ${dep.pname} ${dep.version}
+          ROCKSPEC_FILENAME=${dep.pname}-${dep.version}.rockspec
+          ROCKSPEC=${dep}/${dep.pname}-${dep.version}-rocks/${dep.pname}/${dep.version}/$ROCKSPEC_FILENAME
+          cp $ROCKSPEC ./vendor/$ROCKSPEC_FILENAME
+          cp ${dep.pname}-${dep.version}.*.rock ./vendor/${dep.pname}@${dep.version}.rock
+        '')
+        luarocksDeps)}
+
+      cp -r "$src" "./vendor/${pname}@${rockspecVersion}"
+      chmod -R u+w "./vendor/${pname}@${rockspecVersion}"
+
+      lx --profile release \
+         ${luaVersionFlag} \
+         --vendor-dir "$PWD/vendor" \
+         --tree ./.lux \
+         --no-wrap-bin \
+         install-rockspec ./pkg.rockspec
+
+      runHook postBuild
+    '';
+
+    installPhase = ''
+      runHook preInstall
+
+      mkdir -p "$out/share/lux"
+      cp -rT .lux "$out/share/lux"
+
+      runHook postInstall
+    '';
+  };
+}
--- a/nix/fetch-lux-deps.nix
+++ b/nix/fetch-lux-deps.nix
@@ -1,0 +1,71 @@
+{
+  lib,
+  stdenvNoCC,
+  lux-cli,
+  git,
+  cacert,
+}:
+lib.extendMkDerivation {
+  constructDrv = stdenvNoCC.mkDerivation;
+
+  excludeDrvArgNames = [
+    "hash"
+    "luxLock"
+    "luxRoot"
+  ];
+
+  extendDrvArgs = finalAttrs: {
+    pname ? "lux",
+    version ? "0",
+    hash,
+    src,
+    luxLock ? null,
+    luxRoot ? null,
+    nativeBuildInputs ? [],
+    ...
+  }: {
+    name = "${pname}-${version}-vendor-deps";
+
+    inherit src;
+
+    nativeBuildInputs =
+      [
+        lux-cli
+        git
+        cacert
+      ]
+      ++ nativeBuildInputs;
+
+    impureEnvVars = lib.fetchers.proxyImpureEnvVars;
+
+    env = {
+      SSL_CERT_FILE = "${cacert}/etc/ssl/certs/ca-bundle.crt";
+      GIT_SSL_CAINFO = "${cacert}/etc/ssl/certs/ca-bundle.crt";
+    };
+
+    buildPhase = ''
+      runHook preBuild
+
+      ${lib.optionalString (luxRoot != null) "cd ${luxRoot}"}
+
+      ${lib.optionalString (luxLock != null) "cp ${luxLock} lux.lock"}
+
+      lx vendor "$out" --no-delete
+
+      runHook postBuild
+    '';
+
+    strictDeps = true;
+
+    dontConfigure = true;
+    dontInstall = true;
+    dontFixup = true;
+
+    outputHash = hash;
+    outputHashAlgo =
+      if hash == ""
+      then "sha256"
+      else null;
+    outputHashMode = "recursive";
+  };
+}
--- a/nix/import-lux-lock.nix
+++ b/nix/import-lux-lock.nix
@@ -1,0 +1,75 @@
+{
+  lib,
+  runCommand,
+  fetchurl,
+  fetchgit,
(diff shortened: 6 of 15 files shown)


nix/build-lux-package.nix
```diff
@@ -0,0 +1,182 @@
...
+      + lib.optionalString (lockFile != null) ''
+        for rock in ${lib.escapeShellArgs expectedRocks}; do
+          if [ ! -e "${deps}/$rock" ] && [ ! -e "${deps}/$rock.rock" ]; then
+            echo "ERROR: missing vendored rock: $rock" >&2
+            echo "The vendored dependencies are out of date (stale luxHash, luxDeps, luxLock or luxVendorDir)." >&2
+            echo >&2
+            echo "To fix the issue:" >&2
+            echo '1. Set luxHash to an empty string: `luxHash = "";` (or re-vendor luxDeps/luxLock/luxVendorDir)' >&2
+            echo '2. Build the derivation and wait for it to fail with a hash mismatch' >&2
+            echo '3. Copy the "got: sha256-..." value back into the luxHash field' >&2
```
PerchunPak, line 123: I see this is copied from `buildRustPackage`, and it has the same confusing message if a user tries to override a `luxHash`. Can you add something like `Note: If you are trying to override luxHash, set luxDeps = fetchLuxDeps instead`?

nix/build-lux-package.nix
```diff
@@ -0,0 +1,182 @@
...
+    installPhase = ''
+      runHook preInstall
+
+      mkdir -p "$out/share/lux"
+      cp -rT "${buildSubdir}.lux" "$out/share/lux"
+
+      runHook postInstall
+    '';
+
+    checkPhase = ''
```
PerchunPak, line 162:
```suggestion
    checkPhase = lib.optionalString hasTestSpec ''
```

nix/import-lux-lock.nix
```diff
@@ -0,0 +1,85 @@
...
+  lockSections = [
+    "dependencies"
+    "test_dependencies"
+    "build_dependencies"
+  ];
+
+  rocks = lib.attrValues (
+    lib.listToAttrs (
+      map (rock: {
+        name = "${rock.name}@${rock.version}";
```
PerchunPak: Maybe this should follow Nix's conventions? Feel free to ignore if you think otherwise

```suggestion
        name = "${rock.name}-${rock.version}";
```

nix/import-lux-lock.nix
```diff
@@ -0,0 +1,85 @@
...
+        value = rock;
+      }) (
+        lib.flatten (
+          map (
+            section:
+              mapAttrsToList (_id: rock: rock) (lock.${section}.rocks or {})
```
PerchunPak:
```suggestion
              lib.attrValues (lock.${section}.rocks or {})
```

nix/import-lux-lock.nix
```diff
@@ -0,0 +1,85 @@
+{
+  lib,
+  runCommand,
+  fetchurl,
+  fetchgit,
+}: {
+  lockFile ? null,
+  lockFileContents ? null,
+}:
+assert (lockFile == null) != (lockFileContents == null); let
```
PerchunPak: Maybe use `lib.assertMsg`?

nix/build-lux-package.nix
mrcjkb: Good idea.

nix/build-lux-package.nix
```diff
@@ -0,0 +1,182 @@
...
+      then importLuxLock {lockFile = luxLock;}
+      else if luxHash != null
+      then
+        fetchLuxDeps {
+          inherit src luxRoot;
+          pname = finalAttrs.pname;
+          version = finalAttrs.version;
+          hash = luxHash;
+        }
+      else throw "buildLuxPackage requires either `luxHash`, `luxDeps`, `luxLock` or `luxVendorDir`";
```
teto: would be nice to show them in priority order : 
`"buildLuxPackage requires either  (sorted by precedence ) `luxVendorDir` `luxDeps`, `luxLock` or `luxHash``

teto:
looks promising o/

mrcjkb:
todo(follow-up): rust-mlua projects should also vendor cargo deps and `buildLux*` helpers should be able to handle them.

Edit: Done (#1864)

nix/build-lux-package.nix
```diff
@@ -0,0 +1,182 @@
...
+      then importLuxLock {lockFile = luxLock;}
+      else if luxHash != null
+      then
+        fetchLuxDeps {
+          inherit src luxRoot;
+          pname = finalAttrs.pname;
+          version = finalAttrs.version;
+          hash = luxHash;
+        }
+      else throw "buildLuxPackage requires either  (sorted by precedence) `luxVendorDir`, `luxDeps`, `luxLock` or `luxHash`";
```
PerchunPak:
```suggestion
      else throw "buildLuxPackage requires either (sorted by precedence) `luxVendorDir`, `luxDeps`, `luxLock` or `luxHash`";
```

PerchunPak (approved):

nix/overlay.nix
```diff
@@ -195,6 +292,64 @@
...
+    }
+    ./lux-loader-setup-hook.sh;
+
+  lua5_1 = override-lua prev.lua5_1 final.lux-lua51;
+  lua5_2 = override-lua prev.lua5_2 final.lux-lua52;
+  lua5_3 = override-lua prev.lua5_3 final.lux-lua53;
+  lua5_4 = override-lua prev.lua5_4 final.lux-lua54;
+  lua5_5 = override-lua prev.lua5_5 final.lux-lua55;
+  luajit = override-lua prev.luajit final.lux-luajit;
+  neovim = override-neovim prev.neovim-unwrapped;
```
khaneliman: This replaces the user's wrapped neovim with a completely new one without their configuration: `configure` is silently ignored (verified: same drv with and without it) and `wrapRc` flips to true. Separate attr instead of overriding `neovim`?

nix/overlay.nix
```diff
@@ -5,6 +5,95 @@
...
+        makeWrapperArgs =
+          lib.optionals (luxPackages != []) [
+            "--set"
+            "LUA_INIT"
+            "\"require('lux').loader()\""
+          ]
+          ++ luxPathArgs;
+      };
+  in
+    luaWithLuxLua.overrideAttrs (old: {
```
khaneliman, line 42: `overrideAttrs` doesn't reach the interpreter's self-reference, so anything indirect (`lua.pkgs.lua.withPackages`, neovim's wrapper) still gets the stock `withPackages` without the loader. Verified: direct and indirect spellings build different drvs.

nix/build-lux-rockspec.nix
```diff
@@ -0,0 +1,145 @@
...
+  ...
+}: let
+  luaVersionDir = lua.luaversion;
+  luaVersionFlag =
+    if nvim
+    then "--nvim"
+    else "--lua-version '${luaVersionDir}'";
+
+  # Matches lux's lockfile hash for an unpinned, unconstrained rock.
+  rockId = name: ver: builtins.hashString "sha256" "${name}${ver}falsefalse";
```
khaneliman: This duplicates lux's package-id hash and lock schema, and they already disagree: unnormalized versions ("0.1.2" vs lux's "0.1.2-1") make seeded deps invisible, and the missing `entrypoint_layout` fails `nvim = true` with `MismatchedRockLayout`. Could `lx` emit the tree instead, so nix never has to know the schema?

nix/build-lux-rockspec.nix
```diff
@@ -0,0 +1,145 @@
...
+    constraint = null;
+    binaries = [];
+    source = "local";
+    hashes = {
+      rockspec = dummyHash;
+      source = dummyHash;
+    };
+  };
+
+  lockfileJson = builtins.toJSON {
```
khaneliman: The hand-written lock omits `entrypoint_layout`, so it deserializes as the default rock layout, while `nvim = true` configures the nvim layout and `Lockfile::load` hard-errors on the mismatch (`MismatchedRockLayout`). As far as I can tell that makes `buildLuxRockspec { nvim = true; }` fail 100% of the time, and no test covers the nvim path here. The seeded tree dir also uses `lua.luaversion` ("5.1") while `--nvim` reads the "jit" tree.

nix/overlay.nix
```diff
@@ -5,6 +5,95 @@
...
+      args: let
+        luxPlugins = lib.filter (p: p ? luxPackage && p.luxPackage) (args.plugins or []);
+        luxLuaPath =
+          lib.concatMapStringsSep ";" (p: "${p}/share/lux/${p.luxLuaVersion}/?.lua;${p}/share/lux/${p.luxLuaVersion}/?/init.lua")
+          luxPlugins;
+      in
+        final.wrapNeovimUnstable neovim-unwrapped (
+          args
+          // {
+            extraLuaPackages = ps: ((args.extraLuaPackages or (_: [])) ps) ++ [final.lux-luajit.debug];
```
khaneliman: Every neovim gets a debug-profile lux-lua, even with no lux plugins:

```suggestion
            extraLuaPackages = ps: ((args.extraLuaPackages or (_: [])) ps) ++ lib.optionals (luxPlugins != []) [final.lux-luajit];
```

nix/build-lux-package.nix
```diff
@@ -0,0 +1,182 @@
...
+          inherit src luxRoot;
+          pname = finalAttrs.pname;
+          version = finalAttrs.version;
+          hash = luxHash;
+        }
+      else throw "buildLuxPackage requires either (sorted by precedence) `luxVendorDir`, `luxDeps`, `luxLock` or `luxHash`";
+
+    rootSubdir = lib.optionalString (luxRoot != null) "${luxRoot}/";
+
+    luxToml = fromTOML (lib.readFile "${src}/${rootSubdir}lux.toml");
```
khaneliman: This is IFD: `nix flake show` fetches the source and restricted eval fails outright. Same in build-lux-application.nix:11.

nix/build-lux-package.nix
```diff
@@ -0,0 +1,182 @@
...
+    buildPhase = ''
+      runHook preBuild
+
+      if (( ''${NIX_DEBUG:-0} >= 1 )); then
+        export RUST_LOG="trace"
+      fi
+
+      ${lib.optionalString (buildAndTestSubdir != null) "pushd ${buildAndTestSubdir}"}
+
+      ${lib.optionalString (luxLock != null) "cp ${luxLock} lux.lock"}
```
khaneliman: The store copy is read-only and `lx build`'s sync takes a write guard on the lockfile, so this fails with EACCES when `src` ships no lux.lock:

```suggestion
      ${lib.optionalString (luxLock != null) "install -m 644 ${luxLock} lux.lock"}
```

Related: its covering check (`build-lux-helpers-lux-lock`) is missing from the workflow's hand-kept list, so CI can't see this path at all.

.github/workflows/nix-build.yml
```diff
@@ -76,3 +76,15 @@ jobs:
       shell: bash
       # FIXME: This fails with permission denied on darwin
       if: ${{ matrix.job.target == 'x86_64-linux' }}
+    - name: Lux nix helper tests
+      run: |
+        nix build ".#checks.${{matrix.job.target}}.build-lux-helpers" -Lv --accept-flake-config --log-lines 500 --fallback
```
khaneliman: This hand-listed check set has already drifted from flake.nix: `build-lux-helpers-lux-lock` is defined there but never built here, so the whole `luxLock` vendoring path ships untested. Deriving the list from the checks attrset (or running `nix flake check`) would keep one source of truth.

nix/build-lux-rockspec.nix
```diff
@@ -0,0 +1,145 @@
...
+    '';
+  };
+
+  rockspecArg =
+    if knownRockspec != null
+    then "cp ${knownRockspec} ./pkg.rockspec"
+    else "cp \"$src/${pname}-${rockspecVersion}.rockspec\" ./pkg.rockspec";
+
+  drv = stdenv.mkDerivation {
+    name = "lua${luaVersionDir}-${pname}-${version}";
```
khaneliman, line 67: `toLuxNeovimPlugin (buildLuxRockspec ...)` fails with `attribute 'pname' missing`:

```suggestion
    name = "lua${luaVersionDir}-${pname}-${version}";
    inherit pname version;
```

Separately: as a plain function over mkDerivation, extra caller attrs (`meta`, `postInstall`, ...) are silently dropped; `lib.extendMkDerivation` like buildLuxPackage would fix that too.

nix/build-lux-application.nix
```diff
@@ -0,0 +1,52 @@
...
+          LUA_PATH="$(lx --lua-version "${lua.luaversion}" --tree "$out/share/lux" path lua)"
+          LUA_CPATH="$(lx --lua-version "${lua.luaversion}" --tree "$out/share/lux" path c)"
+          LUA_INIT="$(lx --lua-version "${lua.luaversion}" --tree "$out/share/lux" path init)"
+
+          makeWrapper "${interpreter}" "$out/bin/${args.pname}" \
+            --chdir "$out" \
+            --set LUA_INIT "$LUA_INIT" \
+            --set LUA_PATH "$LUA_PATH" \
+            --set LUA_CPATH "$LUA_CPATH" \
+            --add-flags ${lib.escapeShellArgs resolvedRunArgs}
```
khaneliman: Zero run args and two-plus run args both break makeWrapper here (tested: bare `--add-flags` dies on an unbound variable under `set -u`; extra words leak into makeWrapper's own arg parsing):

```suggestion
            --add-flags ${lib.escapeShellArg (lib.escapeShellArgs resolvedRunArgs)}
```

nix/lux-loader-setup-hook.sh
```diff
@@ -0,0 +1,12 @@
+export LUA_INIT="require('lux').loader()${LUA_INIT:+;$LUA_INIT}"
```
khaneliman: This exports LUA_INIT at hook-source time and the hook is propagated, so every Lua started in any downstream build runs `require('lux').loader()`, including plain `lua` invocations from unrelated tools (busted, luacheck, generators), which die at startup with `module 'lux' not found` since a LUA_INIT failure is fatal. Two lux packages for different lua versions also concatenate the hooks. It also only extends LUA_PATH, not LUA_CPATH, so the lux-lua .so is only reachable if separately propagated.

nix/build-lux-package.nix
```diff
@@ -0,0 +1,182 @@
...
+      then
+        fetchLuxDeps {
+          inherit src luxRoot;
+          pname = finalAttrs.pname;
+          version = finalAttrs.version;
+          hash = luxHash;
+        }
+      else throw "buildLuxPackage requires either (sorted by precedence) `luxVendorDir`, `luxDeps`, `luxLock` or `luxHash`";
+
+    rootSubdir = lib.optionalString (luxRoot != null) "${luxRoot}/";
```
khaneliman: `luxRoot` affects manifest reading, lock discovery, and fetchLuxDeps, but buildPhase only changes directory for `buildAndTestSubdir`, so a caller who sets only `luxRoot` (as the README describes it) evals fine and then `lx build` runs at the source root with no lux.toml. Deriving the build cwd from luxRoot (or folding the two args) would close it.

nix/build-lux-application.nix
```diff
@@ -0,0 +1,52 @@
...
+      "runArgs"
+      "runCommand"
+    ]
+    // {
+      nativeBuildInputs = args.nativeBuildInputs or [] ++ [makeWrapper];
+
+      postInstall =
+        (args.postInstall or "")
+        + ''
+          cp -r "$src/." "$out/"
```
khaneliman: Copying the whole source tree into `$out` ships tests/docs/C sources in every application closure (a 50 MB source means a 50 MB+ app) and can collide with the installed `share/lux` or leave read-only files behind. Also worth noting: `run.command` is wrapped as a literal program resolved from ambient PATH, while `lx run` resolves it with the tree's bin dir prepended, so `[run] command = "busted"` wraps a binary that won't exist at runtime. And line 11 reads `${args.src}/lux.toml` directly, ignoring `luxRoot`.

nix/import-lux-lock.nix
```diff
@@ -0,0 +1,75 @@
...
+
+  lock = builtins.fromJSON lockContents;
+
+  lockSections = [
+    "dependencies"
+    "test_dependencies"
+    "build_dependencies"
+  ];
+
+  rocks = lib.attrValues (
```
khaneliman: Re-keying rocks by `"${name}-${version}"` collapses two lock entries with the same name+version but different sources/hashes, with the survivor decided by section order. The lock's own ids are already unique keys, so `lib.attrValues (lib.mergeAttrsList (map (s: lock.${s}.rocks or {}) sections))` is both simpler and correct.

nix/build-lux-package.nix
```diff
@@ -0,0 +1,182 @@
...
+      runHook postInstall
+    '';
+
+    checkPhase = lib.optionalString hasTestSpec ''
+      runHook preCheck
+
+      ${lib.optionalString (buildAndTestSubdir != null) "pushd ${buildAndTestSubdir}"}
+
+      lx --profile release \
+         --lua-version "${lua.luaversion}" \
```
khaneliman: checkPhase hardcodes `--lua-version` instead of reusing `luaVersionFlag`, so an `nvim = true` package builds an nvim-layout jit tree and then tests against a fresh default-layout 5.1 tree instead of the built artifacts.

nix/overlay.nix
```diff
@@ -5,6 +5,95 @@
...
+    });
+
+  # Override neovim to activate the lux loader and expose lux trees on
+  # package.path when a plugin in the environment was built with lux.
+  override-neovim = neovim-unwrapped:
+    lib.makeOverridable (
+      args: let
+        luxPlugins = lib.filter (p: p ? luxPackage && p.luxPackage) (args.plugins or []);
+        luxLuaPath =
+          lib.concatMapStringsSep ";" (p: "${p}/share/lux/${p.luxLuaVersion}/?.lua;${p}/share/lux/${p.luxLuaVersion}/?/init.lua")
```
khaneliman: The filter above keeps anything with `luxPackage`, but this line reads `p.luxLuaVersion`, which only `toLuxNeovimPlugin` sets, so a raw lux package passed straight into `plugins` is an eval error (`attribute 'luxLuaVersion' missing`). Filter on `p ? luxLuaVersion`, or set the attr in the builders.

nix/build-lux-rockspec.nix
```diff
@@ -0,0 +1,145 @@
...
+        luaDepsList}
+    '';
+  };
+
+  rockspecArg =
+    if knownRockspec != null
+    then "cp ${knownRockspec} ./pkg.rockspec"
+    else "cp \"$src/${pname}-${rockspecVersion}.rockspec\" ./pkg.rockspec";
+
+  drv = stdenv.mkDerivation {
```
khaneliman: Since this is a plain function over a bare mkDerivation (unlike buildLuxPackage's `lib.extendMkDerivation`), every caller attribute beyond the named ones (`meta`, `env`, `patches`, `postInstall`, `doCheck`) is silently dropped, and there are no finalAttrs/overrideAttrs semantics. Driving it through extendMkDerivation like its sibling would fix that and deduplicate the shared name/inputs/installPhase shape.

nix/overlay.nix
```diff
@@ -203,71 +358,94 @@ in {
   };
   lux-lua51 =
```
khaneliman: Nit: these six lux-lua blocks differ only in luaPkg/isLuaJIT, plus the parallel lua5_x override list above encodes the same pairing again. One table plus `lib.mapAttrs` (and a shared `withDebug` helper, also reusable for lux-cli/lux-lsp) would make adding Lua 5.6 a one-line change instead of seven coordinated edits.

nix/build-lux-rockspec.nix
```diff
@@ -0,0 +1,145 @@
...
+        {
+          name = pkgId;
+          value = mkRock pname rockspecVersion (map (dep: dep.id) luaDepsList);
+        }
+      ]
+    );
+    entrypoints = [pkgId];
+  };
+
+  luxTree = stdenv.mkDerivation {
```
khaneliman: Nit: this can be `writeText` for the lockfile plus `linkFarm`/`runCommand` for the symlinks, dropping the full mkDerivation and the indented heredoc, whose terminator only lands at column 0 thanks to Nix indent-stripping (any re-indent breaks it at build time with an opaque bash EOF error).

nix/fetch-lux-deps.nix
```diff
@@ -0,0 +1,71 @@
...
+    "luxLock"
+    "luxRoot"
+  ];
+
+  extendDrvArgs = finalAttrs: {
+    pname ? "lux",
+    version ? "0",
+    hash,
+    src,
+    luxLock ? null,
```
khaneliman: Nit: `luxLock` here appears dead; buildLuxPackage routes lockfiles through `importLuxLock` instead, so nothing in-tree exercises this branch (and it shares the read-only `cp` issue). Removing it until something needs it would shrink the FOD's untested surface.

khaneliman (changes requested):

nix/overlay.nix
mrcjkb: This is just for our tests, and so that we have a reference if we upstream this to nixpkgs.

nix/build-lux-rockspec.nix
mrcjkb: Hmm, Lux does have the ability install vendored packed rocks, so maybe I can come up with something that uses `luarocks pack` to vendor the dependencies for Lux instead.

I'd rather avoid adding a dedicated `lx import-luarocks` command if I can.

nix/build-lux-rockspec.nix
```diff
@@ -0,0 +1,154 @@
...
+    installPhase = ''
+      runHook preInstall
+
+      mkdir -p "$out/share/lux"
+      cp -rT .lux "$out/share/lux"
+
+      runHook postInstall
+    '';
+
+    doCheck = false;
```
khaneliman: `extendMkDerivation` merges with the extended set winning, so this hardcoded value silently overrides a caller's `doCheck = true` (buildLuxPackage leaves it to the caller):

```suggestion
```

nix/build-lux-rockspec.nix
```diff
@@ -0,0 +1,154 @@
...
+              mkdir -p "$out/$ver"
+              cat > "$out/$ver/lux.lock" <<'LOCK'
+        ${lockfileJson}
+        LOCK
+              ${lib.concatMapStringsSep "\n" (dep: ''
+            mkdir -p "$out/$ver/${dep.id}-${dep.name}@${dep.version}"
+            ln -s ${dep.drv}/share/lua/${luaVersionDir} "$out/$ver/${dep.id}-${dep.name}@${dep.version}/src"
+            if [ -d ${dep.drv}/lib/lua/${luaVersionDir} ]; then
+              ln -s ${dep.drv}/lib/lua/${luaVersionDir} "$out/$ver/${dep.id}-${dep.name}@${dep.version}/lib"
+            fi
```
khaneliman: The `src` link isn't guarded like the `lib` one, so any dep without `share/lua/<ver>` gets a dangling symlink in the seeded tree:

```suggestion
            mkdir -p "$out/$ver/${dep.id}-${dep.name}@${dep.version}"
            if [ -d ${dep.drv}/share/lua/${luaVersionDir} ]; then
              ln -s ${dep.drv}/share/lua/${luaVersionDir} "$out/$ver/${dep.id}-${dep.name}@${dep.version}/src"
            fi
            if [ -d ${dep.drv}/lib/lua/${luaVersionDir} ]; then
              ln -s ${dep.drv}/lib/lua/${luaVersionDir} "$out/$ver/${dep.id}-${dep.name}@${dep.version}/lib"
            fi
```

Deeper question behind it, can a lux-built package be a dep of a rockspec build at all right now, given it ships `share/lux` rather than `share/lua`?

nix/overlay.nix
khaneliman: Makes sense. Since they still ship in `overlays.default`, would it be worth moving the test-only interpreter overrides out of the public overlay, or marking them not-API with a comment? As is, importing the overlay just for `lx` also rebuilds every lua consumer downstream.

nix/build-lux-rockspec.nix
khaneliman: That sounds like a better contract than nix reimplementing the schema. Happy to re-test the specrev and nvim cases once it lands.

nix/build-lux-rockspec.nix
mrcjkb: > The src link isn't guarded like the lib one

This will be gone when I implement `luarocks pack` (tested locally, and it works, just need to clean it up)

> Deeper question behind it, can a lux-built package be a dep of a rockspec build at all right now, given it ships share/lux rather than share/lua?

No. Lux will probably need a config option/flag to search other install trees to support that.

nix/build-lux-rockspec.nix
mrcjkb: Done.

nix/build-lux-package.nix
```diff
@@ -0,0 +1,178 @@
...
+      if nvim
+      then "--nvim"
+      else "--lua-version \"${lua.luaversion}\"";
+
+    lockFile =
+      if luxLock != null
+      then luxLock
+      else if lib.pathExists "${src}/${rootSubdir}lux.lock"
+      then "${src}/${rootSubdir}lux.lock"
+      else null;
```
khaneliman, line 72: I think this bit of the earlier IFD comment slipped through: probing `${src}` still realizes it at eval time whenever `luxLock` isn't passed, so restricted eval breaks on the default path even though the `lux.toml` reads are gone. Guarding the probe keeps auto-discovery for local sources (where it's pure) and skips it for fetched ones (where it's IFD):

```suggestion
    lockFile =
      if luxLock != null
      then luxLock
      else if !lib.isDerivation src && lib.pathExists "${src}/${rootSubdir}lux.lock"
      then "${src}/${rootSubdir}lux.lock"
      else null;
```

nix/build-lux-rockspec.nix
```diff
@@ -0,0 +1,120 @@
...
+    inherit (finalAttrs) pname version;
+
+    luaVersionDir = lua.luaversion;
+    luaVersionFlag =
+      if nvim
+      then "--nvim"
+      else "--lua-version '${luaVersionDir}'";
+
+    luaDeps = lib.filter (drv: drv ? luaModule) propagatedBuildInputs;
+    luarocksDeps = lib.filter (drv: !(drv ? luxPackage)) luaDeps;
```
khaneliman: Following up on your note that lux-built deps aren't supported here yet: this filter drops them without a word, so if the rockspec does declare one, the failure shows up later as a confusing resolve error. A warning keeps the legitimate case working (propagating a lux package the rockspec doesn't require) while flagging the broken one:

```suggestion
    luarocksDeps =
      lib.filter (
        drv:
          if drv ? luxPackage
          then lib.warn "buildLuxRockspec: lux-built dependency ${drv.pname} is not vendored; the build will fail if the rockspec requires it" false
          else true
      )
      luaDeps;
```

nix/build-lux-package.nix
```diff
@@ -0,0 +1,178 @@
...
+      runHook postInstall
+    '';
+
+    checkPhase = ''
+      runHook preCheck
+
+      ${lib.optionalString (buildAndTestSubdir != null) "pushd ${buildAndTestSubdir}"}
+
+      lx --profile release \
+         --lua-version "${lua.luaversion}" \
```
khaneliman: Think this one fell out of the earlier lua-version pass: checkPhase still pins `--lua-version` while buildPhase uses `luaVersionFlag`, so `nvim = true` with `doCheck = true` would test the default flavor and layout instead of what was just built.

```suggestion
         ${luaVersionFlag} \
```

nix/build-lux-rockspec.nix
mrcjkb: Opened #1871 for the ability to search other install trees.

khaneliman (approved):

PR #1864, opened 2026-08-19
feat(nix): support rust-mlua/cargo dependencies

mrcjkb:
Follow-up to #1862.

cc @teto @khaneliman @PerchunPak

### DIFF
--- a/nix/build-lux-package.nix
+++ b/nix/build-lux-package.nix
@@ -7,6 +7,9 @@
   luxLoaderSetupHook,
   pkg-config,
   lua,
+  cargo,
+  rustc,
+  writableTmpDirAsHomeHook,
 }:
 lib.extendMkDerivation {
   constructDrv = stdenv.mkDerivation;
@@ -17,6 +20,7 @@ lib.extendMkDerivation {
     "luxLock"
     "luxRoot"
     "luxVendorDir"
+    "rustSupport"
   ];
 
   transformDrv = drv:
@@ -31,6 +35,7 @@ lib.extendMkDerivation {
     luxDeps ? null,
     luxVendorDir ? null,
     luxRoot ? null,
+    rustSupport ? false,
     buildAndTestSubdir ? null,
     nvim ? false,
     nativeBuildInputs ? [],
@@ -38,6 +43,11 @@ lib.extendMkDerivation {
     postUnpack ? "",
     ...
   }: let
+    luaVersionFlag =
+      if nvim
+      then "--nvim"
+      else "--lua-version \"${lua.luaversion}\"";
+
     deps =
       if luxVendorDir != null
       then luxVendorDir
@@ -48,9 +58,10 @@ lib.extendMkDerivation {
       else if luxHash != null
       then
         fetchLuxDeps {
-          inherit src luxRoot;
+          inherit src luxRoot rustSupport nvim;
           pname = finalAttrs.pname;
           version = finalAttrs.version;
+          luaVersion = lua.luaversion;
           hash = luxHash;
         }
       else throw "buildLuxPackage requires either (sorted by precedence) `luxVendorDir`, `luxDeps`, `luxLock` or `luxHash`";
@@ -58,11 +69,6 @@ lib.extendMkDerivation {
     rootSubdir = lib.optionalString (luxRoot != null) "${luxRoot}/";
 
     buildSubdir = lib.optionalString (buildAndTestSubdir != null) "${buildAndTestSubdir}/";
-
-    luaVersionFlag =
-      if nvim
-      then "--nvim"
-      else "--lua-version \"${lua.luaversion}\"";
 
     lockFile =
       if luxLock != null
@@ -98,6 +104,11 @@ lib.extendMkDerivation {
         lux-cli
         pkg-config
         lua
+      ]
+      ++ lib.optionals rustSupport [
+        cargo
+        rustc
+        writableTmpDirAsHomeHook
       ]
       ++ nativeBuildInputs;
 
--- a/nix/build-lux-rockspec.nix
+++ b/nix/build-lux-rockspec.nix
@@ -5,8 +5,12 @@
   luaLib,
   lux-cli,
   luxLoaderSetupHook,
+  fetchLuxDeps,
   pkg-config,
+  cargo,
+  rustc,
   writeTextFile,
+  writableTmpDirAsHomeHook,
 }:
 lib.extendMkDerivation {
   constructDrv = stdenv.mkDerivation;
@@ -15,6 +19,10 @@ lib.extendMkDerivation {
     "knownRockspec"
     "rockspecVersion"
     "nvim"
+    "luxHash"
+    "luxDeps"
+    "luxVendorDir"
+    "rustSupport"
   ];
 
   transformDrv = drv:
@@ -27,6 +35,10 @@ lib.extendMkDerivation {
     knownRockspec ? null,
     rockspecVersion ? finalAttrs.version,
     nvim ? false,
+    luxHash ? null,
+    luxDeps ? null,
+    luxVendorDir ? null,
+    rustSupport ? false,
     nativeBuildInputs ? [],
     propagatedBuildInputs ? [],
     ...
@@ -59,10 +71,26 @@ lib.extendMkDerivation {
       text = lib.generators.toLua {asBindings = true;} luarocksConfig;
     };
 
-    rockspecArg =
+    rockspecFilename =
       if knownRockspec != null
-      then "cp ${knownRockspec} ./pkg.rockspec"
-      else "cp \"$src/${pname}-${rockspecVersion}.rockspec\" ./pkg.rockspec";
+      then knownRockspec
+      else "${pname}-${rockspecVersion}.rockspec";
+
+    vendoredDeps =
+      if luxVendorDir != null
+      then luxVendorDir
+      else if luxDeps != null
+      then luxDeps
+      else if luxHash != null
+      then
+        fetchLuxDeps
+        {
+          inherit src pname rustSupport knownRockspec rockspecFilename nvim;
+          version = rockspecVersion;
+          luaVersion = luaVersionDir;
+          hash = luxHash;
+        }
+      else null;
   in {
     name = "lua${luaVersionDir}-${pname}-${version}";
     inherit src;
@@ -79,6 +107,12 @@ lib.extendMkDerivation {
         pkg-config
         (lua.withPackages (ps: lib.optional (luarocksDeps != []) ps.luarocks))
       ]
+      ++ lib.optionals rustSupport
+      [
+        cargo
+        rustc
+        writableTmpDirAsHomeHook
+      ]
       ++ nativeBuildInputs;
 
     propagatedBuildInputs =
@@ -87,7 +121,7 @@ lib.extendMkDerivation {
     buildPhase = ''
       runHook preBuild
 
-      ${rockspecArg}
+      cp ${rockspecFilename} ./pkg.rockspec
 
       # Provide the package's own source through a vendor dir, so lux fetches
       # it locally instead of from the rockspec's remote URL.
@@ -102,8 +136,10 @@ lib.extendMkDerivation {
         '')
         luarocksDeps)}
 
+      ${lib.optionalString (vendoredDeps != null) "cp -r ${vendoredDeps}/. ./vendor/"}
+
       cp -r "$src" "./vendor/${pname}@${rockspecVersion}"
-      chmod -R u+w "./vendor/${pname}@${rockspecVersion}"
+      chmod -R u+w ./vendor
 
       lx --profile release \
          ${luaVersionFlag} \
--- a/nix/fetch-lux-deps.nix
+++ b/nix/fetch-lux-deps.nix
@@ -4,6 +4,9 @@
   lux-cli,
   git,
   cacert,
+  lua,
+  cargo,
+  writableTmpDirAsHomeHook,
 }:
 lib.extendMkDerivation {
   constructDrv = stdenvNoCC.mkDerivation;
@@ -12,6 +15,11 @@ lib.extendMkDerivation {
     "hash"
     "luxLock"
     "luxRoot"
+    "rockspecFilename"
+    "knownRockspec"
+    "luaVersion"
+    "nvim"
+    "rustSupport"
   ];
 
   extendDrvArgs = finalAttrs: {
@@ -21,9 +29,33 @@ lib.extendMkDerivation {
     src,
     luxLock ? null,
     luxRoot ? null,
+    rockspecFilename ? null,
+    knownRockspec ? null,
+    luaVersion ? null,
+    nvim ? false,
+    rustSupport ? false,
     nativeBuildInputs ? [],
     ...
-  }: {
+  }: let
+    luaVersionFlag =
+      if nvim
+      then "--nvim"
+      else lib.optionalString (luaVersion != null) "--lua-version '${luaVersion}'";
+    vendorCmd =
+      if rockspecFilename != null || knownRockspec != null
+      then ''
+        ${
+          if knownRockspec != null
+          then "cp ${knownRockspec} ./pkg.rockspec"
+          else "cp \"$src/${rockspecFilename}\" ./pkg.rockspec"
+        }
+
+        lx ${luaVersionFlag} vendor "$out" --rockspec ./pkg.rockspec --no-delete
+      ''
+      else ''
+        lx ${luaVersionFlag} vendor "$out" --no-delete
+      '';
+  in {
     name = "${pname}-${version}-vendor-deps";
 
     inherit src;
@@ -31,8 +63,13 @@ lib.extendMkDerivation {
     nativeBuildInputs =
       [
         lux-cli
+        lua
         git
         cacert
+      ]
+      ++ lib.optional rustSupport [
+        cargo
+        writableTmpDirAsHomeHook
       ]
       ++ nativeBuildInputs;
 
@@ -46,11 +83,13 @@ lib.extendMkDerivation {
     buildPhase = ''
       runHook preBuild
 
+      mkdir -p $out
+
       ${lib.optionalString (luxRoot != null) "cd ${luxRoot}"}
 
       ${lib.optionalString (luxLock != null) "cp ${luxLock} lux.lock"}
 
-      lx vendor "$out" --no-delete
+      ${vendorCmd}
 
       runHook postBuild
     '';
--- a/nix/tests/lux-project.nix
+++ b/nix/tests/lux-project.nix
@@ -1,0 +1,59 @@
+{pkgs}: let
+  src = pkgs.runCommandLocal "toml-edit-wrapper-src" {} ''
+        mkdir -p "$out/src"
+        cat > "$out/lux.toml" <<'TOML'
+    package = "toml-edit-wrapper"
+    version = "0.1.0"
+    lua = ">=5.1"
+
+    [description]
+    summary = "test"
+    maintainer = "lumen-oss"
+    license = "MIT"
+
+    [dependencies]
+    toml-edit = "0.7.0-1"
+
+    [build]
+    type = "builtin"
+
+    [build.modules]
+    toml_edit_wrapper = "src/wrapper.lua"
+    TOML
+        cat > "$out/src/wrapper.lua" <<'LUA'
+    local toml_edit = require("toml_edit")
+    return {
+      parse = function(s) return toml_edit.parse(s) end,
+    }
+    LUA
+  '';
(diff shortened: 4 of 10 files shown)


nix/build-lux-package.nix
```diff
@@ -136,6 +141,9 @@ lib.extendMkDerivation {
     buildPhase = ''
       runHook preBuild
 
+      export CARGO_HOME="$NIX_BUILD_TOP/.cargo-home"
+      mkdir -p "$CARGO_HOME"
+
```
mrcjkb: I guess we could add a writeableTmpDirAsHomeHook instead?
But it only complains about CARGO_HOME...

Thoughts?

khaneliman: Personally, have preferred utilizing the hook wherever it makes sense just to keep concerns separated from custom packaging installation commands from generic sandbox concepts.

nix/fetch-lux-deps.nix
```diff
@@ -21,9 +25,27 @@ lib.extendMkDerivation {
...
+          if knownRockspec != null
+          then "cp ${knownRockspec} ./pkg.rockspec"
+          else "cp \"$src/${rockspecFilename}\" ./pkg.rockspec"
+        }
+
+        lx --lua-version ${luaVersion} vendor "$out" --rockspec ./pkg.rockspec --no-delete
+      ''
+      else ''
+        lx --lua-version ${luaVersion} vendor "$out" --no-delete
+      '';
```
khaneliman: This doesn't look guarded by the presence of a `luaVersion` and could cause an issue like `cannot coerce null to a string` when not provided.

nix/fetch-lux-deps.nix
```diff
@@ -21,9 +25,27 @@ lib.extendMkDerivation {
...
+    vendorCmd =
+      if rockspecFilename != null || knownRockspec != null
+      then ''
+        ${
+          if knownRockspec != null
+          then "cp ${knownRockspec} ./pkg.rockspec"
+          else "cp \"$src/${rockspecFilename}\" ./pkg.rockspec"
+        }
+
+        lx --lua-version ${luaVersion} vendor "$out" --rockspec ./pkg.rockspec --no-delete
```
khaneliman: This doesn't look guarded by the presence of a `luaVersion` and could cause an issue like `cannot coerce null to a string` when not provided. Could bind one flag and reuse it in both branches, which also quotes the value like the other call sites:

```nix
luaVersionFlag = lib.optionalString (luaVersion != null) "--lua-version '${luaVersion}'";
```

nix/build-lux-package.nix
```diff
@@ -51,6 +53,7 @@ lib.extendMkDerivation {
           inherit src luxRoot;
           pname = finalAttrs.pname;
           version = finalAttrs.version;
+          luaVersion = lua.luaversion;
```
khaneliman: When `nvim = true`, the build below runs with `--nvim` (which resolves LuaJIT), but this vendor step always pins `--lua-version ${lua.luaversion}` (nixpkgs luajit reports "5.1"). If any rock in the graph resolves differently between the two, the offline build fails on a missing vendored rock. Same mismatch in `build-lux-rockspec.nix`. Passing `nvim` through to `fetchLuxDeps` would close it; the `nvim` + `luxHash` combination currently has no test coverage (`nvim-plugin.nix` uses `luxVendorDir`).

nix/build-lux-package.nix
```diff
@@ -102,6 +105,8 @@ lib.extendMkDerivation {
         lux-cli
         pkg-config
         lua
+        cargo
+        rustc
```
khaneliman: This pulls the rust toolchain (~1.8 GB closure for cargo + rustc) into every lux package, including pure-Lua ones. `luxToml` is already parsed at eval time above, so this could be gated:

```nix
++ lib.optionals (rustSupport || (luxToml.build.type or "builtin") == "rust-mlua") [cargo rustc]
```

with a `rustSupport ? false` opt-in for the case where a *dependency* is rust-mlua.

nix/build-lux-rockspec.nix
```diff
@@ -95,25 +101,50 @@
...
+        )
+      )
+    else null;
+
   drv = stdenv.mkDerivation {
     name = "lua${luaVersionDir}-${pname}-${version}";
     inherit src;
 
-    nativeBuildInputs = [lux-cli pkg-config lua] ++ nativeBuildInputs;
+    nativeBuildInputs = [lux-cli pkg-config lua cargo rustc] ++ nativeBuildInputs;
```
khaneliman: Same toolchain-closure concern as in `build-lux-package.nix`, but here there is no manifest to inspect and no opt-out, so the existing pure C/Lua checks (cjson, luassert) now pay the rust closure per CI target. A `rustSupport ? false` arg (set by the new toml-edit test) would keep pure-Lua rockspec builds lean.

nix/build-lux-rockspec.nix
```diff
@@ -95,25 +101,50 @@
...
+          luaVersion = luaVersionDir;
+          hash = luxHash;
+        }
+        // (
+          if knownRockspec != null
+          then {inherit knownRockspec;}
+          else {rockspecFilename = "${pname}-${rockspecVersion}.rockspec";}
+        )
+      )
+    else null;
```
khaneliman: This duplicates the chain from `build-lux-package.nix` but drops the `luxLock` branch and replaces the `throw` with `null`. Since the builder takes `...`, `luxLock = ./lux.lock` (documented for `buildLuxPackage`) is silently ignored and the build fails much later inside `lx install-rockspec` with a network error; a typo like `luxDep` fails the same way. It also skips the stale-vendor `postUnpack` guard. Extracting one shared resolver both builders call (with an allow-none mode for dependency-free rockspecs) would keep the precedence contract, the error message, and the guard in one place.

nix/build-lux-rockspec.nix
```diff
@@ -95,25 +101,50 @@
...
+          pname = pname;
+          version = rockspecVersion;
+          luaVersion = luaVersionDir;
+          hash = luxHash;
+        }
+        // (
+          if knownRockspec != null
+          then {inherit knownRockspec;}
+          else {rockspecFilename = "${pname}-${rockspecVersion}.rockspec";}
+        )
```
khaneliman: The filename convention `${pname}-${rockspecVersion}.rockspec` is now encoded here, at line 102, and re-derived inside `fetch-lux-deps.nix`. A single `rockspecFilename` let-binding could feed both, and this `//` merge can collapse to a flat attrset since `fetchLuxDeps` already defaults both args to `null` and prefers `knownRockspec`.

nix/build-lux-rockspec.nix
```diff
@@ -95,25 +101,50 @@
...
+
       ${rockspecArg}
 
-      # Provide the package's own source through a vendor dir, so lux fetches
-      # it locally instead of from the rockspec's remote URL.
       mkdir -p ./vendor
+      ${lib.optionalString (vendoredDeps != null) "cp -r ${vendoredDeps}/. ./vendor/"}
       cp -r "$src" "./vendor/${pname}@${rockspecVersion}"
-      chmod -R u+w "./vendor/${pname}@${rockspecVersion}"
+      chmod -R u+w ./vendor
```
khaneliman: Two issues with the ordering here:

1. The deps tree is copied out of the store read-only, and `$src` is copied before the `chmod`. If the vendored tree already contains `${pname}@${rockspecVersion}` (possible with a user-supplied `luxVendorDir`/`luxDeps`), the `cp` either fails with `Permission denied` (reproduced locally with read-only sources) or, if the entry is writable, nests `$src` inside it so lux builds the remote-fetched source instead of `$src`.
2. The `chmod -R u+w` widened from the single package dir to the whole vendor tree, which for a cargo vendor dir means re-walking tens of thousands of files per build.

Suggest: chmod right after the deps copy, `rm -rf` the target entry, then copy `$src` and chmod only that dir. For large cargo trees a symlink farm (`ln -s ${vendoredDeps}/* ./vendor/`) would avoid the copy entirely; the `buildLuxPackage` path already proves lux accepts a read-only vendor dir. Also, the deleted comment about why `$src` is copied into `./vendor` still describes this live line; without it the self-copy looks redundant next to the new deps copy and is at risk of being cleaned up later.

nix/build-lux-package.nix
```diff
@@ -136,6 +141,9 @@ lib.extendMkDerivation {
     buildPhase = ''
       runHook preBuild
 
+      export CARGO_HOME="$NIX_BUILD_TOP/.cargo-home"
+      mkdir -p "$CARGO_HOME"
```
khaneliman: This export is now copy-pasted into three `buildPhase` strings and never applies to `checkPhase`, where `lx test` can invoke cargo for rust-mlua test deps with `CARGO_HOME` unset (`$HOME` is `/homeless-shelter`). A small setup hook via `makeSetupHook` (same pattern as `luxLoaderSetupHook` in `overlay.nix`) registered in `preConfigureHooks` would cover all phases in all three builders and hold future cargo env vars in one place.

nix/tests/lux-project.nix
```diff
@@ -0,0 +1,58 @@
...
+    [build.modules]
+    toml_edit_wrapper = "src/wrapper.lua"
+    TOML
+        cat > "$out/src/wrapper.lua" <<'LUA'
+    local toml_edit = require("toml_edit")
+    return {
+      parse = function(s) return toml_edit.parse(s) end,
+    }
+    LUA
+  '';
```
khaneliman: The heredoc bodies only reach column 0 through Nix indented-string dedent from the current 4-vs-8 space split; any re-indent of the shell lines indents the `TOML`/`LUA` terminators and bash fails with `unexpected EOF`. Building `src` with `runCommandLocal` also makes the `fromTOML` read in `build-lux-package.nix` an import-from-derivation, which breaks restricted eval. Committing the fixture as real files under `nix/tests/fixtures/toml-edit-wrapper/` and using `src = ./fixtures/toml-edit-wrapper;` avoids both and makes the pinned `toml-edit = "0.7.0-1"` constraint diffable.

nix/build-lux-package.nix
teto: dont care either way. grepping nixpkgs, couldn't find CARGO_HOME used with `writeableTmpDirAsHomeHook` (but writeableTmpDirAsHomeHook is relatively recent) so current code is fine.

mrcjkb: > keep concerns separated from custom packaging installation commands from generic sandbox concepts

Makes sense to me. Switched to the hook.

nix/build-lux-package.nix
```diff
@@ -51,6 +54,7 @@ lib.extendMkDerivation {
           inherit src luxRoot;
           pname = finalAttrs.pname;
           version = finalAttrs.version;
+          luaVersion = lua.luaversion;
```
khaneliman: Question on the lua version here. Without a lockfile, `lx vendor` resolves against a Lua-version-specific manifest, and the build below runs with `--nvim` instead when `nvim = true`. LuaJIT resolves via the 5.1 manifest so the window is narrow, but if the effective versions ever differ, vendor and build disagree and the offline build fails on a missing rock. Forwarding the same flag the build computes would rule it out. Nothing tests nvim together with luxHash today.

nix/build-lux-package.nix
```diff
@@ -102,6 +106,9 @@ lib.extendMkDerivation {
         lux-cli
         pkg-config
         lua
+        cargo
+        rustc
```
khaneliman: This puts the rust toolchain into every lux package's build closure, pure-Lua ones included. cargo plus rustc is about 1.8 GB on my machine. `luxToml` is already parsed at eval time above, so it could be gated:

```nix
++ lib.optionals (rustSupport || (luxToml.build.type or "builtin") == "rust-mlua") [cargo rustc]
```

with `rustSupport ? false` for the case where a dependency is rust-mlua rather than the package itself.

nix/build-lux-rockspec.nix
```diff
@@ -95,11 +102,43 @@
...
     inherit src;
 
-    nativeBuildInputs = [lux-cli pkg-config lua] ++ nativeBuildInputs;
+    nativeBuildInputs =
+      [
+        lux-cli
+        pkg-config
+        lua
+        cargo
+        rustc
```
khaneliman: Same rust toolchain cost, and this builder has no manifest to inspect and no opt-out, so the cjson and luassert checks now build with the toolchain too. A `rustSupport ? false` arg that only the toml-edit test sets would keep pure-Lua rockspec builds lean.

nix/build-lux-rockspec.nix
```diff
@@ -95,11 +102,43 @@
...
+          luaVersion = luaVersionDir;
+          hash = luxHash;
+        }
+        // (
+          if knownRockspec != null
+          then {inherit knownRockspec;}
+          else {rockspecFilename = "${pname}-${rockspecVersion}.rockspec";}
+        )
+      )
+    else null;
```
khaneliman: Should `buildLuxRockspec` support `luxLock` the way `buildLuxPackage` does, or is the omission intentional? The two builders now carry slightly different source-precedence chains. If it stays out on purpose, a note in the README would help, since the vendoring docs present the sources as one set.

nix/build-lux-rockspec.nix
```diff
@@ -95,11 +102,43 @@
...
+          pname = pname;
+          version = rockspecVersion;
+          luaVersion = luaVersionDir;
+          hash = luxHash;
+        }
+        // (
+          if knownRockspec != null
+          then {inherit knownRockspec;}
+          else {rockspecFilename = "${pname}-${rockspecVersion}.rockspec";}
+        )
```
khaneliman: The `${pname}-${rockspecVersion}.rockspec` filename is built here and at line 103; one let-binding could serve both. The `//` merge can also flatten out, since `fetchLuxDeps` defaults both args to `null` and already prefers `knownRockspec`.

nix/build-lux-rockspec.nix
```diff
@@ -109,11 +148,10 @@
 
       ${rockspecArg}
 
-      # Provide the package's own source through a vendor dir, so lux fetches
-      # it locally instead of from the rockspec's remote URL.
       mkdir -p ./vendor
+      ${lib.optionalString (vendoredDeps != null) "cp -r ${vendoredDeps}/. ./vendor/"}
       cp -r "$src" "./vendor/${pname}@${rockspecVersion}"
-      chmod -R u+w "./vendor/${pname}@${rockspecVersion}"
+      chmod -R u+w ./vendor
```
khaneliman: Three things on this block.

The `chmod -R u+w` widened from the single package dir to the whole vendor tree, which now re-walks every vendored file (including a cargo vendor dir) on each build. Narrowing it back to `./vendor/${pname}@${rockspecVersion}` looks sufficient; lux copies vendored sources into the build dir itself, so the deps don't appear to need writability. Symlinking the deps instead of copying could avoid the copy too, though I haven't verified that against every source form.

The deleted comment about why `$src` gets copied into `./vendor` still describes this line. Without it the self-copy looks redundant next to the new deps copy, and someone will eventually clean it up and reintroduce fetching the rock's own source from the rockspec URL. Worth restoring.

Defensive nit: if a supplied vendor dir ever contains `${pname}@${rockspecVersion}`, this `cp` lands on a read-only store copy and dies with Permission denied (tried it locally). A `rm -rf` of the target entry first closes that.

nix/build-lux-rockspec.nix
mrcjkb: For now, it's intentional. LuaRocks projects don't check in a `lux.lock`, so if there's a lockfile, it's a Lux project and you can use `buildLuxPackage`.

A `lx generate-lockfile` command for arbitrary rockspecs/packages on luarocks.org could be useful one day, but yagni for now?

khaneliman: Sorry, deleted a few of my own comments while cleaning up my review and took the parent of this one with it. It was my question about whether buildLuxRockspec should support luxLock like buildLuxPackage. The README note in the latest push answers it, thanks.

nix/build-lux-rockspec.nix
```diff
@@ -90,30 +98,53 @@
...
+    then luxDeps
+    else if luxHash != null
+    then
+      fetchLuxDeps
+      {
+        inherit src pname rustSupport knownRockspec;
+        version = rockspecVersion;
+        luaVersion = luaVersionDir;
+        hash = luxHash;
+      }
```
khaneliman: The flattened call drops the rockspec filename: when `knownRockspec` is null, neither `knownRockspec` nor `rockspecFilename` reaches `fetchLuxDeps`, so its `vendorCmd` falls into the project branch and runs `lx vendor` without `--rockspec`. A rockspec build that relies on `<pname>-<version>.rockspec` inside `src` then vendors against a nonexistent lux project. The toml-edit test sets `knownRockspec`, so CI doesn't cover that path. `inherit rockspecFilename;` here should restore it; when `knownRockspec` is set it still wins inside `fetchLuxDeps`.

nix/build-lux-package.nix
```diff
@@ -48,9 +53,10 @@ lib.extendMkDerivation {
       else if luxHash != null
       then
         fetchLuxDeps {
-          inherit src luxRoot;
+          inherit src luxRoot rustSupport;
           pname = finalAttrs.pname;
           version = finalAttrs.version;
+          luaVersion = lua.luaversion;
```
khaneliman: Reposting a question I deleted by accident: does resolution during `lx vendor` depend on the lua version? Without a lockfile it resolves against a Lua-version-specific manifest, and the build below runs with `--nvim` instead when `nvim = true`. LuaJIT resolves via the 5.1 manifest so the window is narrow, but if the effective versions ever differ, vendor and build disagree and the offline build fails on a missing rock. Forwarding the same flag the build computes would rule it out. Nothing tests nvim together with luxHash today.

nix/build-lux-rockspec.nix
```diff
@@ -90,30 +98,53 @@
...
+  vendoredDeps =
+    if luxVendorDir != null
+    then luxVendorDir
+    else if luxDeps != null
+    then luxDeps
+    else if luxHash != null
+    then
+      fetchLuxDeps
+      {
+        inherit src pname rustSupport knownRockspec nvim;
```
khaneliman: This lost `rockspecFilename`: when `knownRockspec` is null, `lx vendor` runs without `--rockspec`, so it either errors ("must be run in a workspace root or with a rockspec argument") or, if `src` ships a `lux.toml`, silently vendors the project's deps instead of the rockspec's. CI misses it because the toml-edit test sets `knownRockspec`.

```suggestion
        inherit src pname rustSupport knownRockspec rockspecFilename nvim;
```

nix/fetch-lux-deps.nix
```diff
@@ -21,9 +28,33 @@ lib.extendMkDerivation {
...
+    nvim ? false,
+    rustSupport ? false,
     nativeBuildInputs ? [],
     ...
-  }: {
+  }: let
+    luaVersionFlag =
+      if nvim
+      then "--nvim"
+      else lib.optionalString (luaVersion != null) "--lua-version '${luaVersion}'";
```
khaneliman, line 43: What does `lx --nvim vendor` do when the probe finds no Lua? With #1861, `--nvim` probes for an installed LuaJIT or 5.1 rather than assuming 5.1, but this sandbox has no Lua in `nativeBuildInputs`, while the build sandbox does. If that makes the two resolve dependencies differently, would the offline build fail on a missing rock; adding the same `lua` here when `nvim` is set would keep them aligned.

khaneliman: Tested this on the branch and it hard-fails. In an env with no `lua`/`luajit` discoverable (mimicking the fetchLuxDeps sandbox)

```sh
lx --nvim vendor ./out --rockspec luassert-1.9.0-1.rockspec --no-delete
```
...
```
error initialising remote package DB
  ├─▶ failed to fetch manifest from https://luarocks.org/:
  │   lua version not set
```

Same rockspec vendors fine with `--lua-version 5.1`, so it's purely the probe coming up empty, a nvim + luxHash test variant would pin this behavior if we want to red-green prove it out in CI.

khaneliman (approved):

khaneliman:
lol I think worked in the wrong direction, taking a look later at https://github.com/lumen-oss/lux/pull/1862 too

khaneliman (approved):

nix/build-lux-package.nix
```diff
@@ -98,7 +104,9 @@ lib.extendMkDerivation {
         lux-cli
         pkg-config
         lua
+        writableTmpDirAsHomeHook
       ]
+      ++ lib.optionals rustSupport [cargo rustc]
```
PerchunPak: Is `writableTmpDirAsHomeHook` required, or can it be placed under `lib.optionals rustSupport`?

PR #1883, opened 2026-08-26
feat!: compatibility with luarocks arbitrary versioning

mrcjkb:
## Problem:

Popular luarocks packages, like [openresty's lua-cjson fork](https://github.com/openresty/lua-cjson), publish releases with non-semver compliant version components (e.g. `2.1.0.9`).
As a result, we can't build lua-cjson on luajit, because Lux sees version `2.1.0` as the latest.

## Solution

(reluctantly) add support for Luarocks's arbitrary version components.
We still use `semver` under the hood, but we allow comparing versions with a prerelease component with versions that don't have a prerelease component.

@vhyrro this is a blocker for being able to build lux.nvim (which depends on lua-cjson via fallo) against luajit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,10 +116,11 @@ The following table provides a brief comparison:
 | Sandbox build execution                                                                 | :white_check_mark: [^4]      | :x:                |
 
 [^1]: Supported via a compatibility layer that uses LuaRocks as a backend.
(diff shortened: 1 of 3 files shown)


vhyrro (approved):
unfortunate but necessary

PR #1884, opened 2026-08-27
feat(lx-new): generate source template

### LINKED ISSUE #1168: [Feature] generate source template with `lx new`
See https://lux.lumen-labs.org/guides/lux-toml#source

Isvane:
Close #1168 

Following: [https://lux.lumen-labs.org/guides/lux-toml/#source](https://lux.lumen-labs.org/guides/lux-toml/#source)

New toml generated from lx-new:
```toml
package = "isvane"
version = "0.1.0"
lua = ">=5.1"

[description]
summary = "testing"
maintainer = "haiqal"
labels = [ "web" ]
license = "MIT"

[source]
url = "https://github.com/haiqal/isvane/archive/refs/tags/$(REF).zip"

[dependencies]
# Add your dependencies here
# `busted = ">=2.0"`

[run]
args = [ "src/main.lua" ]

[build]
type = "builtin"
```

- [x] Fits [CONTRIBUTING.md].

### DIFF
--- a/lux-cli/src/project/new.rs
+++ b/lux-cli/src/project/new.rs
@@ -313,7 +313,7 @@ pub async fn write_project_rockspec(cli_flags: NewProject, config: Config) -> Re
             )?;
 
             let mut source_block =
-                format!("# [source]\n# When publishing, configure the source URL here");
+                "# [source]\n# When publishing, configure the source URL here".to_string();
 
             if is_github_repo {
                 let generated_url = format!(
--- a/lux-cli/src/project/new.rs
+++ b/lux-cli/src/project/new.rs
@@ -336,7 +336,7 @@ pub async fn write_project_rockspec(cli_flags: NewProject, config: Config) -> Re
         format!("[source]\nurl = \"{generated_url}\"")
     } else {
         format!(
-                "# [source]\n# url = \"https://github.com/your-username/{}/archive/refs/tags/$(REF).zip\"",
+                "# [source]\n# When publishing, configure the source URL below\n# url = \"https://github.com/your-username/{}/archive/refs/tags/$(REF).zip\"",
                 validated.name
             )
     };
--- a/lux-cli/src/project/new.rs
+++ b/lux-cli/src/project/new.rs
@@ -325,6 +325,17 @@ pub async fn write_project_rockspec(cli_flags: NewProject, config: Config) -> Re
 
     let rocks_path = validated.target.join(PROJECT_TOML);
 
+    let generated_url = format!(
+        "https://github.com/{}/{}/archive/refs/tags/$(REF).zip",
+        validated.maintainer, validated.name
+    );
+
+    let source = if !generated_url.contains(' ') && url::Url::parse(&generated_url).is_ok() {
+        format!("[source]\nurl = \"{generated_url}\"")
+    } else {
+        String::new()
+    };
+
     std::fs::write(
         &rocks_path,
         format!(
@@ -339,8 +350,7 @@ maintainer = "{maintainer}"
 labels = [ {labels} ]
 {license}
 
-[source]
-url = "https://github.com/<owner>/my-project/archive/refs/tags/$(REF).zip"
+{source}
 
 [dependencies]
 # Add your dependencies here
--- a/lux-cli/src/project/new.rs
+++ b/lux-cli/src/project/new.rs
@@ -81,6 +81,7 @@ struct NewProjectValidated {
     lua_versions: PackageReq,
     main: SourceDirType,
     license: Option<LicenseId>,
+    is_github_repo: bool,
 }
 
 fn clap_parse_license(s: &str) -> std::result::Result<LicenseId, String> {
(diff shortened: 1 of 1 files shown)


mrcjkb (changes requested):
Thanks for the PR 🙏 

I'd suggest to check if the generated URL parses correctly, and exclude it if it doesn't.

Isvane:
> Thanks for the PR 🙏
> 
> I'd suggest to check if the generated URL parses correctly, and exclude it if it doesn't.

Done :saluting_face: 

I used url crate for parsing the source links, it'll skip the source block if there's space on it.

lux-cli/src/project/new.rs
```diff
@@ -325,6 +325,17 @@ pub async fn write_project_rockspec(cli_flags: NewProject, config: Config) -> Re
 
     let rocks_path = validated.target.join(PROJECT_TOML);
 
+    let generated_url = format!(
+        "https://github.com/{}/{}/archive/refs/tags/$(REF).zip",
+        validated.maintainer, validated.name
+    );
+
+    let source = if !generated_url.contains(' ') && url::Url::parse(&generated_url).is_ok() {
```
mrcjkb: suggestion: I don't have a way to test this right now, but I'd assume `Url::parse` would fail if the URL string contains spaces?

Isvane: i also think that at first, but when i removed the !contains check like so:
```rust
let source = if url::Url::parse(&generated_url).is_ok() {
        format!("[source]\nurl = \"{generated_url}\"")
    } else {
        String::new()
    };
```

it still generates the links:
```toml
[source]
url = "https://github.com/jane doe/isvane/archive/refs/tags/$(REF).zip"
```
i think we still need it though, since !contains only check for spaces

mrcjkb: Right, so parsing it probably percent-encodes the space.
Perhaps a more general validity check would be:

```rust
url::Url::parse(&generated_url).is_ok_and(|url| &url.to_string() == &generated_url)
```

Isvane: > Right, so parsing it probably percent-encodes the space. Perhaps a more general validity check would be:
> 
> ```rust
> url::Url::parse(&generated_url).is_ok_and(|url| &url.to_string() == &generated_url)
> ```

Done :saluting_face: 

Also decided to give user a placeholder instead of just skipping when it fails.

lux-cli/src/project/new.rs
```diff
@@ -325,6 +325,22 @@ pub async fn write_project_rockspec(cli_flags: NewProject, config: Config) -> Re
...
+        validated.maintainer, validated.name
+    );
+
+    let source = if url::Url::parse(&generated_url)
+        .is_ok_and(|url| &url.to_string() == &generated_url)
+    {
+        format!("[source]\nurl = \"{generated_url}\"")
+    } else {
+        format!(
+                "# [source]\n# When publishing, configure the source URL below\n# url = \"https://github.com/your-username/{}/archive/refs/tags/$(REF).zip\"",
```
mrcjkb:
```suggestion
                "# [source]\n# When publishing, configure the source URL here
```

lux-cli/src/project/new.rs
```diff
@@ -325,6 +325,22 @@ pub async fn write_project_rockspec(cli_flags: NewProject, config: Config) -> Re
 
     let rocks_path = validated.target.join(PROJECT_TOML);
 
+    let generated_url = format!(
+        "https://github.com/{}/{}/archive/refs/tags/$(REF).zip",
```
mrcjkb: issue: this currently favours github over other source forges.
We should only set a GitHub URL if we've actually detected github repo metadata.

mrcjkb (changes requested):

lux-cli/src/project/new.rs
Isvane: done :saluting_face:

lux-cli/src/project/new.rs
```diff
@@ -308,6 +312,20 @@ pub async fn write_project_rockspec(cli_flags: NewProject, config: Config) -> Re
                 Ok,
             )?;
 
+            let mut source_block =
+                format!("# [source]\n# When publishing, configure the source URL here");
```
mrcjkb: issue: clippy is failing:

```
       > error: useless use of `format!`
       >    --> lux-cli/src/project/new.rs:316:17
       >     |
       > 316 |                 format!("# [source]\n# When publishing, configure the source URL here");
       >     |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: consider using `.to_string()`: `"# [source]\n# When publishing, configure the source URL here".to_string()`
```

mrcjkb (changes requested):

lux-cli/src/project/new.rs
Isvane: my bad

mrcjkb (approved):

PR #1901, opened 2026-08-31
fix(fetch): properly checkout custom branches

vhyrro:


lux-lib/src/operations/fetch.rs
```diff
@@ -292,7 +292,12 @@ async fn fetch_src_impl<R: Rockspec>(
 
                 let checkout_ref = match checkout_ref {
                     Some(checkout_ref) => {
-                        let (object, _) = repo.revparse_ext(checkout_ref)?;
+                        let ref_name = if let Some(GitRef::Branch(branch)) = &git.git_ref {
+                            format!("origin/{branch}")
```
mrcjkb, line 296: question: Can we use libgit to determine the remote name, instead of assuming `origin`?

Git has a `ls-remote` command for that.

mrcjkb: yagni - as discussed on matrix

mrcjkb (approved):

PR #1914, opened 2026-09-03
feat!: relocate `src/` directory on `--nvim`

vhyrro:
Fixes code loading problems in `lux.nvim`. Would we consider this breaking? I've maintained backwards compatibility with the old lockfile format with a serde alias.

### DIFF
--- a/lux-lib/src/config/tree.rs
+++ b/lux-lib/src/config/tree.rs
@@ -4,20 +4,25 @@ use std::path::PathBuf;
 /// Template configuration for a rock's tree layout
 #[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
 pub struct RockLayoutConfig {
-    /// The root of a packages `etc` directory.
-    /// If unset (the default), the root is the package root.
+    /// The root for relocated directories (etc, src).
+    /// If unset (the default), directories are placed in the package directory.
     /// If set, it is a directory relative to the given Lua version's install tree root.
     /// With the `--nvim` preset, this is `site/pack/lux`.
-    pub(crate) etc_root: Option<PathBuf>,
+    #[serde(alias = "etc_root")]
+    pub(crate) root: Option<PathBuf>,
     /// The `etc` directory for non-optional packages
     /// Default: `etc` With the `--nvim` preset, this is `start`
-    /// Note: If `etc_root` is set, the package ID is appended.
+    /// Note: If `root` is set, the package ID is appended.
     pub(crate) etc: PathBuf,
     /// The `etc` directory for optional packages
     /// Default: `etc`
     /// With the `--nvim` preset, this is `opt`
-    /// Note: If `etc_root` is set, the package ID is appended.
+    /// Note: If `root` is set, the package ID is appended.
     pub(crate) opt_etc: PathBuf,
+    /// The `src` directory name
+    /// Default: `src`
+    /// With the `--nvim` preset, this is `lua`
+    pub(crate) src: PathBuf,
     /// The `conf` directory name
     /// Default: `conf`
     pub(crate) conf: PathBuf,
@@ -28,14 +33,16 @@ pub struct RockLayoutConfig {
 
 impl RockLayoutConfig {
     /// Creates a `RockLayoutConfig` for use with Neovim
-    /// - `etc_root`: `site/pack/lux`
+    /// - `root`: `site/pack/lux`
     /// - `etc`: `start`
     /// - `opt_etc`: `opt`
+    /// - `src`: `lua`
     pub fn new_nvim_layout() -> Self {
         Self {
-            etc_root: Some("site/pack/lux".into()),
+            root: Some("site/pack/lux".into()),
             etc: "start".into(),
             opt_etc: "opt".into(),
+            src: "lua".into(),
             conf: "conf".into(),
             doc: "doc".into(),
         }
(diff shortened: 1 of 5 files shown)


lux-lib/src/config/tree.rs
```diff
@@ -4,20 +4,25 @@ use std::path::PathBuf;
...
     /// The `etc` directory for optional packages
     /// Default: `etc`
     /// With the `--nvim` preset, this is `opt`
-    /// Note: If `etc_root` is set, the package ID is appended.
+    /// Note: If `root` is set, the package ID is appended.
     pub(crate) opt_etc: PathBuf,
+    /// The `src` directory name
+    /// Default: `src`
+    /// With the `--nvim` preset, this is `lua`
+    pub(crate) src: PathBuf,
```
mrcjkb: suggestion: Should we also add a `lib` field?

lux-lib/src/tree/mod.rs
```diff
@@ -371,19 +371,23 @@ fn mk_rock_layout(
...
+        OptState::Optional => root.join(&layout_config.opt_etc),
     };
-    if layout_config.etc_root.is_some() {
+    if layout_config.root.is_some() {
         etc = etc.join(format!("{}", package.name()));
     }
     let lib = rock_path.join(lib_dir_name);
-    let src = rock_path.join(src_dir_name);
+    let src = if layout_config.root.is_some() {
+        etc.join(&layout_config.src)
```
mrcjkb: suggestion: merge this and the previous `if layout_config.root.is_some()` check into one.
Alone, this is quite confusing.

mrcjkb (changes requested):

lux-lib/src/config/tree.rs
```diff
@@ -28,14 +33,16 @@ pub struct RockLayoutConfig {
...
     /// - `etc`: `start`
     /// - `opt_etc`: `opt`
+    /// - `src`: `lua`
     pub fn new_nvim_layout() -> Self {
         Self {
-            etc_root: Some("site/pack/lux".into()),
+            root: Some("site/pack/lux".into()),
             etc: "start".into(),
             opt_etc: "opt".into(),
+            src: "lua".into(),
```
mrcjkb, line 49: issue: This breaks the `lux.loader` when using the `--nvim` flag, as it assumes `<hash>-<name>@<version>/src`:

https://github.com/lumen-oss/lux/blob/5561ddb483a2f988c71520dae40f8c4b3791bb28/lux-lua/src/loader.rs?plain=1#L43-L55

The `Lockfile` does have a `RockLayoutConfig`:

https://github.com/lumen-oss/lux/blob/5561ddb483a2f988c71520dae40f8c4b3791bb28/lux-lib/src/lockfile/mod.rs?plain=1#L750-L761

so perhaps we can use that to make the loader work.

note: This is why the nix build is failing.

mrcjkb (changes requested):

mrcjkb:
> Would we consider this breaking?

Since this requires a change to the `loader` and it will no longer find modules for `--nvim` entrypoints with the old layout, I'd say this is breaking.

vhyrro:
Cool beans. Tomorrow I'll update the loader and we should be ready to go :D

lux-lib/src/config/tree.rs
```diff
@@ -4,38 +4,50 @@ use std::path::PathBuf;
...
-    pub(crate) opt_etc: PathBuf,
+    /// Note: If `root` is set, the package ID is appended.
+    pub opt_etc: PathBuf,
+    /// The `src` directory name
+    /// Default: `src`
+    /// With the `--nvim` preset, this is `lua`
+    pub src: PathBuf,
+    /// The `lib` directory name
+    /// Default: `lib`
+    pub lib: PathBuf,
```
vhyrro, line 28: note: I found it excessive to create getters in a basic struct like this. These are now simply `pub` so they can be accessible from `lux-lua`. We want these to be public anyway because being able to access the current rock layout from `lux-lib` is definitely a valid library use case.

lux-lib/src/config/tree.rs
vhyrro: This should now be resolved (the loader reuses code from lux-lib to get the proper layout). Added an extra test for good measure.

mrcjkb:
issue: The test build now fails with:

```
lua5.1-foo-nvim> Error:   × syncing dependencies with the project lockfile failed
lua5.1-foo-nvim>   ├─▶ failed to build bar
lua5.1-foo-nvim>   ├─▶ failed to write to '/build/foo-nvim-src/.lux/jit/
lua5.1-foo-nvim>   │   f6ff7c081612f3ee7399947b50c808ba9407b2c2550c5981c2675be07b1c9e3f-
lua5.1-foo-nvim>   │   bar@1.0.0-1/package.rockspec'
lua5.1-foo-nvim>   ╰─▶ No such file or directory (os error 2)
```

Looks like it's failing to write a `package.rockspec` because the directory hasn't been created.

vhyrro:
> issue: The test build now fails with:
> 
> ```
> lua5.1-foo-nvim> Error:   × syncing dependencies with the project lockfile failed
> lua5.1-foo-nvim>   ├─▶ failed to build bar
> lua5.1-foo-nvim>   ├─▶ failed to write to '/build/foo-nvim-src/.lux/jit/
> lua5.1-foo-nvim>   │   f6ff7c081612f3ee7399947b50c808ba9407b2c2550c5981c2675be07b1c9e3f-
> lua5.1-foo-nvim>   │   bar@1.0.0-1/package.rockspec'
> lua5.1-foo-nvim>   ╰─▶ No such file or directory (os error 2)
> ```
> 
> Looks like it's failing to write a `package.rockspec` because the directory hasn't been created.

Nix my beloved. Worked on my machine :tm:.

Let me see what I can do to fix it.

PR #1919, opened 2026-09-05
feat(install): `--path` flag for installing local projects

mrcjkb:


vhyrro (approved):
If CI passes we may merge :)

PR #1932, opened 2026-09-10
feat!: use symlinks for custom entrypoint layouts instead of physically moving directories

vhyrro:


lux-lib/src/tree/mod.rs
```diff
@@ -389,22 +407,19 @@ pub fn mk_rock_layout(
...
     } else {
-        let etc = rock_path.join(&layout_config.etc);
-        let lib = rock_path.join(&layout_config.lib);
-        let src = rock_path.join(&layout_config.src);
-
+        // Always use default directory names for standard layout
+        let etc = rock_path.join("etc");
+        let lib = rock_path.join("lib");
+        let src = rock_path.join("src");
         (etc, lib, src)
```
mrcjkb, line 423: issue: I don't think that's what we want.
If I want to add a custom layout config to my `~/.config/lux/config.toml` that changes `src` to `lua`, this will ignore that setting.

lux-lib/src/tree/mod.rs
```diff
@@ -421,6 +436,71 @@ pub fn mk_rock_layout(
     }
 }
 
+/// Create symlinks from custom layout paths to standard package directories.
+/// This allows external tools (like Neovim) to find files at expected custom paths
+/// while the actual files remain in standard locations.
```
mrcjkb, line 441: quibble:


```suggestion
/// This allows external tools to find files at expected custom paths
/// while the actual files remain in standard locations.
```

lux-lib/src/tree/mod.rs
```diff
@@ -421,6 +436,71 @@ pub fn mk_rock_layout(
...
+pub fn create_custom_layout_symlinks(
```
mrcjkb, line 442: suggestion: I don't think this needs to be `pub`?


```suggestion
fn create_custom_layout_symlinks(
```

lux-lib/src/tree/mod.rs
```diff
@@ -421,6 +436,71 @@ pub fn mk_rock_layout(
...
+    layout_config: &RockLayoutConfig,
+) -> io::Result<()> {
+    let Some(ref root) = layout_config.root else {
+        return Ok(());
+    };
+
+    // Always use default directory names for standard layout
+    let standard_etc = rock_layout.rock_path.join("etc");
+    let standard_lib = rock_layout.rock_path.join("lib");
+    let standard_src = rock_layout.rock_path.join("src");
```
mrcjkb, line 455: issue: see my previous comment.

lux-lib/src/tree/mod.rs
```diff
@@ -421,6 +436,71 @@ pub fn mk_rock_layout(
...
+    let base = tree_root.join(root);
+    let custom_etc = match package.spec.opt {
+        OptState::Required => base.join(&layout_config.etc),
+        OptState::Optional => base.join(&layout_config.opt_etc),
+    }
+    .join(package.name().to_string());
+
+    fs::sync::create_dir_all(&custom_etc).map_err(io::Error::other)?;
+
+    // Create symlinks for src and lib (they always exist)
```
mrcjkb, line 466: quibble: redundant comment


```suggestion
```

lux-lib/src/tree/mod.rs
```diff
@@ -421,6 +436,71 @@ pub fn mk_rock_layout(
...
+    try_create_symlink(&standard_src, &custom_etc.join(&layout_config.src))?;
+    try_create_symlink(&standard_lib, &custom_etc.join(&layout_config.lib))?;
+
+    // Iterate over contents of standard etc directory and create symlinks for each item
```
mrcjkb, line 470: quibble: redundant comment


```suggestion
```

lux-lib/src/tree/mod.rs
```diff
@@ -421,6 +436,71 @@ pub fn mk_rock_layout(
...
+        }
+    }
+
+    Ok(())
+}
+
+fn try_create_symlink(target: &Path, link: &Path) -> io::Result<()> {
+    if link.exists() || link.symlink_metadata().is_ok() {
+        return Ok(());
+    }
```
mrcjkb, line 486: question: can `link.symlink_metadata().is_ok()` be reached if the link doesn't exist?

lux-lib/src/tree/mod.rs
```diff
@@ -421,6 +436,71 @@ pub fn mk_rock_layout(
...
+
+    let relative = pathdiff::diff_paths(
+        target,
+        link.parent()
+            .ok_or_else(|| io::Error::other("invalid parent directory"))?,
```
mrcjkb, line 491: suggestion: include the link in the error message


```suggestion
            .ok_or_else(|| io::Error::other(format!("invalid link: {}", link.display())))?,
```

lux-lib/src/tree/mod.rs
```diff
@@ -421,6 +436,71 @@ pub fn mk_rock_layout(
...
+    )
+    .ok_or_else(|| io::Error::other("failed to compute relative path for symlink"))?;
```
mrcjkb, line 493: suggestion: ditto


```suggestion
    .ok_or_else(|| io::Error::other(format!("failed to compute relative path for symlink {}", link.display())))?;
```

mrcjkb (changes requested):

lux-lua/src/lua_impls.rs
```diff
@@ -1400,20 +1290,6 @@ Default: A `.lux` directory in the workspace root.
...
-        );
-        methods.param("layout", "");
-        methods.add_method(
-            "entrypoint_layout",
-            |_, this, layout: Option<RockLayoutConfigLua>| {
-                Ok(ConfigBuilderLua(this.0.clone().entrypoint_layout(
-                    layout.map(|l| l.0).unwrap_or_default(),
-                )))
-            },
-        );
```
vhyrro, line 1415: issue: I should readd this. Halfway through the refactor I completely ditched RockLayout, but later came to the conclusion out it's nice to keep.

lux-lib/src/build/command.rs
```diff
@@ -110,7 +112,7 @@ impl BuildBackend for CommandBuildSpec {
     }
 }
 
-#[tracing::instrument(level = "trace", skip(config))]
```
mrcjkb, line 113: suggestion: this is a trace-level trace, so it's nice to include all the args we can. If we can't print `output_paths` anymore, we can add it to the skips: `skip(config, output_paths)`

lux-lib/src/build/mod.rs
```diff
@@ -562,24 +553,31 @@ mod tests {
...
+            &PackageSpec::new(rockspec.package().clone(), rockspec.version().clone()),
+            LockConstraint::Unconstrained,
+            RockBinaries::default(),
+            RemotePackageSource::Test,
+            None,
+            LocalPackageHashes {
+                rockspec: "sha256-uU0nuZNNPgilLlLX2n2r+sSE7+N6U4DukIj3rOLvzek="
+                    .parse()
+                    .unwrap(),
+                source: "sha256-uU0nuZNNPgilLlLX2n2r+sSE7+N6U4DukIj3rOLvzek="
```
mrcjkb, line 572: question: Are these dummy hashes that don't have any meaning for the test?
If so, we should make it obvious:


```suggestion
            LocalPackageHashes {
                rockspec: "sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
                    .parse()
                    .unwrap(),
                source: "sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
```

lux-lib/src/luarocks/install_binary_rock.rs
```diff
@@ -86,7 +86,7 @@ where
         rockspec: &'a RemoteLuaRockspec,
         source: RemotePackageSource,
         rock_bytes: Bytes,
-        entry_type: tree::EntryType,
```
mrcjkb, line 89: quibble: I prefer the qualified prefix here; it clarifies that it relates to the `Tree`. (ditto for the other locations where this was changed).

lux-lib/src/tree/dist.rs
```diff
@@ -104,26 +103,19 @@ impl InstallTree for FlatDistTree {
...
-    fn dependency(&self, package: &LocalPackage) -> io::Result<RockLayout> {
-        self.guard_no_conflicting_package(package)?;
-        Ok(mk_rock_layout(
-            &self.root(),
-            &self.bin(),
-            &self.root_for(package),
-            package,
-            &RockLayoutConfig::default(),
-        ))
+    // Dist trees do not undergo traditional cleanup for packages
```
mrcjkb: quibble: redundant comment

lux-lib/src/tree/layout.rs
```diff
@@ -0,0 +1,181 @@
...
+/// A [`CustomRockLayout`] for Neovim plugins. Packages are exposed under `<tree>/site/pack/lux/{start,opt}/<package>`
+#[derive(Clone, Debug, Default)]
+pub struct NvimLayout;
+
+impl NvimLayout {
+    fn target_for(tree: &Tree, package: &LocalPackage) -> PathBuf {
+        let subdir = match package.spec.opt {
+            OptState::Required => "start",
+            OptState::Optional => "opt",
+        };
```
mrcjkb, line 25: suggestion: since `opt` no longer actually needs to install packages to a different directory, perhaps we should remove it from the `LocalPackageId` hash computation.

This would be another breaking change (perhaps for a follow-up PR?).

suggestion 2: We could remove `opt` altogether, since it doesn't actually have any meaning for Lux outside of Neovim. If we do that, we could always symlink packages to `opt` and let lux.nvim take care of loading the packages.

caveat: Doing this would make it harder to manage Neovim plugins using lux-cli without lux.nvim (yagni?).

lux-lib/src/tree/layout.rs
```diff
@@ -0,0 +1,181 @@
...
+
+fn try_create_symlink(target: &Path, link: &Path) -> fs::Result<()> {
+    if link.symlink_metadata().is_ok() {
+        return Ok(());
+    }
+
+    let relative = pathdiff::diff_paths(
+        target,
+        link.parent()
+            .ok_or_else(|| FsError::Other("invalid parent directory".into()))?,
```
mrcjkb: suggestion: I think this is unreachable?

```suggestion
            .ok_or_else(|| unreachable!("missing parent directory"))?,
```

lux-lib/src/tree/layout.rs
```diff
@@ -0,0 +1,181 @@
...
+    };
+
+    fn mock_hashes() -> LocalPackageHashes {
+        LocalPackageHashes {
+            rockspec: "sha256-uU0nuZNNPgilLlLX2n2r+sSE7+N6U4DukIj3rOLvzek="
+                .parse()
+                .unwrap(),
+            source: "sha256-uU0nuZNNPgilLlLX2n2r+sSE7+N6U4DukIj3rOLvzek="
+                .parse()
+                .unwrap(),
```
mrcjkb, line 111: suggestion:


```suggestion
            rockspec: "sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
                .parse()
                .unwrap(),
            source: "sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
                .parse()
                .unwrap(),
```

lux-cli/src/bin/lx.rs
```diff
@@ -105,7 +105,7 @@ async fn main() -> Result<()> {
         .no_tfa(Some(cli.no_tfa));
 
     if cli.nvim {
-        config_builder = config_builder.entrypoint_layout(RockLayoutConfig::new_nvim_layout());
+        config_builder = config_builder.entrypoint_layout(NvimLayout);
     }
```
mrcjkb, line 109: note(non-blocking for this PR): This works because we currently have a single preset: `--nvim`.
suggestion: Perhaps we should replace the `--nvim` flag with a `--preset=nvim` flag?

mrcjkb (changes requested):
note: This isn't a refactor (it changes behaviour). The commit prefix should be `feat!`.

lux-cli/src/bin/lx.rs
vhyrro: sure, let's address this in a separate PR though. Tracking issue: #1947

lux-lib/src/build/mod.rs
vhyrro: I did this because all tests that create the neorg local package use this set of hashes (with no exceptions). I can change it if we want, but it's just standard in our code, so I kept it.

vhyrro: We may want to consider creating a unified test library, with one function for creating a mock data that is then used everywhere.

lux-lib/src/tree/layout.rs
vhyrro: Since we already have the logic built-in, I think we might be better off keeping it. We may want to eventually support [optional dependencies or feature flags](https://doc.rust-lang.org/stable/cargo/reference/features.html#optional-dependencies). Reimplementing opt packages will just cost us more time.

vhyrro:
Are the build failures being triggered by my changes? Looks like they're permissions issues with the homeless shelter (completely unrelated to my code changes?)

lux-lib/src/build/mod.rs
vhyrro: Let me know if we would like to keep this or go through with the hash changes :)

mrcjkb: I don't feel strongly about this.

lux-lua/src/lua_impls.rs
mrcjkb: Is this still valid?

mrcjkb (approved):

PR #1941, opened 2026-09-13
fix(xtask-lua): pick up `LUX_LIB_DIR` for generated `.pc` file

mrcjkb:


mrcjkb:
Nix build was the only failing check. Fixed and ran it locally.

Issue #274, opened 2024-12-22
[Feature] `lx list --removable` flag


mrcjkb:
Should list any rocks that are not dependencies of other rocks.

vhyrro:
Maybe as part of `rock list` or `rocks remove --dry-run`?

mrcjkb:
`--dry-run` doesn't make sense to me because there's no `rocks remove` without arguments.
But `list` sounds like a good option.

mrcjkb:
Note: most of the logic should be implemented in `lux-lib` so we can expose an api in `lux-lua`

MarcusDavidG:
Gm @mrcjkb, my name is Marcus David. I’m a developer with experience working on CLI features and package managers.

I’d like to work on Issue #274: Add lx list --removable flag. I’ll implement logic to list rocks that aren’t dependencies of others, keeping it consistent with existing CLI patterns.

ETA: 24 hours.

Happy to contribute. Thanks!

PR #1962, opened 2026-09-21
feat(cli): add `--removable` flag to `lx list` (closes #274)

vhyrro:


lux-cli/src/list.rs
```diff
@@ -10,12 +17,28 @@ use crate::args::OutputFormat;
 pub struct ListCmd {
     #[arg(long, default_value = "text", value_enum, ignore_case = true)]
     output_format: OutputFormat,
+
+    /// Only list rocks that are not reachable from an entrypoint.
```
mrcjkb, line 21: suggestion: an entrypoint is removable unless it's also a dependency of another entrypoint. Perhaps this should include removable entrypoints too.

mrcjkb (approved):

lux-cli/src/list.rs
vhyrro: I've found that the best option is to implement two variants: `--orphans`, with the current behaviour. You can think of these as packages that Lux hasn't cleaned up properly and that can be "garbage collected".

I've also added a `--removable` which behaves more like the name suggests: lists all packages that are removable (not a dependency of anything else) and that won't cause errors when running `lx remove`.

mrcjkb (approved):

PR #1963, opened 2026-09-21
test: replace `LUX_SKIP_IMPURE_TESTS` with an `impure_tests` feature (closes #1154)

vhyrro:


mrcjkb (commented):
issue: perviously, `cargo tt` would run the impure tests.
We will need to remember to run `cargo tti` separately.
Perhaps it's best to run them in CI as part of the integration test suite, but mark them as flaky, like we do with some existing integration tests.

vhyrro:
I've done that in the second commit. Let me know if it looks good :)

mrcjkb (approved):

